PR #1, opened 2018-05-23
SmallRye Config initial import

jmesnil:
This initial import takes the code from
https://github.com/wildfly-extras/wildfly-microprofile-config and keeps
only the implementation and tck modules.

The GAV of the project is changed to io.smallrye:smallrye-config:1.0.0-SNAPSHOT

Code is from https://github.com/jmesnil/wildfly-microprofile-config/tree/smallrye


jmesnil:
@smallrye/collaborators This is the initial import for the SmallRye Config project.

Could one of you please review it?

README.md
```diff
@@ -1 +1,27 @@
-# smallrye-config
+# SmallRye Config
+
+
+SmallRye Config is an implementation of [Eclipse MicroProfile Config][microprofile-config].
+
+[![Build Status](https://ci.wildfly.org/app/rest/builds/buildType:(id:MicroProfileJsr_WildflyMicroprofileConfig_MpConfig1xBranch)/statusIcon)](https://ci.wildfly.org/viewType.html?buildTypeId=MicroProfileJsr_WildflyMicroprofileConfig_MpConfig1xBranch&guest=1)
```
jmesnil: I kept the link to WildFly CI that was used for wildfly-microprofile-config.

Should I also set it up for that project or had you in mind another CI infra?

kenfinnigan: I was thinking of using Semaphore CI, as we'll eventually need one more jobs for each repo and will be easier if people outside RH can set them up as well

pom.xml
```diff
@@ -0,0 +1,194 @@
...
+            <dependency>
+                <groupId>io.smallrye</groupId>
+                <artifactId>smallrye-config</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <repositories>
+        <repository>
```
kenfinnigan: Are there artifacts from JBoss Nexus we need that aren't in Central?

jmesnil: No, I tried it and everything can be fetched from Central?

Should I remove this repository from the POM?

README.md
jmesnil: I removed the link

kenfinnigan: Thanks

pom.xml
jmesnil: done

kenfinnigan:
Not stopping the merge, but what are your thoughts regarding /docs?

Were you thinking that would be published somewhere?

jmesnil:
@kenfinnigan I was not sure about the docs. 

They are quite readable from GitHub directly at https://github.com/smallrye/smallrye-config/blob/master/docs/src/main/asciidoc/index.adoc but I'd prefer to have a better place to host them.

I see at least 2 options:
* publish them in GitHub pages for the project
* publish them on smallrye.io

kenfinnigan:
Likely publish on smallrye.io makes sense, though that could be just a GH pages project.

Need to figure that out

PR #4, opened 2018-05-31
Add missing NOTICE

jmesnil:


johnament:
@jmesnil Notice file should include all contents from https://github.com/apache/geronimo-config/blob/trunk/NOTICE

jmesnil:
@johnament the code I used was from https://github.com/struberg/javaConfig, not geronimo-config. I used the same content than its notice https://github.com/struberg/javaConfig/blob/master/NOTICE.

Isn't that correct?

PR #7, opened 2018-06-26
Extensibility

bobmcwhirter:
Extensibility and serializability for 1.3.

Also, better ordinal comparison to avoid integer overflow/underflow when someone, like myself, uses Integer.MAX_VALUE.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -36,7 +36,7 @@
     <version.eclipse.microprofile.config>1.3</version.eclipse.microprofile.config>
     <version.javax.javaee-api>7.0</version.javax.javaee-api>
     <version.javax.enterprise.cdi-api>1.2</version.javax.enterprise.cdi-api>
-    <version.javax.annotation-api>1.2</version.javax.annotation-api>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -36,7 +36,7 @@
     <version.eclipse.microprofile.config>1.3</version.eclipse.microprofile.config>
     <version.javax.javaee-api>7.0</version.javax.javaee-api>
     <version.javax.enterprise.cdi-api>1.2</version.javax.enterprise.cdi-api>
-    <version.javax.annotation-api>1.2</version.javax.annotation-api>
+    <version.javax.annotation-api>1.3.2</version.javax.annotation-api>
```
kenfinnigan: Is this an EE7 -> EE8 change?

bobmcwhirter: Mostly it's that Thorntail v4 uses @Priority against non-Type targets, and 1.2 isn't wide-enough in its application. I could dependency-manage that dependency in TTv4 instead, if we would rather not bring newer annotation-api into this commit.

PR #8, opened 2018-06-29
Zkconfig

sjwoodman:
Added a ZooKeeper backed ConfigSource and example application that uses it.


kenfinnigan:
Should this also be added to the main Maven reactor build?

How is this expected to be released otherwise?

config-sources/zookeeper/src/main/java/io/streamzi/config/zk/ZkConfigSource.java
```diff
@@ -0,0 +1,147 @@
...
+        /*
+         * Explicitly ignore all keys that are prefixed with the prefix used to configure the Zookeeper connection.
+         * Other wise a stack overflow obviously happens.
+         */
+        if (key.startsWith(ignoredPrefix)) {
+            return null;
+        }
+        try {
+
+            final Stat stat = getCuratorClient().checkExists().forPath(applicationId + "/" + key);
```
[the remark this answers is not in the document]
sjwoodman: getCuratorClient() will throw a ZooKeeperConfigException if the properties are not set.

sjwoodman:
@bobmcwhirter I've updated the PR to address your comments.

kenfinnigan:
Doesn't look like it's part of the Maven build yet right?

Any reason for that?

sjwoodman:
@kenfinnigan Added and (now) passing tests.

config-sources/zookeeper/pom.xml
```diff
@@ -0,0 +1,107 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
```
kenfinnigan: This should probably inherit from the config parent POM so that versions for a lot of these things can be pulled in automatically, and also so that release configuration is passed down

config-sources/zookeeper/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+            <version>${version.curator}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.eclipse.microprofile.config</groupId>
+            <artifactId>microprofile-config-api</artifactId>
+            <version>${version.microprofile.config}</version>
+        </dependency>
+
+        <dependency>
```
kenfinnigan: This needs to be pulling in smallrye-config, and nothing from WF Swarm/Thorntail

config-sources/zookeeper/example/pom.xml
```diff
@@ -0,0 +1,104 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
```
kenfinnigan: Given there's already a test, i'm not sure about an example here as well.

My reason for that is it's an additional thing to keep up to date, and could cause issues with changes in smallrye-config being required for the example to work but WF Swarm/Thorntail not having a version available with the necessary smallrye-config changes. We end up in a bit of a loop

config-sources/zookeeper/pom.xml
```diff
@@ -0,0 +1,107 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>io.smallrye</groupId>
+    <artifactId>zookeeper-config-source</artifactId>
```
kenfinnigan: I'm also wondering if we need an intermediate pom in /config-sources that changes the groupId to `io.smallrye.config-sources` and then this artifact inherits that groupId and has an artifactId of `zookeeper`?

sjwoodman:
My branch has become corrupted following a rebase. I'm closing this PR and will submit a new one that addresses your comments.

PR #10, opened 2018-07-17
Fixes #9: Default value for injected collection

### LINKED ISSUE #9: Support default values for injected collections
If the application defines an injected collection with a default value and the property is not configured in any config source:

```
@Inject
@ConfigProperty(name = "myPets", defaultValue = "horse,monkey")
private String[] myArrayPets;
 
@Inject
@ConfigProperty(name = "myPets", defaultValue = "cat,lama")
private List<String> myListPets;
 
@Inject
@ConfigProperty(name = "myPets", defaultValue = "dog,mouse")
private Set<String> mySetPets;
```

Then, the default value must be used to fill the collections:

```

myArrayPets : [horse,monkey]
myListPets  : [cat,lama]
mySetPets   : [dog,mouse]
```

In the current codebase, `defaultValue` is taken into account only for array but not for `List` and `Set`.

Originally reported by https://issues.jboss.org/browse/WFWIP-55

jmesnil:
* Use default values for injected collection if the property is not
  configured.
* add testsuite/extra module to provide tests in addition to the ones
  from the MicroProfile Config TCK
* add CollectionWithConfiguredValueTest & CollectionWithDefaultValueTest
  to verify the expected behaviour
* move tck module under testsuite


testsuite/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+ ~  See the License for the specific language governing permissions and
+ ~  limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.smallrye</groupId>
+    <artifactId>smallrye-config-parent</artifactId>
```
kenfinnigan: I think this is why CI failed. Needs to be `smallrye-config-testsuite`

PR #29, opened 2018-09-06
sm - add string interpolation

kg6zvp:
I believe there are many use cases for the ability to interpolate strings in configuration files and other places.

This pull request adds the ability to transform this:

`<property name="javax..action" value="${jpa.database.action}"/>`

into this:

`<property name="javax..action" value="none"/>`

### Justification

This belongs in smallrye-config because, even though it isn't directly useful as a part of microprofile-config, it will be useful in other projects utilizing smallrye-config and provides a common place for the code to reside.

If this seems like the wrong place for it, I would be happy to integrate this small change into Thorntail directly.

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/StringInterpolator.java
+++ b/implementation/src/main/java/io/smallrye/config/StringInterpolator.java
@@ -1,0 +1,30 @@
+package io.smallrye.config;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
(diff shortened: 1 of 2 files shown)


kenfinnigan:
Would this require changes to Thorntail, or anything else using SmallRye Config, to then perform the interpolation?

kg6zvp:
Yes, but the changes are minimal. All that's necessary is to call the interpolator in order to look for and inject variables.

```java
String strFromFile = load(...);
strFromFile = StringInterpolator.interpolate(strFromFile);
```

PR #34, opened 2018-09-06
skip interpolating escaped strings, closes #33

### LINKED ISSUE #33: Skip interpolation of escaped strings
Strings like `\${config.varname}` are currently treated as config variables. These should be skipped and interpolated into `${config.varname}`

kg6zvp:
- code cleanup
- added test case for escaped vars


kenfinnigan:
Thoughts on this @jmesnil, @mkouba?

mkouba:
If I understand it correctly `StringInterpolator` is just a util class, not used anywhere in `smallrye-config
` codebase?

kenfinnigan:
That's my understanding yes

mkouba:
Well, I'm not a fan of exposing generic util classes like this but since it's already there we should at least add some javadoc, make the class final and add a private constructor ;-)

bobmcwhirter:
I think I intended (if I even wrote the code? I did for Thorntail) for it to be purely internal, not a public utility.  Who knows.  It's friday.

kg6zvp:
I'll take the blame for adding this. I submitted a PR with this yesterday, I think, with the intention of integrating it's use into Thorntail. As stated on PR 36, I'm happy to refactor this into a separate utility for integration with Thorntail.

kg6zvp:
Will be integrated into Thorntail otherwise

PR #36, opened 2018-09-07
add resource interpolator, closes #32

### LINKED ISSUE #32: Add ability to load interpolated resource
Resource files are often read in java programs and it would be helpful to be able to read resource files transparently with their config values inserted into them already.

Proposed uses:

```java
// Load file and obtain reference to interpolated file
File StringInterpolator.interpolate(File file);

// Load URL and obtain reference to interpolated URL
URL StringInterpolator.interpolate(URL url);
```

kg6zvp:


kenfinnigan:
@kg6zvp I'm not sure whether something like this fits within an MP Config implementation. Though it's using MP Config values to interpolate the file, not sure whether it's the job of this impl to interpolate config into everything.

Thoughts @jmesnil, @mkouba?

kg6zvp:
It doesn't fit if we're following the Single Responsibility Principle.

I would be happy to refactor this out into its own library if that's preferred. Though, this seems like a utility that could be considered an enhancement to mp-config as it would be useful in a lot of places you would want mp-config

My only reason for including it is in order to integrate it into both Thorntail v2 and v4 for various files like persistence.xml

kg6zvp:
Bump

kg6zvp:
Will be integrated elsewhere

Issue #38, opened 2018-09-26
SmallRyeConfigProviderResolver is not thread-safe


jmesnil:

kenfinnigan:
Merged the fix

PR #39, opened 2018-09-26
[#38] Make SmallRyeConfigProviderResolver thread-safe

jmesnil:
Issue: https://github.com/smallrye/smallrye-config/issues/38

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/SmallRyeConfigProviderResolver.java
+++ b/implementation/src/main/java/io/smallrye/config/SmallRyeConfigProviderResolver.java
@@ -43,17 +43,20 @@ public class SmallRyeConfigProviderResolver extends ConfigProviderResolver {
     @Override
     public Config getConfig(ClassLoader classLoader) {
         Config config = configsForClassLoader.get(classLoader);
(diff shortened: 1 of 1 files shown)


implementation/src/main/java/io/smallrye/config/SmallRyeConfigProviderResolver.java
```diff
@@ -63,20 +66,22 @@ public ConfigBuilder getBuilder() {
 
     @Override
     public void registerConfig(Config config, ClassLoader classLoader) {
-        configsForClassLoader.put(classLoader, config);
+        synchronized (configsForClassLoader) {
```
kenfinnigan: Noob sync question here. Can you sync on a variable that's already synchronized on?

jmesnil: Noob developer here :)

You are right, I need to synchronise on the instance, not the collection itself. thanks for the catch.

jmesnil:
@kenfinnigan I've amended the commit. is it ok for you now?

PR #41, opened 2018-10-30
Reuse smallrye producer implementation

guhilling:
I’m currently working on extending my junit5 testing-framework for cdi (https://github.com/guhilling/cdi-test) for eclipse microprofile.
To accomplish this I’ll provide an annotation to inject config values on a per-test basis as follows:

    @Test
    @ConfigPropertyValue(name = "some.string.property", value = "value method b")
    public void overrideValuesInMethod() {
        Assertions.assertEquals("value method b", controller.getStringProperty());

I think, you get the idea… The code is already working and you can find it at https://github.com/guhilling/cdi-test-microprofile .
As you would have guessed, I’m using smallrye-config behing the scenes. Now I got the problem that I have to duplicate some of the code from the ConfigProducer class because I need to create a new Config Instance for each test case.

The pull request solves my problem and I also removed some duplications...

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/Converters.java
+++ b/implementation/src/main/java/io/smallrye/config/Converters.java
@@ -42,6 +42,7 @@ class Converters {
                     || "JA".equalsIgnoreCase(value)
                     || "J".equalsIgnoreCase(value)
                     || "SI".equalsIgnoreCase(value)
+                    || "SIM".equalsIgnoreCase(value)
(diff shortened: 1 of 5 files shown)


kenfinnigan:
Sorry for the delayed response!

There was another change made to the same class. Would it be possible for you to rebase?

guhilling:
Sure!

guhilling:
Would You check?
Was just a diff of one line, but I had to apply it manually (to a different file).

Issue #60, opened 2018-12-13
Add built-in support for non-object Optional* types


dmlloyd:
The types `OptionalInt`, `OptionalLong`, and `OptionalDouble` provide a better means to abstract an optional primitive value than boxed types.  It would be nice if it were possible to use these types from `Config`.

Registering a custom `Converter` in conjunction with `Config#getValue` is not sufficient, because converters are only called when a `String` value is present in the configuration.  Using `getOptionalValue` is unwieldy because the result would be of type `Optional<OptionalInt>` or similar.

kenfinnigan:
Resolved with #62

PR #61, opened 2018-12-13
[#60] Support for non-object Optional* types

dmlloyd:


dmlloyd:
This is one possible approach for how `Optional*` could be implemented.  An advantage of this approach is that the existing converters infrastructure is reused.  A disadvantage is that it gets a little weird in that you can have arrays/lists of `Optional*`s which might or might not be an anti-pattern of sorts.

jmesnil:
@dmlloyd thanks, that is a good enhancement.

I took the liberty to take your PR and update it to put the support of these optional types into the `Config#getValue` method (there is no reason these types could not be supported from the Config API directly)
I've also added tests to check the behaviour.

If that's ok with you, I'll close this PR and merge #62 that contains your code + mine.

dmlloyd:
That's fine, thanks. As long as you think that the approach is okay that is.

jmesnil:
superseded by #62

PR #62, opened 2018-12-13
 [#60] Support for non-object Optional* types

jmesnil:


implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -58,6 +58,16 @@ protected SmallRyeConfig(List<ConfigSource> configSources, Map<Type, Converter>
                 return convert(value, aClass);
             }
         }
+
+        // check for  Optional numerical types to return their empty()
+        // if the property is not found
+        if (aClass.isAssignableFrom(OptionalInt.class)) {
+            return (T)OptionalInt.empty();
```
dmlloyd: This could be `aClass.cast(OptionalInt.empty())` and that would make it a safe cast.

jmesnil: fixed, thanks

jmesnil:
@kenfinnigan could you please review and approve?
 
I need an approval from somebody else with write access to be a able to merge

Issue #63, opened 2018-12-13
Config.getValueCollection()


dmlloyd:
It would be great if the base `Config` interface had methods to get collections of single-valued items.  The API could accept a function that accepts a size estimate and yields some kind of collection type.

I'll send up a PR as an example of how the API could look, if added on `SmallRyeConfig`.

jmesnil:
+1

The `Config` API only allows programmatic lookup of arrays (
https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/converters.asciidoc#array-converters)  while the CDI injection supports array, `List` and `Set`.

PR #64, opened 2018-12-13
[#63] A multi-value config get operation

dmlloyd:
This is a possible implementation of #63.


dmlloyd:
Many collection classes have a constructor that takes a single integer which is a size hint; these collections types can be passed in using method references.  For other cases, where the collection type has a no-arg constructor, there could be a `Supplier<C>` variant.  The method would probably have to have a different name in order to disambiguate cases where the class provides both (which includes `ArrayList` and `HashSet`, two choices which are likely to be popular).

jmesnil:
having some tests would have been great

kenfinnigan:
What can I say, I'm over eager

Issue #70, opened 2018-12-15
Spec issue: ConfigSource.getProperties()


dmlloyd:
Implementing `ConfigSource.getProperties()` is a far more onerous burden than the rest of the `ConfigSource` API.  In some cases this can be a simple view of an existing map (like system properties), but in other cases one must implement an entire custom `Map` implementation, which can require as many as 7 classes to implement (the Map, its key set & iterator, its value collection & iterator, and its entry set & iterator).  A simple property-expanding decorator requires 5 classes to implement its map view.

The spec currently implements the far simpler `getPropertyNames` method in terms of the abstract `getProperties` method, however it would have been more convenient had the spec done it the other way around.

It might be worthwhile to provide a `Map` implementation which is backed by the `ConfigSource` `getPropertyNames` and `getValue` methods.

jmesnil:
This issue is a good argument to work on https://github.com/eclipse/microprofile-config/issues/333 to update the ConfigSource API to simplify the requirements from the implementation.

dmlloyd:
Solved by #72.

PR #72, opened 2018-12-15
[#70] Implement a helper Map implementation for ConfigSources to use

dmlloyd:


dmlloyd:
Added tests.

implementation/pom.xml
```diff
@@ -69,6 +69,50 @@
           <forkMode>once</forkMode>
         </configuration>
       </plugin>
+      <plugin>
```
kenfinnigan: Should something like this be moved into `smallrye-parent` so all projects could use it?

implementation/src/main/java/io/smallrye/config/ConfigSourceMap.java
```diff
@@ -0,0 +1,180 @@
...
+ * This should not be used to implement {@link ConfigSource#getProperties()} on {@code ConfigSource}
+ * instances which do not override {@code getPropertyNames()}, as this will result in infinite recursion.
+ *
+ * @implNote The key set of the map is the result of calling {@link ConfigSource#getPropertyNames()}; the rest
+ * of the map operations are derived from this method and {@link ConfigSource#getValue(String)}.
+ * The values collection and entry set are instantiated lazily and cached.
+ * The implementation attempts to make no assumptions about the efficiency of the backing implementation and
+ * prefers the most direct access possible.
+ * <p>
+ * Some mutate operations are supported (specifically removals), where they can be performed on the backing key set, if
```
jmesnil: that's not specified in the spec (although it should be) but my interpretation of `getPropertyNames()` is that it should return an immutable set of names.
In general, `ConfigSource` should not be used to modify/update configuration. If some of these methods provide ambiguity about it, it should be properly specified in the spec or in the API.

I'll open issues in `microprofile-config` to clarify these use cases

implementation/pom.xml
dmlloyd: I'd be okay with that I guess.

implementation/src/main/java/io/smallrye/config/ConfigSourceMap.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    public void clear() {
+        delegate.getPropertyNames().clear();
+    }
+
+    public String remove(final Object key) {
+        final String value = get(key);
+        //noinspection SuspiciousMethodCalls - it's OK in this case
+        return delegate.getPropertyNames().remove(key) ? value : null;
```
jmesnil: `remove` could throw a `UnsupportedOperationException` if the list is unmodifiable

implementation/pom.xml
dmlloyd: But that means there'd have to be a release and so forth.

kenfinnigan: True, but pulling the trigger is pretty easy.

Other than needing a release of parent, any reason not to?

dmlloyd: No, sounds fine.  Is that something you can do?

kenfinnigan: Sure

implementation/src/main/java/io/smallrye/config/ConfigSourceMap.java
jmesnil: The API says:  https://github.com/eclipse/microprofile-config/blob/4611c9433675e34ee73bd1eba026e01631582772/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java#L39

jmesnil: there is also https://github.com/eclipse/microprofile-config/issues/333 where it is discussed

dmlloyd:
The idea was to "punt" on mutability.  If the underlying config source is mutable then the map is; if it isn't, then the map isn't.  This way it doesn't impose any policy one way or the other.

I can make it assume full immutability though if you want me to.

jmesnil:
@dmlloyd I would prefer full immutability (the goal being then to propose this improvement to microprofile-config to update its API wrt `getProperties()/getPropertyNames()` ).

I don't like that the application could change the config sources.
First, the app does not "own" the config source, it may not be the only one that uses the config sources (eg config maps, DBs).
Then the config source itself may not be the same one that provides the properties depending on the stage (e.g. on dev, the prop would come from embedded property files, on prod, from a config map).

Having mutability raise expectation that the application knows the config source that always provides the property which is the opposite of the design behind the `Config` interface.

The whole microprofile-config API is really only about viewing the configuration. The update/modification of the configuraiton is out of scope of the specification.
However, this is not clearly enough stated in the spec and the use of mutable interface such as `Properties` leaves open the possibility to have mutable ConfigSources.

dmlloyd:
OK I'll push an immutable version.

dmlloyd:
OK pushed (also remembered to update the tests ;) )

dmlloyd:
I think CI might have been confused by my rapid pushes.  Can we trigger a retest?

dmlloyd:
I guess I'll just re-amend and push a new commit ID

dmlloyd:
Hmm

dmlloyd:
Ah it's the JavaDoc issue.  We need Ken's parent update.

kenfinnigan:
v2 has been pushed to Sonatype, basically waiting for percolation to Central.

kenfinnigan:
`2` is now in Central

dmlloyd:
Thanks Ken, I'll update the PR.

jmesnil:
@dmlloyd could you please rebase and I'll merge it?

Issue #74, opened 2019-01-09
io.smallrye.config.SmallRyeConfigBuilder#build should be covariant


dmlloyd:
`io.smallrye.config.SmallRyeConfigBuilder#build` should covariantly return `SmallRyeConfig`.  Since that method is implementing `org.eclipse.microprofile.config.spi.ConfigBuilder#build` which returns a `Config`, it is a binary-compatible change.

dmlloyd:
Not fixable due to the config factory SPI.

PR #75, opened 2019-01-09
[#74] Make SmallRyeConfigBuilder#build covariantly return SmallRyeConfig

dmlloyd:


dmlloyd:
Never mind, this can't work due to the config factory SPI.

Issue #87, opened 2019-02-20
SmallRyeConfigProviderResolver.registerConfig(config, null) should register for the current TCCL


jmartisk:
The Javadoc for `ConfigProviderResolver.registerConfig` says:

> If the ClassLoader is {@code null} then the current Application will be used.

However, if that method is called with null in the `classLoader` parameter, then `null` will be put into the `configsForClassLoader` map: https://github.com/smallrye/smallrye-config/blob/1.3.5/implementation/src/main/java/io/smallrye/config/SmallRyeConfigProviderResolver.java#L70

PR #88, opened 2019-02-20
[#87] register for TCCL if registerConfig is called with null class l…

jmartisk:
…oader parameter

https://github.com/smallrye/smallrye-config/issues/87


jmesnil:
I think this method should expect a valid `ClassLoader` parameter. I think it'd be be better to raise an exception if the class loader is `null` (and it the caller wants to register the TCCL, he has to pass it directly).

jmartisk:
Javadoc in the specification says that `null` should be transformed into the current application's class loader: https://github.com/eclipse/microprofile-config/blob/1.3/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigProviderResolver.java#L75 so I guess we should stick to that.

jmesnil:
thanks for the link, you're right, your PR can go as it is :)

PR #102, opened 2019-05-08
Treat empty list values as missing

### LINKED ISSUE #83: Inconsistent behaviour between `Config.getValue()` and `Config.getOptionalValue()`
I stumbled upon the different behaviour of `Config.getValue()` and `Config.getOptionalValue()` method.

When you define a configuration and give it an empty value, e.g. something like `org.foo.bar=` (expected value here is `String[]`, then `getValue()` will (IMO correctly) return empty string, whereas `getOptionalValue()` gives me empty optional which I cannot retrieve.

Now, I don't really see spec mentioning this in detail, but those two methods should behave equally.

### LINKED ISSUE #84: Non-optimal behavior for collection-based configuration keys
Both the standard `List` injection and our own `io.smallrye.config.SmallRyeConfig#getValues` use the same mechanism to expand a list of items.  However neither one has any special handling for empty values so chances are things like this will fail:

```java
// config contains: test.key=1,2,,4,5
config.getValues("test.key", Integer.class, ArrayList::new);
```

The problem is that if we start supporting property expansion, this kind of thing becomes possible when you have an input string like: `1,2,${maybe.three?},4,5`.

I propose that we treat empty list items as not present, so the above example would result in a four-element list of `[1,2,4,5]`.  But this would require a spec change as well because of the list support in CDI.  WDYT?

dmlloyd:
Fixes #83, fixes #84.  Makes handling of an empty string and empty list items consistent with the existing equivalency between an empty string and a `null` or missing value.

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/StringUtil.java
+++ b/implementation/src/main/java/io/smallrye/config/StringUtil.java
@@ -16,20 +16,22 @@
 
 package io.smallrye.config;
 
+import java.util.regex.Pattern;
+
 /**
  * @author <a href="http://jmesnil.net/">Jeff Mesnil</a> (c) 2017 Red Hat inc.
  */
 public class StringUtil {
 
     // delimiter is a comma that is not preceded by a \
-    private static final String DELIMITER = "(?<!\\\\),+";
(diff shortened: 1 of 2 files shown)


dmlloyd:
This is a pretty simple change, but it's broken into commits to show each step.

PR #114, opened 2019-05-13
Empty values are always treated as non-present

### LINKED ISSUE #83: Inconsistent behaviour between `Config.getValue()` and `Config.getOptionalValue()`

dmlloyd:
Fixes #83 (for real)

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
+++ b/implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
@@ -57,6 +57,10 @@ public class SmallRyeConfig implements Config, Serializable {
         for (ConfigSource configSource : configSources) {
             String value = configSource.getValue(name);
             if (value != null) {
+                if (value.isEmpty()) {
+                    // empty collection
+                    break;
+                }
                 String[] itemStrings = StringUtil.split(value);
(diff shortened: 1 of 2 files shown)


dmlloyd:
This is a draft PR because it does not pass TCK due to https://github.com/eclipse/microprofile-config/pull/407

kenfinnigan:
Depending on what happens with the spec, we can "exclude" the failing test(s) so the build passes if we need it released.

It's safe to exclude on the premise we're challenging the TCK

dmlloyd:
OK I've disabled the offending TCK tests.

jmesnil:
Let's merge this and discuss the spec behaviour in https://github.com/eclipse/microprofile-config/pull/407

kenfinnigan:
Do we need another release?

dmlloyd:
Yes please

PR #129, opened 2019-06-27
Add get*Value variants that accept an explicit Converter

dmlloyd:


dmlloyd:
This covers a use case in Quarkus where we want to be able to specify explicit converters on a per-item basis.  Right now we have to get the value as a string and then convert it but this is not ideal.

PR #130, opened 2019-06-28
Fix converters

jmesnil:


implementation/src/main/java/io/smallrye/config/Converters.java
```diff
@@ -94,6 +94,17 @@ private Converters() {}
         }
     });
 
+    @SuppressWarnings("unchecked")
+    static final Converter<Integer> CHARACTER_CONVERTER = BuiltInConverter.of(11, (Converter & Serializable) value -> {
```
dmlloyd: Should this not be `Converter<Character>`?

jmesnil: you're right

PR #133, opened 2019-07-02
Donation from MicroProfile-ext: Json converters

phillip-kruger:
Json converters from MicroProfile-ext.


derekm:
`extensions` folder seems very catch-all ... Does it mean "everything not defined by the spec"? Then as things become defined by MP specs, they move out of the `extensions` folder?

I liked the structure we had in MP-Ext, we could do similarly here with `sources`, `converters`, `events`, `examples`, etc., instead of pushing those down a level under `extensions`...

Can we use @kenfinnigan's technique for preserving history, if at all possible?

There are extensions in `implementation` that need to be moved out if we're strictly separating impls (spec) from exts (non-spec). If we're not strictly separating them, then we can just use component type categories as suggested above and include the standard bits into `implementation` as dependencies.

kenfinnigan:
`extensions` was the original idea I'd had a while back about how to group things. Maybe it's not the best, open to ideas on that.

There is currently no strict separation between spec implementation and everything else, because we intend to go beyond the spec in the implementation.

Now maybe that opens up a question as to where things move to? For instance, maybe `converters` should just be added to the implementation so that it's not necessary for them to be deliberately added as a dependency, they're just there. Conversely, to me it does make sense for config sources to be a "choice" by the developer as to what they want available.

phillip-kruger:
Remember that all extensions should also work on any MicroProfile server, and not just with the once that use Smallrye implementation. So user will have to be able to add them separately

derekm:
`implementation` contains [`DirConfigSource`](https://github.com/smallrye/smallrye-config/blob/master/implementation/src/main/java/io/smallrye/config/DirConfigSource.java), which is an extension and would be nice to use on any MicroProfile implementation provided by other vendors, for example.

Certainly outside the scope of this PR, but considering it helps inform us about the more ideal project structure, which is within the scope of this PR.

In order for me to benefit from it in another MP impl or another runtime, I think to pull it in piecemeal, in order for SmallRye to remain backward compatible with its own prior releases, it needs to pull the piecemeal jar back into `implementation` as a dependency.

Similarly, when we get to the à la carte Config Sources, we have significant overlap with [`PropertiesConfigSource`](https://github.com/smallrye/smallrye-config/blob/master/implementation/src/main/java/io/smallrye/config/PropertiesConfigSource.java) for external properties files on the filesystem or the web.

That's all besides the question of whether the internal/standard config sources should benefit from the base config source we'll be bringing in.

I'm sorry for the information overload, but this PR definitely sets a few of these patterns, so it might be worth considering ahead of time.

kenfinnigan:
I know @phillip-kruger has already put a lot of work into the PRs to bring across MP Extensions project, and I don't want to unnecessarily delay the migration.

However, if there are over-arching concerns around how things should be structured, maybe it's best we go back to the mailing list thread to discuss further: https://groups.google.com/forum/#!topic/smallrye/YdMmpmUCEe4

phillip-kruger:
Hi All. 
No pressure, I am just checking - Any update on this ?

dmlloyd:
I think we're stalled on determining how this should be structured.  I think the choices on the table were:

* Integrate into `implementation` directly
* Add a dedicated submodule e.g. `smallrye-config-json` under `config-sources`
* Integrate into a general `ext` kind of thing

Let me know if I missed an option.

dmlloyd:
FTR I prefer `config-sources/json` with `smallrye-config-json`.  It is very "à la carte" which is nice IMO.

phillip-kruger:
Hi @dmlloyd 
Thanks for the reply. I assume you mean config-converters-json under config-converters for option 2 ? At least for this PR ? In a later PR there will be a json config-source.

I prefer the way it's now (or will be once all PRs is done), which I think is a combination of 2 and 3:
structure

At the end a separate jar is build, so à la carte should be possible.

I do not think option 1 should be done, as these extension are not specific to smallrye implementation, and users should be able to use them in any implementation. What we might discuss at some point is should some of these jars be included in the implementation by default. And should the one or more of the sources already in the implementation be moved here and just included in the implementation by default, as jar dependencies.

dmlloyd:
Oh, right, my bad, I was confusing two things, thinking of the JSON config source idea.  Blame it on a Monday morning. :)

dmlloyd:
But yes a `converters` subdirectory makes a lot of sense to me at least.  I guess the subartifact ought to be `smallrye-config-converter-json` or `...-converters-json` then, to avoid confusion with the (upcoming?) JSON config source?

phillip-kruger:
The artifactId will be **configconverters-json-1.3** and for the upcoming config-source it will be **configsource-json-1.3**. Both under a groupId **io.smallrye.ext**. 

This was done based on feedback on the previous (big, never merged) PR. We can of course discuss more and change it, I do not have too strong opinions about it. I guess as long as it's consistent and in line with the rest of the artifacts.

dmlloyd:
OK well I'm fine with it.  I guess we need to see what @jmesnil and/or @kenfinnigan have to say.

kenfinnigan:
Sorry for taking so long to get back to this.

I'm going to reneg on my original blanked `/extensions` idea for everything.

In this case, and taking into account we will need to pull out converters from the impl, it is probably best to have those under `/converters`.

In terms of groupId and artifactId, I'm almost wondering if we're better off with G of `io.smallrye.config` and A of `smallrye-config-converter-json`. I know, very wordy, but I'm trying to think about the problem of the JAR name being just the artifactId, and not wanting clashes. I've had that issue before and it's a pain.

Having said all this, if we need to bring this back to the SmallRye issue @phillip-kruger raised for the migration in the first place, I'm fine with that.

phillip-kruger:
Hi @kenfinnigan 
Ok cool, I'll change so that it's in  `/converters` (so no more `/extensions/converters`) and I'll change the groupId and artifactId. Should the artifactId still contain the version number ? so `-1.3` at the end ?

kenfinnigan:
Ah yes, it would need the MP version suffix

phillip-kruger:
@kenfinnigan cool thanks. Will update this PR soon with those changes.

phillip-kruger:
@kenfinnigan please check structure change as requested. Let me know. Thanks

kenfinnigan:
@phillip-kruger I think this needs a rebase, as it looks like there's commits in the PR that are already in master?

One other thing, probably a good idea to remove the copyright notices from the source files. 1) because we don't really need them and 2) the copyright for the files would actually be with the original authors from MP Ext.

phillip-kruger:
Closing this PR. Will create a new one. I screwed up my git and can't get it right. ...

PR #135, opened 2019-07-18
Changes for empty values and conversions

dmlloyd:
Also some related misc. small cleanups


dmlloyd:
One side-effect of this change is that a user can call `getConverter` with an array argument and it will work.

PR #146, opened 2019-08-01
unmodified properties in the ConfigSource and small refactoring

ys-kalyakin:
https://github.com/eclipse/microprofile-config/blob/4611c9433675e34ee73bd1eba026e01631582772/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSource.java#L39

> A ConfigSource is always read-only, any potential updates of the configured values must be handled directly inside each ConfigSource.

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/Converters.java
+++ b/implementation/src/main/java/io/smallrye/config/Converters.java
@@ -35,6 +35,7 @@ import java.util.function.IntFunction;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 
+import io.smallrye.config.utils.StringUtil;
 import org.eclipse.microprofile.config.spi.Converter;
 
 /**
--- a/implementation/src/main/java/io/smallrye/config/DirConfigSource.java
+++ b/implementation/src/main/java/io/smallrye/config/DirConfigSource.java
@@ -18,21 +18,25 @@ package io.smallrye.config;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.nio.file.Files;
-import java.util.HashMap;
+import java.nio.file.Path;
+import java.util.Collections;
 import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.eclipse.microprofile.config.spi.ConfigSource;
 import org.jboss.logging.Logger;
+import static io.smallrye.config.utils.ConfigSourceUtil.CONFIG_ORDINAL_100;
+import static io.smallrye.config.utils.ConfigSourceUtil.CONFIG_ORDINAL_KEY;
 
 /**
  * Read configuration from a file directory.
- *
+ * <p>
  * Each file in the directory corresponds to a property where the file name is the property key
  * and the file textual content is the property value.
- *
+ * <p>
  * For example, if a directory structure looks like:
  *
  * <pre>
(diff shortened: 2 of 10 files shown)


implementation/src/main/java/io/smallrye/config/Converters.java
```diff
@@ -35,6 +35,7 @@
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 
+import io.smallrye.config.utils.StringUtil;
```
kenfinnigan: Is this import needed?

implementation/src/main/java/io/smallrye/config/DirConfigSource.java
```diff
@@ -16,23 +16,28 @@
 
 package io.smallrye.config;
 
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.jboss.logging.Logger;
```
kenfinnigan: Was a local build run with these changes, as I don't think the imports should be reordered in the PR?

implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
```diff
@@ -49,22 +51,24 @@ public String getValue(String name) {
             return null;
         }
 
+        final Map<String, String> properties = getProperties();
+
         // exact match
-        String value = AccessController.doPrivileged((PrivilegedAction<String>) () -> System.getenv(name));
```
kenfinnigan: I see this approach results in better re-use, but not sure if it's a good idea to cache the environment properties?

implementation/src/main/java/io/smallrye/config/PropertiesConfigSource.java
```diff
@@ -27,36 +27,41 @@
 
 import org.eclipse.microprofile.config.spi.ConfigSource;
 
+import static io.smallrye.config.utils.ConfigSourceUtil.*;
```
kenfinnigan: Again, not sure if the build was run as we don't want all encompassing static imports

implementation/src/main/java/io/smallrye/config/utils/ConfigSourceUtil.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * utilities and constants for {@link ConfigSource} implementations
+ *
+ * @author helloween
+ */
+public class ConfigSourceUtil {
+    public static final String CONFIG_ORDINAL_KEY = "config_ordinal";
+    public static final String CONFIG_ORDINAL_DEFAULT_VALUE = "100";
```
kenfinnigan: I think the name should include the ordinal value

implementation/src/main/java/io/smallrye/config/Converters.java
ys-kalyakin: yes, class in another package

kenfinnigan: ah

implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
ys-kalyakin: it's not cache, i'ts local copy at the time of the search. i think it's ok

implementation/src/main/java/io/smallrye/config/utils/ConfigSourceUtil.java
ys-kalyakin: CONFIG_ORDINAL_DEFAULT_VALUE_100?

kenfinnigan: Maybe `CONFIG_ORDINAL_100`?

ys-kalyakin: ok

kenfinnigan:
@jmesnil, can you take a look at this on Monday? Thanks

PR #148, opened 2019-08-05
Initital donation from MicroProfile-ext: Config Events

phillip-kruger:
Donation from MicroProfile-ext. This library is used by other config source to fire events on config changes


kenfinnigan:
Currently a compilation failure in CI, but I'm curious as to what the use case for this is given the application code needs to track the properties before/after and then fire the notification.

What are the use cases for these events?

phillip-kruger:
Not sure why the build fails. Is there a way to restart the build ?

phillip-kruger:
W.r.t use cases. Initially when I create this it was for a team that build their own docker cluster and needed a way to notify other services if the config change and to update their own scoped beans.
@derekm - you also had some use-cases ?

kenfinnigan:
Thoughts on this @jmesnil?

kenfinnigan:
@phillip-kruger build failed again, appears to fail when compiling because it can't find Shrinkwrap and other files

phillip-kruger:
Weird. (Working on my PC :) )
I'll have a look.... Thanks.

utils/events/src/test/java/io/smallrye/config/events/ChangeEventNotifierTest.java
```diff
@@ -26,9 +28,14 @@
 
     @Deployment
     public static WebArchive createDeployment() {
+
```
kenfinnigan: Why not SmallRye Config?

Given Thorntail MP Config uses SmallRye Config internally, what does using Thorntail offer?

phillip-kruger: Did not really think about that. Just used code from previous PR to try and see why the build is failing.

phillip-kruger: Seems like it worked and the build is fine now...

jmesnil:
@phillip-kruger I've merged #149 that requires to update `utils/events/pom.xml`

I've the same question than @kenfinnigan about the use of Thorntail MP Config, otherwise it looks good.

phillip-kruger:
Ok I give up with trying to get my PR in sync when another PR happened since. Is this documented somewhere ? I rebase (does not work). I merge (does not work). This feels like it has to be easier ? Or am I just stupid ? Anyway. Closing this PR and starting from scratch. If anyone can point me to some documentation on how to sync properly please do.

PR #152, opened 2019-08-07
Donation from MicroProfile-ext: Config Events

phillip-kruger:

Donation from MicroProfile-ext. Changed test to use smallrye and not TT as requested.

### DIFF
--- a/converters/json/src/test/java/io/smallrye/config/converter/json/JsonArrayConverterTest.java
+++ b/converters/json/src/test/java/io/smallrye/config/converter/json/JsonArrayConverterTest.java
@@ -33,12 +33,14 @@ public class JsonArrayConverterTest {
 
     @Deployment
     public static WebArchive createDeployment() {
-        final File[] thorntailMPConfigFiles = Maven.resolver().resolve("io.thorntail:microprofile-config:2.4.0.Final")
+        final File[] smallryeConfig = Maven.resolver()
+                .loadPomFromFile("pom.xml")
+                .resolve("io.smallrye:smallrye-config")
                 .withoutTransitivity().asFile();
 
         return ShrinkWrap.create(WebArchive.class, "JsonArrayConverterTest.war")
                 .addPackage(JsonArrayConverter.class.getPackage())
-                .addAsLibraries(thorntailMPConfigFiles)
+                .addAsLibraries(smallryeConfig)
                 .addAsResource(
                         new File("src/main/resources/META-INF/services/org.eclipse.microprofile.config.spi.Converter"),
                         "META-INF/services/org.eclipse.microprofile.config.spi.Converter")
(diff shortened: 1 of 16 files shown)


converters/json/src/test/java/io/smallrye/config/converter/json/JsonArrayConverterTest.java
```diff
@@ -33,12 +33,14 @@
 
     @Deployment
     public static WebArchive createDeployment() {
-        final File[] thorntailMPConfigFiles = Maven.resolver().resolve("io.thorntail:microprofile-config:2.4.0.Final")
```
kenfinnigan: Sorry @phillip-kruger, can we have the fix of removing Thorntail and switching to SmallRye Config in a separate PR, as it's unrelated to adding Config Events. Thanks

phillip-kruger: Sorry, yes will do a separate PR. Reverted.

kenfinnigan:
I'm having trouble getting to grips with the use cases for the events.

One thought I had is how does it work if a change event is triggered in the middle of application code using that config value. Should what's happening restart, be halted, does something unexpected? And however it's handled, it's the application developer that needs to handle it?

That's putting a lot of onus on the developer to know exactly what they're doing.

Would appreciate thoughts from others on this.

phillip-kruger:
The event themselves is not doing any changes to the config values. It just notifies that a change has happened.

Some of the still to be donated config sources use this, for example if you make a change to the yaml file if you are using the YamlConfigSource, the event will fire. 

I asked @derekm to also comment here with his use case.

kenfinnigan:
Right, my thoughts are around how an application developer would handle these events, especially for "in progress" work that relies on a value that has changed, and therefore what's the benefit/use case?

Would be great to hear @derekm thoughts

jclingan:
@kenfinnigan  I have had customers ask for this feature in the context of [Spring Cloud Kubernetes](https://cloud.spring.io/spring-cloud-static/spring-cloud-kubernetes/2.1.0.RC1/single/spring-cloud-kubernetes.html#_propertysource_reload). Properties annotated with @RefreshScope are automatically updated, IIRC. Feature toggles is one such use case.  I could also see this as underlying Spring Boot annotation support (like @RefreshScope) for Quarkus **if** we wanted to do this.

In the context of Kubernetes, dynamic property changes is considered an anti-pattern (immutable infrastructure and all).  Not all MicroProfile Config users will be running in Kubernetes.

I don't think this is a risky feature since all this does is notify the developer of property changes and its up to the developer to decide what to do.

derekm:
@kenfinnigan & @phillip-kruger —

I'm sorry for the late reply! I've been quite busy lately! (Just got back from presenting at DEFCON!)

We use ZooKeeper as a command & control system for our event stream processing engines, and we use the config events for signalling to our microservices instances via ZK configuration changes.

Phillip invented a more robust config event system after I explained our more rudimentary config events. Phillip was able to accommodate almost every feature I had on my wishlist for event subscription capabilities.

PR #153, opened 2019-08-07
Refactor Zookeeper Config Source to a different directory

kenfinnigan:


derekm:
@kenfinnigan —

Can we do the same as here for the directory config source in the implementation folder?

PR #155, opened 2019-08-12
Donation from MicroProfile-ext: CDI Providers

phillip-kruger:

This util allow developers to Inject a specific config source or a map of all config sources. It's also used by the (upcoming) MemoryConfigSource.

See the README for details on how this can be used.


utils/cdi-provider/src/main/java/io/smallrye/config/cdi/providers/Name.java
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * The define the name of a config source
+ * 
+ * @author <a href="mailto:phillip.kruger@redhat.com">Phillip Kruger</a>
+ */
+@Qualifier
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.METHOD, ElementType.FIELD })
+public @interface Name {
```
kenfinnigan: Does it make sense to re-use `@Named` instead?

utils/cdi-provider/src/main/java/io/smallrye/config/cdi/providers/ConfigSourceProvider.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.enterprise.inject.spi.InjectionPoint;
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+/**
+ * Making the Config sources available via CDI
+ * 
+ * @author <a href="mailto:phillip.kruger@redhat.com">Phillip Kruger</a>
+ */
```
kenfinnigan: Would it be more efficient if `ConfigSourceProvider` and `ConfigSourceMapProvider` were in a single class?

utils/cdi-provider/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+    
+    <parent>
+        <groupId>io.smallrye</groupId>
+        <artifactId>smallrye-config-parent</artifactId>
+        <version>1.3.10-SNAPSHOT</version>
+        <relativePath>../../</relativePath>
+    </parent>
+    
+    <groupId>io.smallrye.config</groupId>
+    <artifactId>smallrye-config-cdi-provider</artifactId>
```
kenfinnigan: Not sure this artifact name is descriptive enough, but I don't have a good alternative at the moment

utils/cdi-provider/src/main/java/io/smallrye/config/cdi/providers/Name.java
phillip-kruger: It does, but I could not get it right :(

If anyone can help me on how to do that, that would be awesome. The problem I have is that config sources that get @Produces is dynamic. So this works with @Named:

    ```
@Produces
    @Named("SysPropConfigSource")
    public ConfigSource produceConfigSource() {
        return configSourceMap.get("SysPropConfigSource");
    }
```
(So I do an explicit @Produces of a known Config Source)

However, if I look through all config sources, I can not get @Named to work:

```
        @Produces
        @Named
        public ConfigSource produceConfigSource(final InjectionPoint injectionPoint) {
            Set<Annotation> qualifiers = injectionPoint.getQualifiers();
            String name = getName(qualifiers);
            return configSourceMap.get(name);
        } 
```
That is why I created my own Qualifier. but would love to reuse @Named. 
Any syuggestions ?

utils/cdi-provider/src/main/java/io/smallrye/config/cdi/providers/ConfigSourceProvider.java
phillip-kruger: Yes ! I made the change.

utils/cdi-provider/pom.xml
phillip-kruger: Not sure either. maybe just providers (so smallrye-config-providers) ?

phillip-kruger:
> What's the use case for being able to inject ConfigSource or a map of all them?

The upcoming MemoryConfigSource use it. Also, building a GUI to see and potentially change config values. (I have see this before).

utils/cdi-provider/src/main/java/io/smallrye/config/cdi/providers/Name.java
manovotn: @phillip-kruger Your `@Name` has nonbinding value, that means actual value on the `String` will be disregarded for resolution.
That's not the case with `@Named` where this is binding. E.g. if you have an injection points `@Inject @Name("foo") ConfigSource foo1` and `@Inject @Name("bar") ConfigSource bar1` then those can be served by the same producer method which has annotation `@Produces @Name("something")` because the annotation value is not taken into account.
With `@Named` you would need producer for each specific String value. I suppose this is the culprit here, does that make sense?

phillip-kruger: Yes it does ! Thanks ! I'll add some test to make sure this works as expected.

kenfinnigan:
@phillip-kruger are you waiting on me for anything here, or you're just busy?

phillip-kruger:
Hi @kenfinnigan , I just want to add a test case with multiple injections, then we are good to go. Traveling at the moment, will look at this soon

utils/cdi-provider/src/main/java/io/smallrye/config/cdi/providers/Name.java
phillip-kruger: Added some test. It's working as expected, can not use @Named because of above mentioned binding.

utils/cdi-provider/pom.xml
phillip-kruger: Done

phillip-kruger:
Closing this PR. Opening new one with new base.

PR #156, opened 2019-08-22
More converter updates

dmlloyd:
Note that this adds one more TCK exception.


dmlloyd:
Sorry this contains more import sorting.  I don't know why they weren't included in the initial reformat patch(es).

PR #158, opened 2019-08-26
Deprecate ConfigFactory

dmlloyd:
This interface is not really useful anymore.  A custom subclass of `SmallRyeConfig` can be created by subclassing `SmallRyeConfigBuilder`.  A custom implementation of `Config` doesn't need to be created by the SmallRye API.

By ultimately removing this interface, the `io.smallrye.config.SmallRyeConfigBuilder#build` can be made to covariantly return `SmallRyeConfig`.


kenfinnigan:
@jmesnil any views on this?

PR #161, opened 2019-09-12
Add HOCON configuration file format support

### LINKED ISSUE #160: Support for HOCON file format
My configuration files are using [HOCON](https://github.com/lightbend/config/blob/master/HOCON.md) format and [Lightbend's config library](https://github.com/lightbend/config).

It would be awesome to be able to use this file format. It's much nicer than plain `.properties` files.

francois-travais:
Using Lightbend's config library to support HOCON file format. I disabled much of what the lib can do (config file merge...), I limited it to looks at these files only:
- `META-INF/microprofile-config.conf`
- `WEB-INF/classes/META-INF/microprofile-config.conf`

The default ordinal for these files is `60` and `50`, which is lower than their `.properties` counterparts, therefore not overwriting them.

Fixes #160.


francois-travais:
@kenfinnigan I removed all the JSON stuff and references to Lightbend.

francois-travais:
@kenfinnigan I just noticed that an implementation already exists in Microprofile Extensions https://github.com/microprofile-extensions/config-ext/blob/master/configsource-typesafeconfig/README.md

It's not exactly doing the same thing than mine but both could be easily merged. Theirs is following the usual conventions for Lightbend configs regarding file location and default override, while mine is just adding support for HOCON syntax

kenfinnigan:
I think it's fine to have similar but different sources, so long as the naming is clear as to what each does

kenfinnigan:
@francois-travais looks like there's a conflict in `index.adoc`, can you rebase and fix?

Thanks

PR #165, opened 2019-09-18
Donation from MicroProfile-ext: Abstract and Abstract File Sources.

phillip-kruger:

This PR include two abstract sources that forms the base for many source to come.

Abstract - allows you to enable/disable and set the ordinal via config
Abstract File - does many of the functions for json, properties, yaml and xml sources.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,10 +61,16 @@
     <module>implementation</module>
     <module>testsuite</module>
     <module>docs</module>
-    <module>sources/zookeeper</module>
+    <!-- Config Converters -->
     <module>converters/json</module>
+    <!-- Utils -->
     <module>utils/events</module>
     <module>utils/cdi-provider</module>
+    <!-- Config Sources -->
+    <module>sources/abstract</module>
+    <module>sources/abstract-file</module>
+    <module>sources/zookeeper</module>
+    
   </modules>
 
   <dependencyManagement>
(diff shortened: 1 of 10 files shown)


kenfinnigan:
Any comment @dmlloyd or @jmesnil?

dmlloyd:
I guess this would mean we're going all-in on mutable config sources.  Do we want to do that without some kind of formal transaction or snapshot support?  We've been advocating for keeping the spec simple (immutable) but that's not necessarily the "right" answer.

kenfinnigan:
Totally missed that, good catch @dmlloyd 

Sounds like we need to have a wider discussion on the mailing list about reloadable config sources, etc, if only from a SmallRye Config perspective

phillip-kruger:
Hi @dmlloyd . I agree that ideally everyone should run on k8n and a config change should redeploy the application / pod. However many people are starting their journey towards the cloud and containers by starting to refactory their current code and starting to use MP libraries but still deploy to tradition Applications servers. At the moment, you can switch this mutable functionality off using config, giving you the imutable you are looking for, however, if you what this, you can use it. So it's up to the user / developer.

phillip-kruger:
@kenfinnigan @dmlloyd do we have a go/no go on this. If we don't want this we can also remove the events (as events do not make sense if config values can not change) I guess we can also then not have the MemoryConfigSource as this allows for realtime config.

dmlloyd:
To me this is, at least for now, a "no-go".  I think we really need to think up some answer to mutability at a spec level though, and probably soon.

kenfinnigan:
Can you comment @jmesnil

phillip-kruger:
I am closing this PR. Will open a new one with the dynamic stuff moved to a separate artifact.

PR #173, opened 2019-10-21
Add range validation converter convenience variants that accept strings as arguments

dmlloyd:


implementation/src/main/java/io/smallrye/config/Converters.java
```diff
@@ -274,6 +274,22 @@ public static Type getConverterType(Class<?> clazz) {
         return new RangeCheckConverter<>(delegate, minimumValue, inclusive, null, false);
     }
 
+    /**
+     * Get a wrapping converter which verifies that the configuration value is greater than, or optionally equal to,
+     * the given minimum value (in string form).
+     *
+     * @param delegate the delegate converter (must not be {@code null})
+     * @param minimumValue the minimum value (must not be {@code null})
```
jmesnil: Should we add `Objects.requireNonNull` to the code to enforce this requirement?

dmlloyd: I want to use smallrye-common-constraint once the parent pom issue is sorted out.

PR #174, opened 2019-10-22
Donation from MicroProfile-ext: Abstract, File, Dynamic and Yaml.

phillip-kruger:

Re-factored all the dynamic stuff out of the Abstract File Class. When you now use the yaml (and other file based sources to be donated) you have no dynamic option build in.

You can however add a jar that will allow you to enable this. see https://github.com/phillip-kruger/smallrye-config/tree/master/utils/dynamic on how to do that.

Let me know if this is more acceptable.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,11 +61,23 @@
     <module>implementation</module>
     <module>testsuite</module>
     <module>docs</module>
-    <module>sources/hocon</module>
-    <module>sources/zookeeper</module>
+    
+    <!-- Config Converters -->
     <module>converters/json</module>
+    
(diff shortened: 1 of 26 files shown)


gsmet:
@kenfinnigan any progress on this one? We had a Quarkus user asking for in on Twitter.

Note we will also need a way to support https://github.com/quarkusio/quarkus/issues/4568 .

kenfinnigan:
@gsmet hopefully within the next week or two

radcortez:
I believe we should close this for now. A lot was already pulled in #197. If we need something else, we can always return here.

PR #175, opened 2019-10-23
Add TypeSafe Config config source

francois-travais:
This implementation is just of port of [Microprofile config extension](https://github.com/microprofile-extensions/config-ext/tree/master/configsource-typesafeconfig) into SmallRye.

### DIFF
--- a/sources/typesafe-config/src/main/java/io/smallrye/configsource/TypeSafeConfigConfigSource.java
+++ b/sources/typesafe-config/src/main/java/io/smallrye/configsource/TypeSafeConfigConfigSource.java
@@ -1,0 +1,53 @@
+package io.smallrye.configsource;
+
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+public class TypeSafeConfigConfigSource implements ConfigSource, Serializable {
+
+    static final int DEFAULT_ORDINAL = 50;
+
+    private final Map<String, String> properties;
+    private final int ordinal;
+
+    public TypeSafeConfigConfigSource() {
+        final Config config = ConfigFactory.load();
+        this.ordinal = config.hasPath(CONFIG_ORDINAL) ? config.getInt(CONFIG_ORDINAL) : DEFAULT_ORDINAL;
+        this.properties = Collections.unmodifiableMap(config.entrySet().stream()
+                .collect(
+                        Collectors.toMap(Map.Entry::getKey, entry -> config.getString(entry.getKey()))));
+    }
+
+    @Override
+    public Map<String, String> getProperties() {
+        return this.properties;
+    }
+
+    @Override
+    public int getOrdinal() {
+        return this.ordinal;
+    }
+
+    @Override
+    public String getValue(String s) {
+        return this.properties.get(s);
+    }
+
+    @Override
+    public String getName() {
+        return TypeSafeConfigConfigSource.class.getSimpleName();
+    }
+
+    @Override
+    public String toString() {
+        return getName();
+    }
+
+}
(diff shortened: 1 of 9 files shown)


sources/typesafe-config/src/main/java/io/smallrye/configsource/TypeSafeConfigConfigSource.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.stream.Collectors;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+public class TypeSafeConfigConfigSource implements ConfigSource, Serializable {
+
+    static final int DEFAULT_ORDINAL = 50;
```
kenfinnigan: Any particular reason to set the default ordinal to 50?

sources/typesafe-config/src/main/java/io/smallrye/configsource/TypeSafeConfigConfigSource.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private final Map<String, String> properties;
+    private final int ordinal;
+
+    public TypeSafeConfigConfigSource() {
+        final Config config = ConfigFactory.load();
```
kenfinnigan: I'm not very familiar with TypeSafe Config.

Is there any advantage to loading the properties at init vs just keeping the `Config` instance?

sources/typesafe-config/src/main/java/io/smallrye/configsource/TypeSafeConfigConfigSource.java
francois-travais: I've to say it's arbitrary. I just thought it should be lower than 100. If you've a better suggestion, no problem. In any case it can be overwritten in the config.

sources/typesafe-config/src/main/java/io/smallrye/configsource/TypeSafeConfigConfigSource.java
francois-travais: When calling `config.entrySet()`, the config is looping over all the properties. I'm thinking it's better to do it once and then rely on the performance of a `map.getValue()`.

sources/typesafe-config/src/main/java/io/smallrye/configsource/TypeSafeConfigConfigSource.java
phillip-kruger: After #174 you can also extend EnabledConfigSource (that extends SelfConfiguringConfigSource) where you get a way to enable/disable the source and to configure the Ordinal.

francois-travais: @phillip-kruger DO you suggest we should wait for #174 to be merged before merging this PR with the `EnabledConfigSource` in use?

phillip-kruger: No, we can always add that later. Also #174 is not approved yet and that might take a while

dmlloyd:
How does this config source implementation relate to the existing HOCON one?

francois-travais:
The HOCON config source only add support for the HOCON format. Instead of using `META-INF/microprofile-config.properties`, one can use `META-INF/microprofile-config.conf` in HOCON format.
While this config source uses the standard behavior of TypeSafe config, in short: it will look for a `reference.conf`, usually bundled in the library or app jar, and then look for an `application.conf` and merge both. You can find all the details [here](https://github.com/lightbend/config#standard-behavior).

dmlloyd:
OK.  So rather than have both the HOCON and TypeSafe config sources, would it make sense to replace the HOCON one with this one?  Or are there significant non-overlapping use cases for both?

francois-travais:
Personally I would have use for this one only. The rest is up to you (maintainers)

radcortez:
@francois-travais Would it be acceptable for you if we add this behaviour to the current `HoconConfigSource`?

We can the standard TypeSafe Config, and also merge in there the `META-INF/microprofile-config.conf` if one is found. 

I believe having both is redundant.

radcortez:
Closing this PR for now. If this is relevant, please reopen @francois-travais. Thank you.

PR #177, opened 2019-11-07
Ensure that all missing values return Optional.empty()

### LINKED ISSUE #176: 1.3.10 started returning Optional[false] for non existent boolean properties
In 1.3.9, `config.getOptionalValue("NON_EXISTENT_PROPERTY", Boolean.class)`, returns `Optional.empty`

In 1.3.10, the same call started returning `Optional[false]` instead. This is a backward incompatible change in behavior and is not consistent with the behavior for types other than boolean, for which this call still returns `Optional.empty`.

dmlloyd:
Fixes #176


kenfinnigan:
Want a release @dmlloyd?

dmlloyd:
Yeah I guess we'd better release.  Thanks.

kenfinnigan:
Release done, will be in Central in about half an hour

Issue #179, opened 2019-11-08
SmallRyeConfig doesn't support "short" type config


jaikiran:
The MicroProfile config spec[1] states:

> The following Converters are provided by MicroProfile Config by default:
> 
>     boolean and java.lang.Boolean , values for true (case insensitive) "true", "1", "YES", "Y" "ON". Any other value will be interpreted as false
> 
>     byte and java.lang.Byte
> 
>     short and java.lang.Short
> ....
> 

However, when trying out some unrelated thing in Quarkus, injecting a `short` config type ended up with:

```
Caused by: java.lang.IllegalArgumentException: No Converter registered for class short
	at io.smallrye.config.SmallRyeConfig.getConverter(SmallRyeConfig.java:215)
	at io.smallrye.config.SmallRyeConfig.getConverter(SmallRyeConfig.java:203)
	at io.smallrye.config.SmallRyeConfig.convert(SmallRyeConfig.java:198)
	at io.smallrye.config.SmallRyeConfig.getValue(SmallRyeConfig.java:99)
	at io.quarkus.arc.runtime.ConfigBeanCreator.create(ConfigBeanCreator.java:48)
```
[1] https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/converters.asciidoc

dmlloyd:
This was fixed.

PR #180, opened 2019-11-08
issue-179 Add support for "short" type config

jaikiran:
The commit in this PR fixes the issue reported in https://github.com/smallrye/smallrye-config/issues/179

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/Converters.java
+++ b/implementation/src/main/java/io/smallrye/config/Converters.java
@@ -130,6 +130,10 @@ public final class Converters {
         return null;
     });
 
+    @SuppressWarnings("unchecked")
+    static final Converter<Short> SHORT_CONVERTER = BuiltInConverter.of(12,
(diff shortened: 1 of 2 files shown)


implementation/src/main/java/io/smallrye/config/Converters.java
```diff
@@ -130,6 +130,10 @@ private Converters() {
         return null;
     });
 
+    @SuppressWarnings("unchecked")
+    static final Converter<Short> SHORT_CONVERTER = BuiltInConverter.of(12,
```
jmesnil: You need to add the SHORT_CONVERTER for the `12` value in the `readResolve`method  https://github.com/smallrye/smallrye-config/pull/180/files#diff-dd5791f68a75fe49b471b3733ecd25aeR601

jaikiran: Thank you for the review @jmesnil :) I've updated the PR to include that change.

jmesnil:
@jaikiran thanks

PR #181, opened 2019-11-08
Trimming of input string upon conversion to Class, boolean or numeric types

yrodiere:
As discussed [here](https://github.com/quarkusio/quarkus/issues/5305).

I did not add trimming to the built-in converters for `String`, `Character`, or `InetAddress`, since that seemed dangerous.

I did not add any kind of warning when leading/trailing whitespaces are detected. Let me know if you want that, too.

Note that some types were already handling leading/trailing whitespaces correctly (`Double.parse` and `Float.parse` trim the string before parsing, in particular), but most did not (`Integer`, ...).

Please do not merge yet as @dmlloyd is preparing a large pull request. Just let me know when his PR is merged, I will rebase mine.

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/Converters.java
+++ b/implementation/src/main/java/io/smallrye/config/Converters.java
@@ -54,58 +54,60 @@ public final class Converters {
             (Converter & Serializable) value -> value != null && !value.isEmpty() ? value : null);
 
     @SuppressWarnings("unchecked")
-    static final Converter<Boolean> BOOLEAN_CONVERTER = BuiltInConverter.of(1, (Converter & Serializable) value -> {
-        if (value != null && !value.isEmpty()) {
-            return "TRUE".equalsIgnoreCase(value)
-                    || "1".equalsIgnoreCase(value)
-                    || "YES".equalsIgnoreCase(value)
-                    || "Y".equalsIgnoreCase(value)
-                    || "ON".equalsIgnoreCase(value)
-                    || "JA".equalsIgnoreCase(value)
-                    || "J".equalsIgnoreCase(value)
-                    || "SI".equalsIgnoreCase(value)
-                    || "SIM".equalsIgnoreCase(value)
-                    || "OUI".equalsIgnoreCase(value);
-        }
-        return null;
-    });
+    static final Converter<Boolean> BOOLEAN_CONVERTER = BuiltInConverter.of(1, StringCleanup.TRIM,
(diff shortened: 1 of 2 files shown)


implementation/src/main/java/io/smallrye/config/Converters.java
```diff
@@ -562,26 +566,44 @@ public A convert(final String str) {
...
-        private BuiltInConverter(final int id, final Converter<T> function) {
+        private BuiltInConverter(final int id, final StringCleanup stringCleanup, final Converter<T> function) {
             this.id = id;
+            this.stringCleanup = stringCleanup;
             this.function = function;
         }
 
         public T convert(final String value) {
-            return function.convert(value);
+            final String cleanValue;
```
kenfinnigan: Maybe use a ternary operator instead? Then a new variable doesn't need to be allocated

kenfinnigan:
@dmlloyd @jmesnil can you both take a look at this too, thanks

implementation/src/main/java/io/smallrye/config/Converters.java
```diff
@@ -562,26 +566,44 @@ public A convert(final String str) {
...
+
+            return function.convert(cleanValue);
         }
 
         Object writeReplace() {
             return new Ser(id);
         }
     }
 
+    enum StringCleanup {
```
kenfinnigan: I can see that using an Enum vs a Boolean means we can easily expand the behavior in the future.

But I wonder whether we would need anything here beyond "trim" and "don't trim". Can you think of any string manipulation other than trim we might need in the future? If not, I'd lean towards just using a boolean

dmlloyd: It might even make more sense to wrap the converter with a `TrimmingConverter` which could be reused.

kenfinnigan: That's a cool idea, then it can be easily added/removed from converters by us, and used separately by others if desired

dmlloyd: It overlaps with another change I was doing so I'll send a single PR which supercedes this one.  /cc @yrodiere

dmlloyd: Opened #185.  I borrowed a commit from this PR.

yrodiere:
Superseded by #185, which fixes everything I was trying to fix. Closing.

PR #182, opened 2019-11-08
Fixes #178

### LINKED ISSUE #178: Extract ConfigSource implementations from core

kenfinnigan:


sources/abstract-parent/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+    <parent>
+        <artifactId>smallrye-config-parent</artifactId>
+        <groupId>io.smallrye</groupId>
+        <version>1.3.12-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <groupId>io.smallrye.config</groupId>
+    <artifactId>smallrye-config-source-abstract-parent</artifactId>
+    <name>SmallRye: Abstract ConfigSource</name>
```
dmlloyd: Is this maybe a little overzealous?  It makes sense IMO to keep the base classes in the root project, lest the number of artifacts get truly out of hand.

kenfinnigan: How would the base class be used by the extracted config sources, but also have those config sources as dependencies to the main implementation?

dmlloyd: On the issue I suggested using service loader on the implementation's own class loader, with a service type of e.g. `interface DefaultConfigurationSourceProvider extends ConfigurationSourceProvider` to load up the default config source instances (during static init of `SmallRyeConfigBuilder` perhaps).  The broken out config sources would depend on the main implementation then (rather than the reverse), fitting the SPI/provider pattern pretty well.

Also I think we do want the separated config source impls to be able to take advantage of any API extensions offered by the base implementation, so that would be another reason to do it that way.

kenfinnigan: Sorry, I missed that comment on the issue. Will look to do that

kenfinnigan: One thing with this approach, that means anything using SRye Config would need to include the impl and all the other config source artifacts, correct?

Should we rename the current impl to `smallrye-config-core` and then have a `smallrye-config` that bundles the default config sources as dependencies to simplify usage for downstream?

dmlloyd: Maybe we should keep the couple of spec-mandated default classes in the core and only break out the extra ones?

Otherwise having a dependency-aggregating artifact seems reasonable - I don't think we'd want to make a shaded aggregate though.  That could cause conflicts/confusion.

kenfinnigan: Agree on no shading, that's nasty.

So do we aggregate or just leave in core?

dmlloyd: It's a small number of classes; let's leave it in core.

implementation/src/main/java/io/smallrye/config/AbstractConfigSource.java
```diff
@@ -0,0 +1,33 @@
+package io.smallrye.config;
+
+import java.io.Serializable;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+public abstract class AbstractConfigSource implements ConfigSource, Serializable {
```
dmlloyd: Needs a `serialVersionUID`.

implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
```diff
@@ -16,22 +16,20 @@
...
 import java.util.Map;
 import java.util.regex.Pattern;
 
-import org.eclipse.microprofile.config.spi.ConfigSource;
-
 /**
  * @author <a href="http://jmesnil.net/">Jeff Mesnil</a> (c) 2017 Red Hat inc.
  */
-public class EnvConfigSource implements ConfigSource, Serializable {
+public class EnvConfigSource extends AbstractConfigSource {
```
dmlloyd: I think we might have forgotten `serialVersionUID` for all of these.

implementation/src/main/java/io/smallrye/config/AbstractConfigSource.java
```diff
@@ -0,0 +1,33 @@
...
+    private int ordinal;
```
dmlloyd: Making `ordinal` mutable could cause trouble.  It should be final.  That does raise a challenge however for config sources which configure their ordinal from their own data (is that all of them?); it might be worth introducing an intermediate class or classes to support this (e.g. a `MapBasedConfigSource` which accepts a `Map` parameter could call `super(name, Integer.parseInt(map.getOrDefault(...));` and the properties config source could call e.g. `super(propertiesToMap(properties))` or some such).

kenfinnigan:
This morning I realized I messed up and we do need the parent separate.

If we don’t, any config source artifact requires the full implementation, which negates the benefit of config sources being separate as they then can’t be used in other implementations

dmlloyd:
Ah I didn't realize that it was a goal to be able to use these config sources with other implementations. Even the default ones?

dmlloyd:
TBH I'm not sure this is a worthwhile goal. Does it really make sense? What is an example use case?

kenfinnigan:
It's just the parents that would need to be extracted actually. The default ones can still be inside the main impl.

It's a two-part goal. One is because that was the goal of MP Extensions, which we're accepting a donation from, and two it would help drive usage and adoption of SmallRye, possibly with a view to users/implementors switching to the full config impl at some point

dmlloyd:
Okay so then I guess the strategy could be:
* Put the base classes and common utils in one module which only depends on the API
* Implementation contains the built-in default sources, depends on CS base module
* Sources modules depend only on CS base module

kenfinnigan:
That's a good explanation of what I was thinking, thanks

dmlloyd:
:+1: Sorry for being dense!

kenfinnigan:
To be fair, I wasn't really explaining what I was thinking.

So the clarification was good for both of us

kenfinnigan:
Ok, think we're on the right track now

common/src/main/java/io/smallrye/config/utils/ConfigSourceUtil.java
```diff
@@ -44,4 +47,12 @@ private ConfigSourceUtil() {
         return properties.entrySet().stream().collect(Collectors.toMap(e -> String.valueOf(e.getKey()),
                 e -> String.valueOf(e.getValue())));
     }
+
+    public static Map<String, String> urlToMap(URL locationOfProperties) throws IOException {
+        try (InputStream in = locationOfProperties.openStream()) {
+            Properties p = new Properties();
+            p.load(in);
```
dmlloyd: This should explicitly wrap with an `InputStreamReader` for `StandardCharsets.UTF_8` otherwise it'll want Latin-1 input.

kenfinnigan: Fixed, thanks!

kenfinnigan:
Thanks @dmlloyd. Do you think this warrants a bump to 1.4.x?

dmlloyd:
Yes I think that's reasonable.  There are now two artifacts to dependency-manage which is a pretty substantial change.

PR #188, opened 2019-11-15
Minor quality related fixes

kenfinnigan:


implementation/src/main/java/io/smallrye/config/Converters.java
```diff
@@ -406,7 +408,7 @@ public T convert(final String value) {
             if (pattern.matcher(value).matches()) {
                 return delegate.convert(value);
             }
-            throw new IllegalArgumentException("Value does not match pattern " + pattern + " (value was \"" + value + "\")");
+            throw new IllegalArgumentException("Value does not match pattern " + pattern + ORIGINAL_VAL_PREFIX + value + "\")");
```
dmlloyd: Maybe these should be updated to use `String.format` instead?

kenfinnigan: Good point, I was being lazy

PR #189, opened 2019-11-18
Fix Jacoco reporting

kenfinnigan:


implementation/src/main/java/io/smallrye/config/Converters.java
```diff
@@ -408,7 +408,8 @@ public T convert(final String value) {
             if (pattern.matcher(value).matches()) {
                 return delegate.convert(value);
             }
-            throw new IllegalArgumentException(String.format(ILLEGAL_ARGUMENT_EXCEPTION_FORMAT, "Value does not match pattern", pattern, value));
+            throw new IllegalArgumentException(
```
dmlloyd: Looks like that reformat issue snuck in here :)

kenfinnigan: I just pushed the format to master and rebased this

kenfinnigan:
Anyone around to approve?

@jmartisk?

PR #192, opened 2019-11-20
I'll show myself out

dmlloyd:


implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -48,7 +48,7 @@
     static final Comparator<ConfigSource> CONFIG_SOURCE_COMPARATOR = new Comparator<ConfigSource>() {
         @Override
         public int compare(ConfigSource o1, ConfigSource o2) {
-            int res = Long.signum((long) o2.getOrdinal() - (long) o1.getOrdinal());
+            int res = Integer.compare(o2.getOrdinal(), o1.getOrdinal());
```
jmesnil: Is it correct? (do we have a test for config source ordering?)

It seems strange to me that the method parameters for `Comparator#compare` are `o1, o2` but the parameters for `Integer.compare` are `o2, o1`. I'd have expected parameters to be in the same order for both

jmesnil: never mind, the comparison order for `ConfigSource` is determined by the *highest* ordinal so it's normal that parameters order is inverted.

kenfinnigan:
Argh, let me re-add properties I removed for testing Jacoco stuff so this passes

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
dmlloyd: Yes there is a test for it.  I initially had it backwards :)

PR #197, opened 2019-11-28
Introduce YAML configuration source

### LINKED ISSUE #171: Support YAML ConfigSource
This is a follow-up to https://github.com/quarkusio/quarkus/issues/3904. It would be nice to support YAML files as a new ConfigSource

dmlloyd:
Fixes #171; partially replaces #174.

### DIFF
--- a/sources/yaml/src/main/java/io/smallrye/config/source/yaml/YamlConfigSource.java
+++ b/sources/yaml/src/main/java/io/smallrye/config/source/yaml/YamlConfigSource.java
@@ -1,0 +1,191 @@
+package io.smallrye.config.source.yaml;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.yaml.snakeyaml.Yaml;
+
+import io.smallrye.config.MapBackedConfigSource;
+
+/**
+ * Yaml config source
+ *
+ * @author <a href="mailto:phillip.kruger@redhat.com">Phillip Kruger</a>
+ * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
+ */
+public class YamlConfigSource extends MapBackedConfigSource {
+
+    static final int ORDINAL = ConfigSource.DEFAULT_ORDINAL + 10;
+
+    private static final long serialVersionUID = -418186029484956531L;
+
+    public YamlConfigSource(String name, InputStream stream) throws IOException {
+        this(name, stream, ORDINAL);
+    }
+
+    public YamlConfigSource(String name, InputStream stream, int defaultOrdinal) throws IOException {
+        super(name, streamToMap(stream), defaultOrdinal, false);
+    }
+
+    public YamlConfigSource(String name, String str) {
+        this(name, str, ORDINAL);
+    }
+
+    public YamlConfigSource(String name, String str, int defaultOrdinal) {
+        super(name, stringToMap(str), defaultOrdinal, false);
+    }
+
+    @SuppressWarnings("unchecked")
+    private static Map<String, String> streamToMap(InputStream inputStream) throws IOException {
+        final Map<String, Object> yamlInput;
+        try {
+            yamlInput = new Yaml().loadAs(inputStream, HashMap.class);
+            inputStream.close();
+        } catch (Throwable t) {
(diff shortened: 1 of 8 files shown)


sources/yaml/src/main/java/io/smallrye/config/source/yaml/YamlConfigSource.java
```diff
@@ -0,0 +1,191 @@
...
+    @SuppressWarnings("unchecked")
+    private static Map<String, String> streamToMap(InputStream inputStream) throws IOException {
+        final Map<String, Object> yamlInput;
+        try {
+            yamlInput = new Yaml().loadAs(inputStream, HashMap.class);
+            inputStream.close();
+        } catch (Throwable t) {
+            try {
+                inputStream.close();
+            } catch (Throwable t2) {
```
kenfinnigan: I'm not familiar with this pattern, but should `t2` be used?

sources/yaml/src/test/resources/org.eclipse.microprofile.config.spi.ConfigSourceProvider
```diff
@@ -0,0 +1 @@
+io.smallrye.config.source.yaml.YamlConfigSourceProvider
```
kenfinnigan: Should this be in src/main/resources so that users only need to add the dependency to get the ConfigSource?

sources/yaml/src/main/java/io/smallrye/config/source/yaml/YamlConfigSource.java
dmlloyd: It essentially replicates the exception handling used with `try`-with-resources, but in this case the secondary exception is suppressed, not the primary.

sources/yaml/src/test/resources/org.eclipse.microprofile.config.spi.ConfigSourceProvider
dmlloyd: Ah, yes that was the intent.  Will fix it.

sources/yaml/src/main/java/io/smallrye/config/source/yaml/YamlConfigSource.java
dmlloyd: But yes this was typo'd.

sources/yaml/src/test/resources/org.eclipse.microprofile.config.spi.ConfigSourceProvider
kenfinnigan: Sorry, totally missed that it should also be in `/META-INF/services` too

kenfinnigan:
Any idea why the Yaml test is crashing the JVM?

dmlloyd:
Heh, looking now.  It works locally!

dmlloyd:
Is there any way to see the produced files on Semaphore?

dmlloyd:
Could this be related?

```console
[INFO] ------------------------------------------------------------------------
[INFO] Skipping SmallRye: MicroProfile Config Parent
[INFO] This project has been banned from the build due to previous failures.
[INFO] ------------------------------------------------------------------------
```

kenfinnigan:
From https://semaphoreci.com/smallrye/smallrye-config/branches/pull-request-197/builds/3 it looks like there's a "Launch SSH" button that might help

dmlloyd:
I don't have that button, is there an authz step?

kenfinnigan:
Do you have a semaphore account? I can set you up

dmlloyd:
Yeah it's linked to my GitHub.

kenfinnigan:
Ok, you should have an invite

dmlloyd:
Testing now.

dmlloyd:
The error seems related to Jacoco:

```console
# Created at 2019-12-02T17:14:00.513
Error: Could not find or load main class @{jacocoArgLine}

```

kenfinnigan:
Can you reproduce locally with "-Pcoverage"?

Maybe the module needs: https://github.com/smallrye/smallrye-config/blob/master/implementation/pom.xml#L81

dmlloyd:
Yup that seems to have been the case.  Pushed up the correction.

sources/yaml/src/test/resources/org.eclipse.microprofile.config.spi.ConfigSourceProvider
dmlloyd: Oops!

dmlloyd:
OK everything should be fixed now.

PR #212, opened 2019-12-11
Align empty value handling with the programmatic API

### LINKED ISSUE #211: Overwriting Collection Converter not possible / Collections should never be NULL
Hi,

I'm using SmallRye Config 1.5 and my aim is to NEVER EVER return `null` for any Collection. For me this is a bug in the current implementation.

However, according to the spec it should be possible to [overwrite any converter with a custom converter](https://github.com/eclipse/microprofile-config/blob/1.3/spec/src/main/asciidoc/converters.asciidoc#adding-custom-converters).
I was not able to do so for the [CollectionConverter](https://github.com/smallrye/smallrye-config/blob/7111a2ba1d3dcc0b61021b26df1c123a4be52bc6/implementation/src/main/java/io/smallrye/config/Converters.java#L614).

So I think the actual "bug" is located in these 2 positions, where `null` is returned:
https://github.com/smallrye/smallrye-config/blob/7111a2ba1d3dcc0b61021b26df1c123a4be52bc6/implementation/src/main/java/io/smallrye/config/Converters.java#L631
and
https://github.com/smallrye/smallrye-config/blob/7111a2ba1d3dcc0b61021b26df1c123a4be52bc6/implementation/src/main/java/io/smallrye/config/Converters.java#L643

Any advice or (bug-)fix is very much appreciated.

dmlloyd:
Fixes #211

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/inject/ConfigProducerUtil.java
+++ b/implementation/src/main/java/io/smallrye/config/inject/ConfigProducerUtil.java
@@ -42,15 +42,21 @@ public class ConfigProducerUtil {
         if (rawValue == null) {
             rawValue = getDefaultValue(injectionPoint);
         }
+        T converted;
         if (rawValue == null) {
             // convert an empty value
             try {
-                return converter.convert("");
+                converted = converter.convert("");
             } catch (IllegalArgumentException ignored) {
                 throw propertyNotFound(name);
             }
+        } else {
+            converted = converter.convert(rawValue);
         }
-        return converter.convert(rawValue);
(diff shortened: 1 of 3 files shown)


dmlloyd:
@svenhaag this should fix the issue; could you please verify?  Also I'm going to see about adding that unit test as well.

dmlloyd:
Just added the unit test and it still fails ;-|  Going to have to hit the debugger.

dmlloyd:
Fixed, and tests added.  Fortunately it was just a problem in the test!

Issue #218, opened 2020-01-16
NullPointerException when application.yaml is empty


gastaldi:
```
12:35:05,582 ERROR [io.qua.dev.DevModeMain] Failed to start quarkus: java.lang.NullPointerException
	at io.smallrye.config.source.yaml.YamlConfigSource.populateFromMapNode(YamlConfigSource.java:76)
	at io.smallrye.config.source.yaml.YamlConfigSource.yamlInputToMap(YamlConfigSource.java:70)
	at io.smallrye.config.source.yaml.YamlConfigSource.streamToMap(YamlConfigSource.java:58)
	at io.smallrye.config.source.yaml.YamlConfigSource.<init>(YamlConfigSource.java:33)
	at io.quarkus.config.yaml.runtime.ApplicationYamlProvider.getConfigSources(ApplicationYamlProvider.java:36)
	at io.smallrye.config.SmallRyeConfigBuilder.lambda$discoverSources$0(SmallRyeConfigBuilder.java:78)
	at java.lang.Iterable.forEach(Iterable.java:75)
	at io.smallrye.config.SmallRyeConfigBuilder.discoverSources(SmallRyeConfigBuilder.java:78)
	at io.smallrye.config.SmallRyeConfigBuilder.build(SmallRyeConfigBuilder.java:169)
	at io.quarkus.deployment.ExtensionLoader.loadStepsFrom(ExtensionLoader.java:190)
	at io.quarkus.deployment.QuarkusAugmentor.run(QuarkusAugmentor.java:85)
	at io.quarkus.runner.RuntimeRunner.run(RuntimeRunner.java:114)
	at io.quarkus.dev.DevModeMain.doStart(DevModeMain.java:178)
	at io.quarkus.dev.DevModeMain.restartApp(DevModeMain.java:208)
	at io.quarkus.dev.RuntimeUpdatesProcessor.doScan(RuntimeUpdatesProcessor.java:136)
	at io.quarkus.vertx.http.deployment.devmode.VertxHotReplacementSetup$1.handle(VertxHotReplacementSetup.java:51)
	at io.quarkus.vertx.http.deployment.devmode.VertxHotReplacementSetup$1.handle(VertxHotReplacementSetup.java:43)
	at io.vertx.core.impl.ContextImpl.lambda$executeBlocking$2(ContextImpl.java:316)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[...]

PR #220, opened 2020-01-16
NullPointerException when application.yaml is empty #218

udavPit:
Fixes the issue #218: NullPointerException when application.yaml is empty


gastaldi:
Fixes #218

PR #222, opened 2020-01-21
Add SmallRyeConfigBuilder.withSources(Collection)

gastaldi:
This enables a minor improvement on not requiring collections to be transformed into array
eg. https://github.com/quarkusio/quarkus/blob/34aa20e42c47ddc516de9b701f107cc7b0c63b9c/core/runtime/src/main/java/io/quarkus/runtime/configuration/ConfigUtils.java#L68

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
+++ b/implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
@@ -18,6 +18,7 @@ package io.smallrye.config;
 
 import java.lang.reflect.Type;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
@@ -115,6 +116,11 @@ public class SmallRyeConfigBuilder implements ConfigBuilder {
     @Override
     public SmallRyeConfigBuilder withSources(ConfigSource... configSources) {
         Collections.addAll(sources, configSources);
+        return this;
+    }
+
+    public SmallRyeConfigBuilder withSources(Collection<ConfigSource> configSources) {
+        sources.addAll(configSources);
         return this;
     }
 


implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
```diff
@@ -118,6 +119,11 @@ public SmallRyeConfigBuilder withSources(ConfigSource... configSources) {
         return this;
     }
 
+    public SmallRyeConfigBuilder withSources(Collection<ConfigSource> configSources) {
+        sources.addAll(configSources);
```
kenfinnigan: What's the difference between this, and the other method that uses `Collections.addAll(sources, configSources)`?

Wondering if it makes sense for them to be aligned?

gastaldi: There is no difference, it does the same thing but avoids converting to array, which is an unnecessary operation in this case

kenfinnigan: Should the other `withSources()` be modified to use this approach then?

gastaldi: I don't think it makes sense to change the other `withSources` method implementation-wise. There is no need to convert an array into a Collection (just as it doesn't make sense to convert a Collection into an array) just to iterate and add to an internal Set :)

kenfinnigan: It might be because it's late, but I'm not understanding what you're getting at, possibly because I'm explaining it badly.

The method you've added here:

```
    public SmallRyeConfigBuilder withSources(Collection<ConfigSource> configSources) {
        sources.addAll(configSources);
        return this;
    }
```

is different from the existing `withSources()`:
```
    public SmallRyeConfigBuilder withSources(ConfigSource... configSources) {
        Collections.addAll(sources, configSources);
        return this;
    }
```

My question was whether the two methods should have the same implementation?

gastaldi:
They do the same thing. The static `Collections.addAll` (1) takes a
Collection and an array as parameters. The instance method `addAll` (2)
takes another Collection.

1 -
https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#addAll-java.util.Collection-T...-

2-
https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#addAll-java.util.Collection-

Em Ter, 21 de jan de 2020 00:25, Ken Finnigan <notifications@github.com>
escreveu:

implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
gastaldi: The answer is no, because although they are functionally equivalent, the parameters are different

kenfinnigan: Thanks for your patience, clearly it's time for bed!

PR #228, opened 2020-01-24
Null value for configuration property.

radcortez:


dmlloyd:
I think this is a spec error and I raised eclipse/microprofile-config#512 in response.

radcortez:
Ok, thanks!

radcortez:
This was retracted in MP Config in https://github.com/eclipse/microprofile-config/issues/365 and https://github.com/eclipse/microprofile-config/pull/512, so closing the PR without merging.

Issue #231, opened 2020-02-05
FileSystemConfigSource no longer works with ConfigMap


jmesnil:
A regression was detected with smallrye-config 1.4.1 when reading properties for Kubernetes ConfigMap.

Before, [code in DirConfigSource.java](https://github.com/smallrye/smallrye-config/blob/1.3.6/implementation/src/main/java/io/smallrye/config/DirConfigSource.java) worked.
Now, [code in FileSystemConfigSource.java doesn't](https://github.com/smallrye/smallrye-config/blob/1.4.1/sources/file-system/src/main/java/io/smallrye/config/FileSystemConfigSource.java#L77).

It seems related to the change in FileSystemConfigSource that does not take into account that the ConfigMaps uses symlinks.
I've not much exception as the new code swallow the thrown exception and just got a log:

> 23:38:15,980 INFO  [org.wildfly.extension.microprofile.config.smallrye._private] (MSC service thread 1-2) WFLYCONF0003: Use directory for MicroProfile Config Source: /etc/config
> 23:38:15,993 WARN  [io.smallrye.config] (MSC service thread 1-2) Unable to read content from file /etc/config

radcortez:
Fixed in #232.

PR #232, opened 2020-02-05
Do not scan configMap sub directories

jfdenise:
Fix issue: https://github.com/smallrye/smallrye-config/issues/231
Sub directory shouldn't be scanned. Added exception message to the trace.

For example, in openshift, the configMap dir contains an hidden directory that contains the actual files. Files in the configMap dir are links to hidden files. This hidden dir is discovered when scanning, so keys discovered twice and failure occurs.

### DIFF
--- a/sources/file-system/src/main/java/io/smallrye/config/source/file/FileSystemConfigSource.java
+++ b/sources/file-system/src/main/java/io/smallrye/config/source/file/FileSystemConfigSource.java
@@ -77,12 +77,13 @@ public class FileSystemConfigSource extends MapBackedConfigSource {
 
     private static Map<String, String> scan(File directory) {
         if (directory != null && directory.isDirectory()) {
-            try (Stream<Path> stream = Files.walk(directory.toPath())) {
+            try (Stream<Path> stream = Files.walk(directory.toPath(), 1)) {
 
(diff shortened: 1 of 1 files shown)


sources/file-system/src/main/java/io/smallrye/config/source/file/FileSystemConfigSource.java
```diff
@@ -77,12 +77,13 @@ public FileSystemConfigSource(File dir, int ordinal) {
 
     private static Map<String, String> scan(File directory) {
         if (directory != null && directory.isDirectory()) {
-            try (Stream<Path> stream = Files.walk(directory.toPath())) {
+            try (Stream<Path> stream = Files.walk(directory.toPath(), 1)) {
```
jmesnil: there is also a variant to follow symlinks. Should we use it?

java.nio.file.Files#walk(java.nio.file.Path, int, java.nio.file.FileVisitOption...)

jfdenise: I don't think we need to follow symlink, we don't want to enter sub directory. if a file is a symlink, then it will be opened. Perhaps I am missing advanced case with symlinks?

PR #237, opened 2020-02-07
Wrap primitive types before looking up converters for them

### LINKED ISSUE #236: Customer Converter<Integer> is no longer used for int types.
This is a regression.

In smallrye-config 1.4.1, a `Converter<Integer>` was taken into account to convert an injected `int` field.
In smallrye-config 1.6.0, this is no longer the case.

`git bisect` with the test from PR #235 identified commit a52d63b2f19eae1155434abac2911c6138c16993 as the one where regression appeared.

dmlloyd:
Fixes #236.  The test in #235 passes with this PR.


radcortez:
Something is happening with the CI. Checking...

dmlloyd:
Ah this would be because of the release.  You have to manually bump up the version in the coverage POM.

dmlloyd:
Ran into that last time too :slightly_smiling_face:

radcortez:
Yeah, just noticed. Thanks :)

PR #238, opened 2020-02-13
Kubernetes Config/Secrets

lordofthejars:
Hi, mates, I have created a PoC of a `ConfigSource` that loads configuration or secrets from Kubernetes Pod into MicroProfile Config.

Notice three things:

1) I am using other MicrpProfile ConfigSources
2) WatchService to detect changes of the configuration files (K8s updates them on the fly)
3) Send a CDI event if the code is inside a CDI container

This PR is just a first draft to start the discussion if it has the sense to have it here or not and what can be changed.

Moreover, the idea is that the user (ie Quarkus) will implement the `ConfigSourceProvider` and how to configure it.


radcortez:
I'll give it a try. 

First thing that comes to my mind, is tests :D

lordofthejars:
There are tests in the same way as other external sources. In a unit test
fashion.

radcortez:
Hum, I've just looked into CI build and it failed because no tests were executed. Maybe because your test class ends with `TestCase` and not `Test`? Most likely not being picked up by Maven?

lordofthejars:
Funny thing is that there are other tests that ends with this suffix, no
idea.

radcortez:
Well as far as I remember, surefire plugin picks up tests ending in `Test` not sure about `TestCase`.

Hum... the other `TestCase` file seem to be in module with regular `Test` classes, so some actuals tests are run and Maven doesn't fail because no tests were run.

lordofthejars:
I'll rename my test and see

radcortez:
Nvm, just tried it locally and they seem to execute even with `TestCase`.

radcortez:
But in you module, when running mvn verify, the test is not being picked up for some reason.

radcortez:
Got it. It's the surefire configuration in the module.

Ladicek:
Silly question, but I can't resist: what's the point of having yet another config source that reads from a filesystem? If it has more features than the existing filesystem-based config source, does it make sense to add features there?

lordofthejars:
Actually, the great thing is not that it reads from the filesystem. It is that you set the directory where configs stand. Can read YAML, Properties, and filesystem where the key is the filename and value is the content. It is extensible so anyone can add more and the important important thing is that it detects any modification of this file and updates the config. And this is important because in k8s you can for example change a secret and automatically the secret is updated on all pods. So you need to react to this change and update the model accordantly.

PR #241, opened 2020-02-20
Make serialization of implicit converters more flexible and secure

### LINKED ISSUE #240: Serialization of implicit converters could be more flexible
If an implicit converter for a class is serialized, and the source JVM's implicit converter was constructor-based, but the deserializing JVM's version of that class has a static method instead, our serialization representation should be flexible enough to account for that.  We can do this by dropping the `io.smallrye.config.ImplicitConverters.StaticMethodConverter.Serialized` and `io.smallrye.config.ImplicitConverters.ConstructorConverter.Serialized` classes in favor of a single `io.smallrye.config.ImplicitConverters.Serialized` class which stores only the class, and which calls `io.smallrye.config.ImplicitConverters#getConverter(Class)` in its `readResolve`.

This would avoid all the accessibility checks as well as possibly obnoxious compatibility issues that could arise from an otherwise reasonable evolution of a class, which might even be outside of the user's control in some cases.

It might be worthwhile to update the specification to this effect as well.

Compatibility of serialized objects would be impacted by this change.  We could keep the old serialized classes if we want users to continue to be able to deserialize old payloads into a new SmallRye Config, but we cannot go the other way using this approach.

A less elegant but more compatible approach would be to keep the two serialized form classes, and just change their `readResolve` to ignore all arguments other than the class, which is then simply passed in to `getConverter(Class)` as described above.

dmlloyd:
Fixes #240

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/ImplicitConverters.java
+++ b/implementation/src/main/java/io/smallrye/config/ImplicitConverters.java
@@ -15,10 +15,10 @@
  */
 package io.smallrye.config;
 
-import java.io.InvalidObjectException;
 import java.io.ObjectStreamException;
 import java.io.Serializable;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Executable;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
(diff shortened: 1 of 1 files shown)


dmlloyd:
This is one implementation option.  Another option would be to drop the two `Serialized` classes and have just one optimized `Serialized` class which contains only the target class.  But this would make 1.6.2 (or 1.7.0) be serialization-incompatible with 1.6.1 and earlier.

We have broken serialization compatibility in the past, but it seems like we would want to minimize that kind of thing.

Issue #248, opened 2020-03-03
Implement ConfigSource Interceptor Prototype


radcortez:
Discussed here:
https://github.com/eclipse/microprofile-config/issues/526

radcortez:
Implemented in #249

PR #249, opened 2020-03-04
#248 - Prototype work to implement a ConfigSource Interceptor Extension

radcortez:
This is just some prototype work on how the interceptor implementation could look like. In no way this is a final solution, just something to start the discussion.

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
+++ b/implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
@@ -16,14 +16,19 @@
 
 package io.smallrye.config;
 
+import static java.util.Collections.unmodifiableList;
+
 import java.io.Serializable;
 import java.lang.reflect.Type;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Objects;
@@ -31,8 +36,11 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
 import java.util.function.IntFunction;
 import java.util.function.UnaryOperator;
+
+import javax.annotation.Priority;
 
 import org.eclipse.microprofile.config.Config;
 import org.eclipse.microprofile.config.spi.ConfigSource;
@@ -59,11 +67,99 @@ public class SmallRyeConfig implements Config, Serializable {
     private final AtomicReference<List<ConfigSource>> configSourcesRef;
     private final Map<Type, Converter<?>> converters;
     private final Map<Type, Converter<Optional<?>>> optionalConverters = new ConcurrentHashMap<>();
+    private final ConfigSourceInterceptorContext interceptorChain;
 
+    SmallRyeConfig(SmallRyeConfigBuilder builder) {
+        this.configSourcesRef = buildConfigSources(builder);
+        this.interceptorChain = buildInterceptorChain(builder);
+        this.converters = buildConverters(builder);
+    }
+
+    @Deprecated
     protected SmallRyeConfig(List<ConfigSource> configSources, Map<Type, Converter<?>> converters) {
-        configSourcesRef = new AtomicReference<>(Collections.unmodifiableList(configSources));
+        this.configSourcesRef = new AtomicReference<>(Collections.unmodifiableList(configSources));
         this.converters = new ConcurrentHashMap<>(Converters.ALL_CONVERTERS);
         this.converters.putAll(converters);
+        this.interceptorChain = buildInterceptorChain(new SmallRyeConfigBuilder());
+    }
+
+    private AtomicReference<List<ConfigSource>> buildConfigSources(final SmallRyeConfigBuilder builder) {
+        final List<ConfigSource> sourcesToBuild = new ArrayList<>(builder.getSources());
+        if (builder.isAddDiscoveredSources()) {
+            sourcesToBuild.addAll(builder.discoverSources());
+        }
+        if (builder.isAddDefaultSources()) {
+            sourcesToBuild.addAll(builder.getDefaultSources());
+        }
+
+        sourcesToBuild.sort(CONFIG_SOURCE_COMPARATOR);
+        // wrap all
+        final Function<ConfigSource, ConfigSource> sourceWrappersToBuild = builder.getSourceWrappers();
+        final ListIterator<ConfigSource> it = sourcesToBuild.listIterator();
+        while (it.hasNext()) {
+            it.set(sourceWrappersToBuild.apply(it.next()));
+        }
+
+        return new AtomicReference<>(unmodifiableList(sourcesToBuild));
+    }
+
+    private Map<Type, Converter<?>> buildConverters(final SmallRyeConfigBuilder builder) {
+        final Map<Type, SmallRyeConfigBuilder.ConverterWithPriority> convertersToBuild = new HashMap<>(builder.getConverters());
+
+        if (builder.isAddDiscoveredConverters()) {
+            for (Converter converter : builder.discoverConverters()) {
+                Type type = Converters.getConverterType(converter.getClass());
+                if (type == null) {
+                    throw new IllegalStateException(
+                            "Can not add converter " + converter + " that is not parameterized with a type");
+                }
+                SmallRyeConfigBuilder.addConverter(type, converter, convertersToBuild);
+            }
+        }
+
+        final ConcurrentHashMap<Type, Converter<?>> converters = new ConcurrentHashMap<>(Converters.ALL_CONVERTERS);
+        convertersToBuild.forEach(
+                (type, converterWithPriority) -> converters.put(type, converterWithPriority.getConverter()));
+
+        return converters;
+    }
+
+    private ConfigSourceInterceptorContext buildInterceptorChain(final SmallRyeConfigBuilder builder) {
+        final List<ConfigSourceInterceptor> interceptors = new ArrayList<>(builder.getInterceptors());
+        if (builder.isAddDiscoveredInterceptors()) {
+            interceptors.addAll(builder.discoverInterceptors());
+        }
+
+        interceptors.sort((o1, o2) -> {
+            final Integer p1 = Optional.ofNullable(o1.getClass().getAnnotation(Priority.class)).map(Priority::value)
+                    .orElse(100);
+            final Integer p2 = Optional.ofNullable(o2.getClass().getAnnotation(Priority.class)).map(Priority::value)
+                    .orElse(100);
+
+            return Integer.compare(p2, p1);
+        });
+
+        SmallRyeConfigSourceInterceptorContext current = new SmallRyeConfigSourceInterceptorContext(
+                (ConfigSourceInterceptor) (context, name) -> {
+                    for (ConfigSource configSource : getConfigSources()) {
+                        String value = configSource.getValue(name);
+                        if (value != null) {
+                            return ConfigValue.builder()
+                                    .withName(name)
+                                    .withValue(value)
+                                    .withConfigSourceName(configSource.getName())
+                                    .withConfigSourceOrdinal(configSource.getOrdinal())
+                                    .build();
+                        }
+                    }
+                    return null;
+                }, null);
+
+        for (int i = interceptors.size() - 1; i >= 0; i--) {
+            current = new SmallRyeConfigSourceInterceptorContext(interceptors.get(i), current);
+        }
+
+        return current;
     }
 
     // no @Override
@@ -117,13 +213,8 @@ public class SmallRyeConfig implements Config, Serializable {
      * @return the raw value, or {@code null} if no property value was discovered for the given property name
      */
     public String getRawValue(String name) {
-        for (ConfigSource configSource : getConfigSources()) {
-            String value = configSource.getValue(name);
-            if (value != null) {
-                return value;
-            }
-        }
-        return null;
+        final ConfigValue configValue = interceptorChain.proceed(name);
+        return configValue != null ? configValue.getValue() : null;
     }
 
     @Override
@@ -175,7 +266,7 @@ public class SmallRyeConfig implements Config, Serializable {
             newVal = Arrays.asList(oldVal.toArray(new ConfigSource[oldSize + 1]));
             newVal.set(oldSize, configSource);
             newVal.sort(CONFIG_SOURCE_COMPARATOR);
-        } while (!configSourcesRef.compareAndSet(oldVal, Collections.unmodifiableList(newVal)));
+        } while (!configSourcesRef.compareAndSet(oldVal, unmodifiableList(newVal)));
     }
 
     public <T> T convert(String value, Class<T> asType) {
--- a/implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
+++ b/implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
@@ -21,9 +21,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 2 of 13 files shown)


implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -118,14 +131,21 @@ public boolean rawValueEquals(String name, String expected) {
...
+            String value = applyInterceptors(ConfigValue.of(configSource, name, configSource.getValue(name)));
             if (value != null) {
                 return value;
             }
         }
         return null;
     }
 
+    private String applyInterceptors(final ConfigValue configValue) {
+        final SmallRyeConfigSourceInterceptorContext context = new SmallRyeConfigSourceInterceptorContext(
```
dmlloyd: We don't want to construct a new context on every request for a configuration value; there should be a constant, pre-built context for every stage of the interception.  Relatedly, we want to apply the interceptors *before* acquiring the value from the config source.

So the invocation sequence should be something like:

```
-> getRawValue(name) (returns a ConfigValue now) calls `return ctxt0.proceed(name)`
   -> interceptor context 0 calls `return interceptor.getValue(ctxt1, name);`
      -> user interceptor 0 (can return a ConfigValue immediately or call proceed to...)
         -> interceptor context 1 calls `return interceptor.getValue(ctxt2, name);`
            -> user interceptor 1 ...
               -> interceptor context 2 (the last one, which calls...) ...
                  -> our "final" interceptor, which always consists of:
                      for (ConfigSource configSource : getConfigSoruces()) {
                          String value = configSource.getValue(name);
                          if (value != null) return new ConfigValue(name, value, ...);
                      }
                      return null;
```

implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
```diff
@@ -124,6 +127,11 @@ public SmallRyeConfigBuilder withSources(Collection<ConfigSource> configSources)
         return this;
     }
 
+    public SmallRyeConfigBuilder withInterceptors(ConfigSourceInterceptor... interceptors) {
```
dmlloyd: I think we'll also need the service loader based `ConfigSourceInterceptorProvider` that was discussed earlier, though that can come in a subsequent iteration.

implementation/src/main/java/io/smallrye/config/lifecycle/ConfigSourceInterceptor.java
```diff
@@ -0,0 +1,7 @@
+package io.smallrye.config.lifecycle;
+
+import java.io.Serializable;
+
+public interface ConfigSourceInterceptor extends Serializable {
+    String getValue(ConfigSourceInterceptorContext context);
```
dmlloyd: This would change to:

```java
ConfigValue getValue(ConfigSourceInterceptorContext context, String name);
```

which allows an interceptor to aggregate multiple configuration values by calling `context.proceed()` multiple times.

implementation/src/main/java/io/smallrye/config/lifecycle/ConfigSourceInterceptorContext.java
```diff
@@ -0,0 +1,35 @@
...
+
+public interface ConfigSourceInterceptorContext {
+    /**
+     * This will provide the ConfigSourceInterceptor with the lookup value of the key on a ConfigSource.
+     *
+     * Should the ConfigSource be exposed?
+     *
+     * @return a ConfigValue
+     */
+    ConfigValue getConfigValue();
```
dmlloyd: The config value would not be available yet at this point.

implementation/src/main/java/io/smallrye/config/lifecycle/ConfigSourceInterceptorContext.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Lookup a key only into the current or previously evaluated ConfigSources
+     *
+     * @param key the key to lookup.
+     * @return the lookup key value.
+     */
+    String getValue(String key);
```
dmlloyd: This would be dropped.

implementation/src/main/java/io/smallrye/config/lifecycle/ConfigSourceInterceptorContext.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Proceeds to the next interceptor.
+     *
+     * @return the config key lookup value.
+     */
+    String proceed();
```
dmlloyd: This would be dropped.

implementation/src/main/java/io/smallrye/config/lifecycle/ConfigSourceInterceptorContext.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Proceeds to the next interceptor.
+     *
+     * @param value the transformed value.
+     * @return the config key lookup value.
+     */
+    String proceed(String value);
```
dmlloyd: This would change to:
```java
ConfigValue proceed(String name);
```

implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
radcortez: Yes, I didn't bother about it, since it is just boilerplate and would take the focus out of the API.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
radcortez: I think we are considering two different solutions.

My idea was to intercept after the getRawValue (so having the original value from the ConfigSource) and then allow the user to either proceed with logging, validation, etc. or provide a new value, by performing substitution or looking up other configs.

If I understood correctly, your proposal is to do it with the key instead?

implementation/src/main/java/io/smallrye/config/lifecycle/ConfigSourceInterceptorContext.java
radcortez: The idea with this was to plug it in the implementation to lookup additional keys into the previous iterated ConfigSources in the interceptor code. This will allow you to compose keys with the ConfigSource stack.

implementation/src/main/java/io/smallrye/config/lifecycle/ConfigSourceInterceptorContext.java
radcortez: This was just a shorthand to proceed if you didn't want to push a new value. In terms of API, for instance if you are writing the LoggerInterceptor, which you only want to log stuff and don't change anything, it looks weird to me that you have to .process and pass something into the method call.

implementation/src/main/java/io/smallrye/config/lifecycle/ConfigSourceInterceptorContext.java
radcortez: Ok, in here we need to discuss a little further both or approaches and figure out what we want to do. See my previous comment in the applyInterceptors method review.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
dmlloyd: I'm considering the following requirements:

* Minimal API
* Allow for non-allocating or minimally-allocating implementations
* Avoid requiring thread locals or other "hidden" context passing
* Maximize flexibility

To avoid allocating, the interceptor chain has to be static (fixed).  But to have a static interceptor chain, any variable context (i.e. the name) has to be passed in as a thread-local value or a parameter.  Thread local values have implementation overhead in addition to being potentially bug-prone.  So, that essentially leaves what I've proposed above.

Given the above, I didn't notice any benefit to differentiating between the "original" configuration key and "additional" keys.  Furthermore I think there's a potential benefit to allowing interceptors to be able to "short circuit" the configuration value lookup process, returning a value (or lack of value) before any config source work is done.

If an interceptor wants the "original" value (from its perspective), it can call `proceed` with the original key name that was passed in, and acquire it that way.  At that point it can do whatever it wants with it - e.g. transform it, combine it with more values (via `proceed`).  Alternatively, if an interceptor doesn't care about the original value (or maybe it's better to say "the value associated with the original key"), it doesn't have to cause any work to be done to access it.

For example, we can implement Quarkus-style configuration profiles using an interceptor like this:

```java
public class ProfileInterceptor implements ConfigSourceInterceptor {
    private final String profileName = ....;
    public ConfigValue getValue(ConfigSourceInterceptorContext ctxt, String name) {
        ConfigValue v = ctxt.proceed("%" + profileName + "." + name);
        if (v == null) v = ctxt.proceed(name);
        return v;
    }
}
```

If a value is found for the profile-specific key, then the config source does not even need to be queried for the "original" value.  Furthermore, if another interceptor transforms the name some other way, who is to say what the "original" value even is?

radcortez: I always thought about the key being something immutable and that only the value should be allowed to be transformed. I like the Profile sample, on the other hand, I'm not sure about allowing to change the key. There is already a lot of confusion around key loading and key sources and developers trying to understand where the key came from. If for some reason you change the key name dramatically you might not even able to properly map your code, with configs and logs. You need to have very good logging.

So for me, keeping the original values and the transformed values, would at least give me some clear indication that a transformation happen, but I still now the origin and where I can change it.

Anyway, let me prototype it and see how it looks like.

radcortez:
@dmlloyd I've made some changes that should align better which the discussed ideas. Let me know what you think.

Thanks!

radcortez:
Added sample tests on how Logging, Profiles and Variable Replacement could look like if implemented with interceptors.

implementation/src/main/java/io/smallrye/config/lifecycle/ConfigValue.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private String value;
+
+    private ConfigValue(final ConfigSource configSource, final String name, final String value) {
+        this.name = name;
+        this.value = value;
+        this.configSource = configSource;
+    }
+
+    public ConfigSource getConfigSource() {
```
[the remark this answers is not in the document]
radcortez: Yeah I've added mostly for logging purposes at the interceptor level. How about if we just expose ConfigSource#getName? At least developers will know the origin of they config resolution.

implementation/src/main/java/io/smallrye/config/lifecycle/ConfigValue.java
```diff
@@ -0,0 +1,39 @@
+package io.smallrye.config.lifecycle;
```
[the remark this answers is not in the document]
radcortez: Yeah, this was with the other first implementation. The new package was also to make it easier to follow the changes. Since we are now coming to a more final solution, we can just move them back to the main one.

implementation/src/main/java/io/smallrye/config/lifecycle/SmallRyeConfigSourceInterceptorContext.java
```diff
@@ -0,0 +1,18 @@
+package io.smallrye.config.lifecycle;
+
+public class SmallRyeConfigSourceInterceptorContext implements ConfigSourceInterceptorContext {
+    private ConfigSourceInterceptor interceptor;
+    private SmallRyeConfigSourceInterceptorContext next;
+
+    public SmallRyeConfigSourceInterceptorContext(
```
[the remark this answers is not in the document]
radcortez: It was because of the package. Moving them all of the same, we can now change this to be default.

implementation/src/main/java/io/smallrye/config/lifecycle/ConfigValue.java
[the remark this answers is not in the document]
dmlloyd: Yes, name is good.  We're going to need the priority as well (in order to implement config profile as an interceptor, and possibly for some other things as well).

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -57,15 +64,43 @@ public int compare(ConfigSource o1, ConfigSource o2) {
...
+
+    protected SmallRyeConfig(
+            List<ConfigSource> configSources,
+            List<ConfigSourceInterceptor> interceptors,
+            Map<Type, Converter<?>> converters) {
+        configSourcesRef = new AtomicReference<>(unmodifiableList(configSources));
+        this.interceptorChain = buildInterceptorChain(unmodifiableList(interceptors));
         this.converters = new ConcurrentHashMap<>(Converters.ALL_CONVERTERS);
         this.converters.putAll(converters);
     }
```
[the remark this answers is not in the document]
radcortez: Sure. My concern was to keep retro compatibly. Yes. the builder will make it more flexible. It was related with the other call where the constructor is called that I didn't want to change just yet.

implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
```diff
@@ -46,6 +48,7 @@
 
     // sources are not sorted by their ordinals
     private List<ConfigSource> sources = new ArrayList<>();
+    private List<ConfigSourceInterceptor> interceptors = new ArrayList<>();
     private Function<ConfigSource, ConfigSource> sourceWrappers = UnaryOperator.identity();
```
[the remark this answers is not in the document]
radcortez: Sure.

implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
```diff
@@ -216,6 +224,22 @@ protected SmallRyeConfig newConfig(List<ConfigSource> sources, Map<Type, Convert
         return (SmallRyeConfig) factory.newConfig(sources, configConverters);
     }
 
+    // TODO - See if we want to add a new method here to keep the old one for compatibility or change the old one.
```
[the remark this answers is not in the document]
radcortez: Sure.

implementation/src/main/java/io/smallrye/config/lifecycle/ConfigSourceInterceptor.java
```diff
@@ -0,0 +1,7 @@
+package io.smallrye.config.lifecycle;
+
+import java.io.Serializable;
+
+public interface ConfigSourceInterceptor extends Serializable {
+    ConfigValue getValue(ConfigSourceInterceptorContext context, String name);
```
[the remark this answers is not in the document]
radcortez: Sure.

implementation/src/main/java/io/smallrye/config/lifecycle/ConfigValue.java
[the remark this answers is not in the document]
radcortez: Ok, lets expose both ConfigSource#getName and ConfigSource#getOrdinal directly, without exposing the actual ConfigSource instance.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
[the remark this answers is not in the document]
radcortez: I made a huge refactor here. Not sure if this is the proper way, but it seemed to be the most clean way to do it. I've moved all the init code to the Config object and had to expose some stuff in the Builder.

implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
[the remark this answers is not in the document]
radcortez: See above.

dmlloyd:
Looks good but maybe it should be squashed before merging.

radcortez:
Sure. I'll squash it. I'm now just finishing the ServiceLoader and Priority implementation.

dmlloyd:
I'm thinking that the `SmallRyeConfigProviderResolver` should discover interceptors by default, wdyt?

radcortez:
We can add discovery by default. Right now, I was mostly worried about not breaking anything on the current behaviour. Let me see.

radcortez:
BTW, what do you think about providing default interceptors as well? Like some os the ones we discussed? 

Or should this be a separate module with all the interceptors and you have to opt in to have them?

dmlloyd:
I think it makes sense to provide interceptors for things that are either in the spec or that we're proposing for the spec.  But we can do that on the next pass.

radcortez:
Ok, do you think it is good to go as is?

PR #261, opened 2020-03-13
Change getConverter to match proposed MP Config 2.0 API without breaking compatibility

dmlloyd:
Depends on eclipse/microprofile-config#495; don't take out of draft until/unless that gets merged.

### DIFF
--- a/implementation/pom.xml
+++ b/implementation/pom.xml
@@ -78,6 +78,19 @@
           <trimStackTrace>false</trimStackTrace>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.jboss.bridger</groupId>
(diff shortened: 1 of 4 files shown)


radcortez:
Not only we need the merge in MP Config, but also the release of the new artifact. 

What do you think if we create a MP 2.0 branch for breaking changes dependant on the MP Config API binaries?

dmlloyd:
We don't need the release of the new (MP Config 2.0) artifact (unless we want to add the `@Override` annotation which I do not view as important) in order to merge this change; it's binary compatible with both the old and new APIs AFAICT.

radcortez:
Ok!

radcortez:
Merged in branch `mpconfig20`.

PR #262, opened 2020-03-13
ConfigValueConfigSource to retrieve Configuration value location.

radcortez:
This is just a prototype attempt to provide a way to retrieve the Configuration Value location and origin.


implementation/src/main/java/io/smallrye/config/ConfigValue.java
```diff
@@ -3,14 +3,14 @@
 public class ConfigValue {
     private final String name;
     private final String value;
-    private final String configSourceName;
-    private final int configSourceOrdinal;
+    private final String trace;
+    private final ConfigSourceInfo info;
```
dmlloyd: I don't know if I agree with this part.  To me it makes more sense to add fields for each type of information: for example an `int lineNumber` which can be -1 if there is no line. This field could be overloaded if useful from other configuration types - for example a database might store the primary key here if there is a numeric primary key and if it makes sense to do so.  The other information being stored in `trace` could just as well be a part of the config source name.

What I'm concerned about is that we're increasing the number of objects that have to be created every time a configuration property is requested.  Ideally we'd be doing the opposite.

implementation/src/main/java/io/smallrye/config/TracedConfigSource.java
```diff
@@ -0,0 +1,5 @@
+package io.smallrye.config;
+
+public interface TracedConfigSource {
+    String getTrace(final String name);
```
dmlloyd: I don't like that getting a configuration property with its information entails a minimum of two lookups.

implementation/src/main/java/io/smallrye/config/TracedPropertiesConfigSource.java
```diff
@@ -0,0 +1,47 @@
...
+import io.smallrye.config.common.MapBackedConfigSource;
+import io.smallrye.config.common.utils.ConfigSourceUtil;
+
+public class TracedPropertiesConfigSource extends MapBackedConfigSource implements TracedConfigSource {
+    private static final long serialVersionUID = 3969193230016229478L;
+
+    private static final String NAME_PREFIX = "TracedPropertiesConfigSource[source=";
+
+    private final URL url;
+    private final Map<String, Integer> tracedProperties = new HashMap<>();
```
dmlloyd: Using a second Map with boxed Integer objects could add an overhead of as much as 100 or more bytes per configuration property, which doesn't seem great.

implementation/src/main/java/io/smallrye/config/TracedPropertiesConfigSource.java
```diff
@@ -0,0 +1,47 @@
...
+        generateTracingInfo(url);
+    }
+
+    private void generateTracingInfo(URL locationOfProperties) throws IOException {
+        try (LineNumberReader reader = new LineNumberReader(
+                new InputStreamReader(locationOfProperties.openStream(), StandardCharsets.UTF_8))) {
+
+            String line;
+            while ((line = reader.readLine()) != null) {
+                findName(line).ifPresent(name -> tracedProperties.put(name, reader.getLineNumber()));
```
dmlloyd: This is inherently racy... we probably don't want to get the properties two separate times to generate the total set of information.

implementation/src/main/java/io/smallrye/config/TracedPropertiesConfigSource.java
```diff
@@ -0,0 +1,47 @@
...
+            }
+        }
+    }
+
+    private Optional<String> findName(String line) {
+        return getProperties().keySet().stream().filter(line::contains).findFirst();
```
dmlloyd: This linear search introduces a polynomial cost to property loading.  10 keys takes 100 iterations to search; 1000 keys takes 1,000,000 iterations.

implementation/src/main/java/io/smallrye/config/ConfigValue.java
radcortez: Because we don't know all the ConfigSources, I thought that maybe we would want to leave that generic. A `lineNumber` property seems very specific to some ConfigSources, so I would prefer to have something more generic.

implementation/src/main/java/io/smallrye/config/TracedPropertiesConfigSource.java
radcortez: I know, this is not meant to be a final solution, just something to have a look. This was just to have minimal work on implementation, or I would have to rewrite Properties class all together :)

implementation/src/main/java/io/smallrye/config/TracedPropertiesConfigSource.java
radcortez: Again, this was just to have a quick implementation. A final one would require us to write our own Properties object so we can read, parse the value and store the line number at the same time. I'm lazy, just wanted to have an implementation going :)

implementation/src/main/java/io/smallrye/config/TracedPropertiesConfigSource.java
radcortez: See previous comments.

implementation/src/main/java/io/smallrye/config/TracedConfigSource.java
radcortez: We don't have to, once we have accessors directly in the ConfigSource, but this was just an easy way to do it.

Alternative, we can turn this into a wrapper around a ConfigSource, so we can control the getValue method and perform only a single lookup.

radcortez:
Sure. This was not meant to be a final implementation, but just to get the discussion going. For instance, the Properties API is somehow limited, so if we want to load the keys and values, plus keep track of the line number, I believe we need to write our own Properties Loader.

Once that one is in place, then we can actually have a single lookup per key and improve the rest of the code.

Do you think we should do it?

dmlloyd:
> Sure. This was not meant to be a final implementation, but just to get the discussion going. For instance, the Properties API is somehow limited, so if we want to load the keys and values, plus keep track of the line number, I believe we need to write our own Properties Loader.
> 
> Once that one is in place, then we can actually have a single lookup per key and improve the rest of the code.
> 
> Do you think we should do it?

Yes, I think we should.

implementation/src/main/java/io/smallrye/config/ConfigValue.java
dmlloyd: True but the harm in having a `lineNumber` that isn't always used is very low compared to the cost of a generic framework which isn't always used.   OTOH the value in being able to introspect the line number is high even if it isn't always set (for example, an application build system like Quarkus might have a common diagnostic infrastructure including line numbers, and being able to extract line numbers (if present) in this case is very useful to provide consistent reporting).

So while it might seem impure from one perspective, it's quite practical from another.

radcortez:
Ok, I'll work on a custom implementation for Properties that allows us to do this more cleanly.

implementation/src/main/java/io/smallrye/config/ConfigValue.java
radcortez: And how using a specific object just to hold ConfigSource info? Or do you think everything should be flatten out?

dmlloyd: I'm thinking that flat is best in this case.

radcortez:
Ok, I've redone the PR to take into account our discussions.

implementation/src/main/java/io/smallrye/config/ConfigValueConfigSourceInterceptor.java
```diff
@@ -0,0 +1,25 @@
...
+    @Override
+    public ConfigValue getValue(final ConfigSourceInterceptorContext context, final String name) {
+        final ConfigValue configValue = context.proceed(name);
+
+        final String key = configValue.getName();
+        final String value = configValue.getValue();
+        final String configSource = configValue.getConfigSourceName();
+        final int lineNumber = configValue.getLineNumber();
+
+        LOG.infov("The key {0} was loaded from {1} with the value {2} with additional information: {3}", key,
```
dmlloyd: Maybe this text needs updating?  Also FWIW, it's recommended to use `infof` rather than `infov`.

implementation/src/main/java/io/smallrye/config/ConfigValueProperties.java
```diff
@@ -0,0 +1,283 @@
...
+class ConfigValueProperties extends HashMap<String, ConfigValue> {
+    public synchronized void load(Reader reader) throws IOException {
+        load0(new LineReader(reader));
+    }
+
+    public synchronized void load(InputStream inStream) throws IOException {
+        load0(new LineReader(inStream));
+    }
+
+    private void load0(LineReader lr) throws IOException {
```
dmlloyd: This would take a fairly long time to analyze in depth; it sufficies to have test cases for each state though.

implementation/src/main/java/io/smallrye/config/MapBackedConfigValueConfigSource.java
```diff
@@ -0,0 +1,105 @@
...
+                .collect(toMap(Map.Entry::getKey,
+                        new Function<Map.Entry<String, ConfigValue>, ConfigValue>() {
+                            @Override
+                            public ConfigValue apply(
+                                    final Map.Entry<String, ConfigValue> entry) {
+                                final ConfigValue value = entry.getValue();
+                                if (value != null) {
+                                    return value.from()
+                                            .withConfigSourceName(getName())
+                                            .withConfigSourceOrdinal(getOrdinal()).build();
```
dmlloyd: Why not use the original `ConfigValue` directly?

implementation/src/main/java/io/smallrye/config/MapBackedConfigValueConfigSource.java
```diff
@@ -0,0 +1,105 @@
...
+                        }));
+    }
+
+    @Override
+    public Set<String> getPropertyNames() {
+        return properties.keySet();
+    }
+
+    @Override
+    public String getValue(final String propertyName) {
```
dmlloyd: This should be a `default` method on `ConfigValueConfigSource` IMO.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -142,6 +142,14 @@ private ConfigSourceInterceptorContext buildInterceptorChain(final SmallRyeConfi
...
                     for (ConfigSource configSource : getConfigSources()) {
+                        if (configSource instanceof ConfigValueConfigSource) {
+                            ConfigValueConfigSource configValueConfigSource = (ConfigValueConfigSource) configSource;
+                            ConfigValue value = configValueConfigSource.getConfigValue(name);
+                            if (value != null) {
+                                return value;
+                            }
+                        }
+
                         String value = configSource.getValue(name);
```
dmlloyd: If `configSource` is a `ConfigValueConfigSource` then we should never call `getValue(String)` on it.  IOW, this should be in an `else`.

radcortez:
I'm writing the documentation as well :)

radcortez:
@dmlloyd actually just pushed it now. I can also add the javadoc portions.

radcortez:
@dmlloyd Check if this is better now :)

doc/modules/ROOT/pages/interceptors/interceptors.adoc
```diff
@@ -53,6 +53,10 @@ exists.
 The `ConfigValueConfigSourceInterceptor` is not registered by default. You need to register it via the ServiceLoader
 mechanism with your application.
 
+WARNING: The `ConfigValueConfigSourceInterceptor` may log sensitive information present in your configuration, so be
```
dmlloyd: TBH if we're going to do this, we should properly address secrets in a global way instead?

implementation/src/main/java/io/smallrye/config/ConfigValueConfigSourceInterceptor.java
```diff
@@ -1,23 +1,44 @@
...
     private static final Logger LOG = Logger.getLogger("io.smallrye.config");
+    private static final String EXCLUDE_KEYS_PATTERN_KEY = "smallrye.log.exclude.keys.pattern";
+
+    private AtomicBoolean initPattern = new AtomicBoolean(false);
+    private Pattern excludeKeyPattern;
 
     @Override
     public ConfigValue getValue(final ConfigSourceInterceptorContext context, final String name) {
-        final ConfigValue configValue = context.proceed(name);
+        if (!initPattern.get()) {
```
dmlloyd: This is a poor way to provide atomicity.  If this is called from two threads, one could be initializing still while the exclude pattern is still being built.  The second thread could observe `initPattern` being `true` while `excludeKeyPattern` is still `null`.

implementation/src/main/java/io/smallrye/config/ConfigValue.java
```diff
@@ -1,5 +1,17 @@
 package io.smallrye.config;
 
+/**
+ * The ConfigValue is a metadata object that holds additional information after the lookup of a configuration.
+ *
+ * Right now, it is able to hold information like the configuration name, value, the Config Source from where the
```
dmlloyd: JavaDocs use HTML formatting, so you need a `<p>` in between paragraphs.

doc/modules/ROOT/pages/interceptors/interceptors.adoc
radcortez: I believe this is the first time we have something that can output the config value directly.

Yes, maybe we should address that in a more global way, but I think it is out of scope of this PR. On the other hand, I didn't want to leave this completely "open".

implementation/src/main/java/io/smallrye/config/ConfigValueConfigSourceInterceptor.java
radcortez: Ops, you are right.

implementation/src/main/java/io/smallrye/config/ConfigValueConfigSource.java
```diff
@@ -8,9 +8,30 @@
...
     ConfigValue getConfigValue(String propertyName);
 
+    /**
+     * Return the properties in this configuration source as a Map of String and ConfigValue.
+     *
+     * @return a map containing properties of this configuration source
+     */
     Map<String, ConfigValue> getConfigValueProperties();
 
     @Override
```
dmlloyd: There probably should be some JavaDocs explaining how these default methods are implemented.

implementation/src/main/java/io/smallrye/config/ConfigValueConfigSource.java
```diff
@@ -0,0 +1,39 @@
...
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+public interface ConfigValueConfigSource extends ConfigSource {
+    ConfigValue getConfigValue(String propertyName);
+
+    Map<String, ConfigValue> getConfigValueProperties();
+
+    @Override
+    default Map<String, String> getProperties() {
+        return getConfigValueProperties().entrySet()
```
dmlloyd: This is likely to be very expensive... perhaps it would be better to return a map that is a view of the map returned by the delegate.

radcortez:
Ok, so I hope I've fixed all of your comments. Thank you :)

implementation/src/main/java/io/smallrye/config/ConfigValueMapView.java
```diff
@@ -0,0 +1,147 @@
...
+        return configValue != null ? configValue.getValue() : null;
+    }
+
+    @Override
+    public String put(final String key, final String value) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public String remove(final Object key) {
```
dmlloyd: If you extend `AbstractMap`, you don't have to provide method bodies for these mutation operations.

radcortez: But AbstractMap has actual implementations for these. Don't we want to make this read only and mark this methods with UnsupportedOperationException?

dmlloyd: We can; however, the abstract implementations ultimately result in throwing that exception if the underlying collection is not writable.  Code which tries to write to these collections is already erroneous so it doesn't really matter if it misbehaves efficiently or inefficiently.

radcortez: Ok. Let me change that.

radcortez: Actually, thinking a little better about this, I think it becomes somehow confusing. Extending AbstractMap, will perform the operations on the instance class and not the delegate Map view we are exposing. So, if you do something like:

```java
map.size() == 3
map.remove(x) // x exists, but the method returns null
map.size() == 3 and not 2
```
If I understood correctly, your suggestion was to remove these methods and not provide the implementation bodies. To support this properly we would still need to provide the right delegates here. Is that ok?

dmlloyd: If the contract of `Map` and its entry set are correctly implemented, if `x` exists it should throw `UnsupportedOperationException` and if it doesn't exist, `null` is returned.  This is correct behavior AFAICT.  Whether or not `AbstractMap` is used (or `AbstractCollection`/`AbstractSet` in the case of the entry sets), any implementation must obey the contract or incorrect/unexpected behavior will result.

radcortez: Sorry, maybe I didn't explain myself correctly. That is true.

In this case, for our implementation the `.remove` is called on the instance class of `ConfigValueMapView extends AbstractMap` if we don't override it, while for instance `.size` is called on our delegate.

So if we omit the `remove`, `put` ,`putAll` and `clear` operations, they will be invoked on the instance Map (without delegation) they will not give you a `UnsupportedOperationException` but won't perform the expected action either, because they are not acting on the delegate map.

dmlloyd: > So if we omit the `remove`, `put` ,`putAll` and `clear` operations, they will be invoked on the instance Map (without delegation) they will not give you a `UnsupportedOperationException` but won't perform the expected action either, because they are not acting on the delegate map.

This is only partially correct.  Yes, they will be invoked on the instance map without delegation.  But they *will* give you `UnsupportedOperationException`.  Not because the delegate gives it, but because the abstract implementations will result in it.

radcortez: Apparently not. Let me push so we can see it in the CI build.

radcortez: Or maybe I'm forgetting something.

dmlloyd: The only reason it wouldn't work should be if something is not implemented correctly (the entry set perhaps).

radcortez: It is mostly on the input parameters. I'm still thinking about `Collections.unmodifiableX`, which gives you a `UnsupportedOperationException` straight away.

Like this, if you pass an empty map to `putAll`, it will work because nothing is being added and only if there is something to add,`put` is called and then the `UnsupportedOperationException` is thrown.

Same for `remove` if you try to remove something that is not there you get the null and not the `UnsupportedOperationException`.

I would prefer for this to behave as a `Collections.unmodifiableMap`, because it may give you the false sense of these operations to be possible, depending on what you use to call them.

dmlloyd: If one chooses to ignore method contracts, whatever happens is probably their own fault. :-)  But if you want exactly `unmodifiableMap` behavior, why not wrap the instance with `unmodifiableMap`?  It's still substantially less code that way.

radcortez: Yes, I thought about that. It was to avoid yet another wrapper, but if we are fine with that, I'll do it :)

radcortez: Ok, check if everything is ok now :)

implementation/src/main/java/io/smallrye/config/ConfigValueMapView.java
```diff
@@ -0,0 +1,127 @@
...
+        if (keySet == null) {
+            keySet = Collections.unmodifiableSet(delegate.keySet());
+        }
+        return keySet;
+    }
+
+    @Override
+    public Set<Map.Entry<String, String>> entrySet() {
+        if (entrySet == null) {
+            entrySet = Collections.unmodifiableSet(new AbstractSet<Entry<String, String>>() {
```
dmlloyd: If you are using `unmodifiableMap` above, you don't need `unmodifiableSet` here (it does it for you already).

implementation/src/main/java/io/smallrye/config/ConfigValueMapView.java
```diff
@@ -0,0 +1,127 @@
...
+                @Override
+                public Iterator<Entry<String, String>> iterator() {
+                    return new Iterator<Entry<String, String>>() {
+                        private Iterator<Entry<String, ConfigValue>> delegate = ConfigValueMapView.this.delegate.entrySet()
```
dmlloyd: This should be final (or else captured as a local var, which makes it implicitly final).

implementation/src/main/java/io/smallrye/config/ConfigValueMapView.java
```diff
@@ -0,0 +1,127 @@
...
+                }
+            });
+        }
+        return entrySet;
+    }
+
+    @Override
+    public Collection<String> values() {
+        if (values == null) {
+            values = Collections.unmodifiableCollection(new AbstractCollection<String>() {
```
dmlloyd: Same thing here about the wrapper.

implementation/src/main/java/io/smallrye/config/ConfigValue.java
```diff
@@ -62,6 +107,7 @@ public static ConfigValueBuilder builder() {
         private String value;
         private String configSourceName;
         private int configSourceOrdinal;
+        private int lineNumber;
```
dmlloyd: This should probably default to -1 TBH.  If some sources are using this as an index, then 0 might be valid (think database rows or something like that).

radcortez:
@dmlloyd Do you think this is now ready to go in? :)

implementation/src/main/java/io/smallrye/config/ConfigValueConfigSourceInterceptor.java
```diff
@@ -0,0 +1,42 @@
...
+                    configValue.getName(),
+                    configValue.getValue(),
+                    configValue.getConfigSourceName(),
+                    configValue.getLineNumber());
+        }
+
+        return configValue;
+    }
+
+    private void ensureInitPattern(final ConfigSourceInterceptorContext context) {
```
dmlloyd: This is still incorrect.  Other threads can observe a time where the pattern is "".

implementation/src/main/java/io/smallrye/config/ConfigValueConfigSourceInterceptor.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.regex.Pattern;
+
+import javax.annotation.Priority;
+
+import org.jboss.logging.Logger;
+
+@Priority(100)
+public class ConfigValueConfigSourceInterceptor implements ConfigSourceInterceptor {
```
dmlloyd: This interceptor is a logging interceptor and nothing more, so the name should reflect that fact.

implementation/src/main/java/io/smallrye/config/ConfigValueConfigSourceInterceptor.java
```diff
@@ -0,0 +1,42 @@
...
+    private static final String EXCLUDE_KEYS_PATTERN_KEY = "smallrye.log.exclude.keys.pattern";
+
+    private AtomicReference<Pattern> pattern = new AtomicReference<>();
+
+    @Override
+    public ConfigValue getValue(final ConfigSourceInterceptorContext context, final String name) {
+        ensureInitPattern(context);
+
+        final ConfigValue configValue = context.proceed(name);
+        if (!pattern.get().matcher(configValue.getName()).matches()) {
```
dmlloyd: I don't think a pattern is sufficient to determine if a key should be logged.  In Quarkus for example, assembling such a pattern is going to be fairly complex (particularly for a use case such as this one).  Given we want to solve the masking issue, why don't we remove this interceptor (for now) and reintroduce it once masking is solved?

radcortez: Ok. I'll remove this, but keep everything else (it is going to be needed anyway in the future) and we focus on #269

radcortez:
@dmlloyd I've removed the logging interceptor now, which was the main cause of the discussion. 

Everything else is how we discussed and implemented. Rebased the PR. Hope it is good to go now.

radcortez:
Thanks @dmlloyd :)

implementation/src/main/java/io/smallrye/config/ConfigValueConfigSource.java
```diff
@@ -0,0 +1,67 @@
...
+ * Extends the original {@link ConfigSource} to expose methods that return a {@link ConfigValue}. The
+ * {@link ConfigValue} allows to retrieve additional metadata associated with the configuration resolution.
+ * <p>
+ *
+ * This is to work around the limitation from the original {@link ConfigSource}. It exposes everything as plain Strings
+ * and it is not possible to retrieve additional information associated with the Configuration. The
+ * ConfigValueConfigSource tries to make this possible.
+ * <p>
+ *
+ * Ideally, this should move the the MicroProfile Config API, once the concept is well-proven.
```
svenhaag, line 18: typo: move **the** the MicroProfile
..move to the..

PR #266, opened 2020-03-17
#Fixed 198 - Add interceptor to expand expressions in Config Values.

radcortez:


implementation/src/test/java/io/smallrye/config/ExpressionConfigSourceInterceptorTest.java
```diff
@@ -0,0 +1,57 @@
...
+        SmallRyeConfig config = (SmallRyeConfig) buildConfig("my.prop", "1234", "expression", "${${compose}}", "compose",
+                "my.prop");
+
+        final String value = config.getValue("expression", String.class);
+        Assert.assertEquals("1234", value);
+    }
+
+    @Test
+    public void defaultExpression() {
+        SmallRyeConfig config = (SmallRyeConfig) buildConfig("expression", "${my.prop:1234}");
```
dmlloyd: Could you add one more test where the default value is an expression e.g. `${prop1:${prop2}}`?

dmlloyd:
Actually a second comment.  Quarkus needs the ability to disable expansion for its configuration reading code, so we need to carry something over to enable that.

implementation/src/test/java/io/smallrye/config/ExpressionConfigSourceInterceptorTest.java
radcortez: Sure.

radcortez:
Doesn't Quarkus bootstraps different instances of SmallRyeConfig with the Builder? If so, we could achieve this by just registering the required interceptors. For Quarkus configs, we can leave the Expression interceptor out.

Would that work?

dmlloyd:
Yes there are different instances, but we toggle expansion on a per-property basis as the configuration is being read.  Some of the properties have to be preserved unexpanded for delivery to the next stage of execution.

radcortez:
Ok, I see that in Quarkus, that is done with a ThreadLocal. Do we want to do the same here? It is somehow tricky to do it on a per-property basis unless we change the API.

- Should we add a `getValue(String name, boolean expand)`?
- Or some sort of marked in the key / value that signals to skip expansion?

radcortez:
Or even better: add a new method to `ConfigValue` to retrieve the original value without expansion?

dmlloyd:
Yeah I think a thread-local is totally reasonable.  The API for Quarkus is unsophisticated but effective, and it doesn't leak the details of property expansion to any consumer that doesn't care about it.

```java
turn off expansion;
try {
    do things;
} finally {
    restore expansion to previous setting;
}
```

For additional safety it could be done like this:
```java
withExpansionTurnedOff(() -> { do things; });
```

radcortez:
Well, if we thing more generally, maybe other developers or consumer don't want expansion on all properties as well. In that case, we need to provide a more friendly way to enable / disable?

dmlloyd:
That could be generally true of any feature provided by an interceptor, really...

radcortez:
Hehe sure. So you prefer the ThreadLocal approach? Since Quarkus uses SmallRye Config directly, I think we could add the API to retrieve an expanded variable or not, or just expose the accessor we are creating. Its the goal in the end to have it exposed in Config, so you could either retrieve the original value or the expanded value through ConfigValue.

dmlloyd:
I think the thread local/static approach is best (maybe for upstream too).  This way we don't have to explode the Config or ConfigValue API for every possible thing that the user might want to toggle, which could get out of hand quickly.

Using a separated API allows for better separation of concerns AFAICT.

radcortez:
Well that can also be implemented in Quarkus only by its own interceptor to be applied before this one and skipping the chain :)

I think it makes sense for ConfigValue to also hold the original expression. This will allow the user to build custom runtime expressions.

Maybe we could support both. We can add the TheadLocal option, but provide a more friendly way for developers to access values. In fact, we may want to turn ConfigValue into an interface and then anyone is free to build their custom objects with additional metadata if they need.

dmlloyd:
Wouldn't we then need combinatorial versions of all the possible ConfigValue implementations?  What would it end up looking like if there were 4 or 5 interceptors that the user wanted to selectively control?  Would we end up with chains of ConfigValue implementations that have to be unwrapped?  This is the kind of thing I worry about.

If ConfigValue is a class - preferably final - then it will always only be exactly what it is, and users can control cross-cutting behaviors in an independent way in any combination.  It's future-proof; any number of additional interceptor behaviors can be added without affecting the basic contract of ConfigValue or even the core API in any way.

radcortez:
Not exactly. You know that the each interceptor just returns a ConfigValue that obeys to a specific contract. As long as the contract is respected (by implementing the interface), then you only return an implementation of it with each interceptor call. You can chain and return the one you got from the context, or return your own.

implementation/src/main/java/io/smallrye/config/Expressions.java
```diff
@@ -0,0 +1,9 @@
+package io.smallrye.config;
+
+public class Expressions {
+    public static void withoutExpansion(final Runnable action) {
```
radcortez: @dmlloyd What this what you had in mind?

dmlloyd: More or less, yes.  It should be done in a `try`/`finally` though, and if expansion was already disabled it should not disable/enable it again.

```java
if (disable()) try { // returns true if the state was changed to disable
    action.run();
} finally {
    enable();
} else {
    action.run();
}
```

radcortez: Ok, pushed a new one. Check if it is better now. Thank you!

implementation/src/test/java/io/smallrye/config/ExpressionConfigSourceInterceptorTest.java
```diff
@@ -0,0 +1,79 @@
...
+        assertEquals("1234", value);
+    }
+
+    @Test
+    public void defaultComposedExpression() {
+        SmallRyeConfig config = (SmallRyeConfig) buildConfig("expression", "${my.prop:${compose}}", "compose", "1234");
+
+        final String value = config.getValue("expression", String.class);
+        assertEquals("1234", value);
+    }
```
dmlloyd: Last thing (I think!): we need test cases for these cases:

* Property expression including `${foo:}` where `foo` does not exist (should expand to empty string)
* Property expression including `${foo}` where `foo` does not exist (should result in `NoSuchElementException`, preferably one which gives the path to the original key as well (e.g. "Could not expand property `foo` in property `bar` in property `baz`" or something of the sort).

radcortez: Sure!

radcortez: Interesting,

In the case of `${foo:}` we get a `Property foo not found`. This is because our default built in converter for Empty String values returns null.

dmlloyd: That's fine, in this case I'd expect it.  For the test you could do e.g. `123${foo:}456` and ensure that the result is `123456`.

dmlloyd: The expander of course has to treat `null` as empty though...

radcortez: To clarify, the expansion does work and expands to empty string.

radcortez: Ok, new push. Lets see now :)

PR #267, opened 2020-03-23
Antora Documentation.

radcortez:


doc/modules/ROOT/pages/attributes.adoc
```diff
@@ -0,0 +1 @@
+:version: 1.7.1-SNAPSHOT
```
jmartisk: Oh I didn't know you can specify attributes here as well. 
Ok we can either have them here, or move them to the `antora.yml` descriptor, depends on what you prefer.
(Bear in mind that attributes in the `antora.yml` descriptor require Antora 2.3.0-beta.1)

radcortez: Usually, that's how I've done it in the past working with adoc. Most likely, the `antora.yml` might be better. I guess you don't require to import the file on each doc where you want to use it.

PR #270, opened 2020-03-30
Fixes #268. Added interceptor to Relocate configuration properties.

### LINKED ISSUE #268: Fallback configuration property name
Configuration property names change and this requires you to update all of your Config Sources to make sure you have the right name and find the right configuration.

Also, on the implementations side, there are vendor based configurations that may move to the API spec level, effectively changing the configuration namespace, but serving the same purpose. 

For instance in SmallRye JWT, the configurations `smallrye.jwt.token.cookie` and `smallrye.jwt.token.header` are going to be part of MP JWT in the namespace `mp.jwt.token.cookie` and `mp.jwt.token.header`. Now, in the implementation side, we need to support both. 

It could be interesting to support some sort of fallback, that allows you to relocate the namespace of the configuration and falling back to other known names for retro compatibility purposes.

radcortez:
Still missing documentation. Add if we are happy with this approach.

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/RelocateConfigSourceInterceptor.java
+++ b/implementation/src/main/java/io/smallrye/config/RelocateConfigSourceInterceptor.java
@@ -1,0 +1,29 @@
+package io.smallrye.config;
+
+import java.util.Map;
+import java.util.function.Function;
+
+import javax.annotation.Priority;
+
+@Priority(400)
+public class RelocateConfigSourceInterceptor implements ConfigSourceInterceptor {
+    private final Function<String, String> mapping;
+
+    public RelocateConfigSourceInterceptor(final Function<String, String> mapping) {
+        this.mapping = mapping != null ? mapping : Function.identity();
+    }
+
+    public RelocateConfigSourceInterceptor(final Map<String, String> mappings) {
+        this(name -> mappings.getOrDefault(name, name));
+    }
+
+    @Override
+    public ConfigValue getValue(final ConfigSourceInterceptorContext context, final String name) {
+        final String map = mapping.apply(name);
+        ConfigValue configValue = context.proceed(map);
+        if (configValue == null && !name.equals(map)) {
+            configValue = context.proceed(name);
+        }
+        return configValue;
+    }
+}
--- a/doc/modules/ROOT/pages/interceptors/interceptors.adoc
+++ b/doc/modules/ROOT/pages/interceptors/interceptors.adoc
@@ -5,19 +5,19 @@ include::../attributes.adoc[]
 
 = Interceptors
 
-SmallRye Config provides an interceptor chain that allows you to hook into the Configuration Value resolution. This is
+SmallRye Config provides an interceptor chain that hooks into the Configuration Value resolution. This is
 useful to implement features like Property Substitution, Configuration Profiles, or just Logging to find out where
 the Config was loaded from.
 
 == Usage
 
-You can create your interceptor by implementing the
+An interceptor can be created by implementing the
 https://github.com/smallrye/smallrye-config/blob/master/implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptor.java[ConfigSourceInterceptor]
 interface.
 
 The `ConfigSourceInterceptor` has a single method `ConfigValue getValue(ConfigSourceInterceptorContext context, String name)`
-to intercept the resolution of a configuration key. You can use the `ConfigSourceInterceptorContext` to proceed with
-the interceptor chain. The chain can be short-circuited by returning your instance of `ConfigValue`.
+to intercept the resolution of a configuration key. The `ConfigSourceInterceptorContext` is used to proceed with
+the interceptor chain. The chain can be short-circuited by returning an instance of `ConfigValue`.
 
 The `ConfigValue` objects hold information about the key name, value, config source origin and ordinal.
 
@@ -25,10 +25,16 @@ The Interceptor Chain is applied before any Conversion is performed on the Confi
 
 === Registration
 
-You can register an interceptor in the interceptor chain by using the `ServiceLoader` mechanism and provide the
+Registration of an interceptor in the interceptor chain is done via the `ServiceLoader` mechanism and provide the
 implementation classes in a `META-INF/services/io.smallrye.config.ConfigSourceInterceptor` file.
 
-Alternatively, you can register interceptors via the Programmatic API in `SmallRyeConfigBuilder#withInterceptors`.
+Registration can also be done with a `ConfigSourceInterceptorFactory`, via the`ServiceLoader` mechanism
+and provide the implementation class in a `META-INF/services/io.smallrye.config.ConfigSourceInterceptorFactory` file.
+The `ConfigSourceInterceptorFactory` may initialize the interceptor with access to the current chain
+(so it can be used to configure the interceptor and retrieve configuration values) and set the priority.
+
+Alternatively, interceptors may be registered via the Programmatic API in `SmallRyeConfigBuilder#withInterceptors` or
+`SmallRyeConfigBuilder.withInterceptorFactories`.
 
 === Priority
 
@@ -42,6 +48,12 @@ SmallRye Config provides the following built-in Interceptors:
 
 * <<expression-interceptor>>
 * <<profile-interceptor>>
+* <<relocate-interceptor>>
+* <<fallback-interceptor>>
+
+None of the interceptors is registered by default. Registration needs to happen via the ServiceLoader
+mechanism, via the Programmatic API or by calling `SmallRyeConfigBuilder.addDefaultInterceptors`, which adds the
+`ExpressionConfigSourceInterceptor` and the `ProfileConfigSourceInterceptor`.
 
 [[expression-interceptor]]
 === ExpressionConfigSourceInterceptor
@@ -49,7 +61,7 @@ SmallRye Config provides the following built-in Interceptors:
 The `ExpressionConfigSourceInterceptor` provides expression expansion on Configuration Values. An expression string is
 a mix of plain strings and expression segments, which are wrapped by the sequence `${ ... }`.
 
-For instance, if you have the following configuration:
+For instance, the following configuration properties file:
 
 [source,properties]
 ----
@@ -67,17 +79,14 @@ Additionally, the Expression Expansion engine supports the following segments:
 
 If an expression cannot be expanded and no default is supplied a `NoSuchElementException` is thrown.
 
-The `ExpressionConfigSourceInterceptor` is not registered by default. You need to register it via the ServiceLoader
-mechanism with your application.
-
 [[profile-interceptor]]
 === ProfileConfigSourceInterceptor
 
-The `ProfileConfigSourceInterceptor` allows you to have multiple configurations with the same name and select them via
+The `ProfileConfigSourceInterceptor` allows multiple configurations with the same name and selects them via
 a profile property.
 
-To be able to set properties with the same name, you need to prefix each property with `%` followed by the profile name
-and a dot:
+To be able to set properties with the same name, each property needs to be prefixed with `%` followed by the profile
+name and a dot:
 
 [source,properties]
 ----
@@ -85,15 +94,59 @@ my.prop=1234
 %dev.my.prop=5678
 ----
 
-Lookup is always done using the `my.prop` property name. If you want to use the profile `dev`, you need to set the
-configuration `smallrye.config.profile=dev` into any valid ConfigSource.
+Lookup is always done using the `my.prop` property name. To use the profile `dev`, the configuration
+`smallrye.config.profile=dev` has to be set into any valid ConfigSource.
 
 When looking up the property `my.prop` with the `dev` profile active the value is `5678`.
 
-Only one profile can be active in any given time.
+Only one profile can be active at any given time.
 
 ==== Config Priority over Profiles
 
-A ConfigSource with a highest priority, that defines `my.prop` will take priority over another low priority
-ConfigSource that defines `%dev.my.prop`. This allows you to override profiles properties regardless of the active
+A ConfigSource with the highest priority, that defines `my.prop` will take priority over another low priority
+ConfigSource that defines `%dev.my.prop`. This allows overriding profiles properties regardless of the active
 profile.
+
+[[relocate-interceptor]]
+=== RelocateConfigSourceInterceptor
+
+The `RelocateConfigSourceInterceptor` allows relocating a configuration name to another name, by providing a
+transformation function or just a simple key value map.
+
+Consider when a configuration key is renamed, lookup needs to happen on the new name, but also on the old name if the
+ConfigSources are not updated yet. The relocation function gives priority to the new resolved configuration name or
+resolves to the old name if no value is found under the new relocation name.
+
+The following `RelocateConfigSourceInterceptor` can relocate configuration names in the `smallrye.config`
+namespace to the `microprofile.config` namespace:
+
+[source,java]
+----
+new RelocateConfigSourceInterceptor(
+    name -> name.startsWith("smallrye.config") ?
+    name.replaceAll("smallrye\\.config", "microprofile.config") :
+    name));
+----
+
+Relocation can also be done with Expression expansion.
+
+[[fallback-interceptor]]
+=== FallbackConfigSourceInterceptor
+
+The `FallbackConfigSourceInterceptor` allows to fallback to another configuration name, by providing a transformation
+function or just a simple key value map.
+
+Consider when a configuration name does not exist, but there might be another configuration name that the config can
+fallback to provide the same expected behavior. The fallback function is only applied if the original resolved
+configuration name is not found and resolved to the fallback name.
+
+The following `FallbackConfigSourceInterceptor` can fallback configuration names in the `microprofile.config`
+namespace to the `smallrye.config` namespace:
+
+[source,java]
+----
+new FallbackConfigSourceInterceptor(
+    name -> name.startsWith("microprofile.config") ?
+    name.replaceAll("microprofile\\.config", "smallrye.config") :
+    name));
+----
--- a/implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptor.java
+++ b/implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptor.java
@@ -4,15 +4,18 @@ import java.io.Serializable;
 
 /**
  * The ConfigSourceInterceptor allows you to intercept the resolution of a configuration key name before the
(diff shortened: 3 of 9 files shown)


implementation/src/main/java/io/smallrye/config/RelocateConfigSourceInterceptor.java
```diff
@@ -0,0 +1,18 @@
+package io.smallrye.config;
+
+import javax.annotation.Priority;
+
+@Priority(200)
+public class RelocateConfigSourceInterceptor implements ConfigSourceInterceptor {
+    private static final String RELOCATE_PREFIX = "%relocate.";
```
dmlloyd: What would be the use case of this interceptor as written?

radcortez: Yes, added this for us to discuss :)

The idea was to implement relocation similar to what Quarkus did with the profiles, but with the relocation prefix in the value instead.

**Use Case**: I want to relocate / fallback to a different configuration name in case I change my configuration name or another configuration name is introduced with the same business behaviour (#268). Also, I want to be able to prioritise such relocations / fallbacks.

**Possible Implementation**:
Relocation happens when the configuration value has the prefix `%relocate.` followed by the new key name.

**Sample 1: (simple relocate)**
```properties
smallrye.jwt.token.header=%relocate.mp.jwt.token.header
mp.jwt.token.header=Authorization
```

I search a configuration for `smallrye.jwt.token.header` or `mp.jwt.token.header` and get `Authorization` in both.

**Sample 2: (specific config with higher priority)**
```properties priority 1000
smallrye.jwt.token.header=Authorization
```
```properties priority 900
smallrye.jwt.token.header=%relocate.mp.jwt.token.header
mp.jwt.token.header=Cookie
```

I search a configuration for `smallrye.jwt.token.header` and get `Authorization`.

I'm happy to hear other ideas. This seems to be a clean approach, because it will give you control at the ConfigSource plus priority on what you want to relocate and it will be very explicit. Now, if we have a complete namespace change and have to do this for a bunch of keys, we can expand this one and add a regex style configuration.

dmlloyd: First I'll note (for my own future reference) that the relationship with Quarkus profiles is purely related to the usage of `%` (this change does not introduce profiles per se).

> **Possible Implementation**:
> Relocation happens when the configuration value has the prefix `%relocate.` followed by the new key name.

This implementation is functionally identical to having a value which is an expression referencing the new key name.  So this is really just a second way to accomplish this exact thing; I would argue we should stick with just the one implementation.  So (for example) sample 1 could be exactly accomplished by `smallrye.jwt.token.header=${mp.jwt.token.header:}` without introducing a second mechanism.

Going back to the use case:

> **Use Case**: I want to relocate / fallback to a different configuration name in case I change my configuration name or another configuration name is introduced with the same business behaviour (#268). Also, I want to be able to prioritise such relocations / fallbacks.

This can be done as well using a more blanket interception.  I think this use case broadly falls under one of two categories:

* A configuration namespace has relocated to a new namespace
* A single configuration property is been moved to a new location

For the first case, I think a prefix-matching property mapper is effective.  By using interceptors to access properties at both the old and new locations, any prioritization rule that makes sense can be applied; furthermore, it is very memory efficient, not requiring the retention of the complete set of keys with placeholder values.

For the second case, a plain string to string mapping would be effective, similarly via interception.

As a final note: Quarkus has explicitly avoided supporting any kind of relocation of properties, generally speaking, since it is presumed that it would cause some usability difficulties.

radcortez: > First I'll note (for my own future reference) that the relationship with Quarkus profiles is purely related to the usage of `%` (this change does not introduce profiles per se).
> 

I've just said that the notion was inspired by Quarkus profile, but in no way wants to reimplement Quarkus notion of profile.

> > **Possible Implementation**:
> > Relocation happens when the configuration value has the prefix `%relocate.` followed by the new key name.
> 
> This implementation is functionally identical to having a value which is an expression referencing the new key name. So this is really just a second way to accomplish this exact thing; I would argue we should stick with just the one implementation. So (for example) sample 1 could be exactly accomplished by `smallrye.jwt.token.header=${mp.jwt.token.header:}` without introducing a second mechanism.

Well, that was my first proposal, and I thought you were worried about the expressions objects, so I took a more direct approach with no expression involved to retrieve the config directly. If it is acceptable to use expressions for this use case, I'm happy to implement it that way.

> 
> Going back to the use case:
> 
> > **Use Case**: I want to relocate / fallback to a different configuration name in case I change my configuration name or another configuration name is introduced with the same business behaviour (#268). Also, I want to be able to prioritise such relocations / fallbacks.
> 
> This can be done as well using a more blanket interception. I think this use case broadly falls under one of two categories:
> 
> * A configuration namespace has relocated to a new namespace
> * A single configuration property is been moved to a new location
> 
> For the first case, I think a prefix-matching property mapper is effective. By using interceptors to access properties at both the old and new locations, any prioritization rule that makes sense can be applied; furthermore, it is very memory efficient, not requiring the retention of the complete set of keys with placeholder values.
> 
> For the second case, a plain string to string mapping would be effective, similarly via interception.

Ok. But where do you suggest to write these mappings / rules? For me, the obvious choice is to add them as some sort of config, and these requires to choose some kind of discriminator so we don't clash with configs already into place. Or do you think we should not provide a basic implementation to serve the simple use cases and each project / app should implement their own?

> 
> As a final note: Quarkus has explicitly avoided supporting any kind of relocation of properties, generally speaking, since it is presumed that it would cause some usability difficulties.

I understand. But I think this is a clear use case and we are experiencing it now with SmallRye JWT. It is very likely that is going to happen a few more times in the future for other specs, while we keep innovating at the implementations and pushing stuff for the API's. Do you think we should drop this and let projects deal with this as they see fit?

dmlloyd: > > > **Possible Implementation**:
> > > Relocation happens when the configuration value has the prefix `%relocate.` followed by the new key name.
> > 
> > 
> > This implementation is functionally identical to having a value which is an expression referencing the new key name. So this is really just a second way to accomplish this exact thing; I would argue we should stick with just the one implementation. So (for example) sample 1 could be exactly accomplished by `smallrye.jwt.token.header=${mp.jwt.token.header:}` without introducing a second mechanism.
> 
> Well, that was my first proposal, and I thought you were worried about the expressions objects, so I took a more direct approach with no expression involved to retrieve the config directly. If it is acceptable to use expressions for this use case, I'm happy to implement it that way.

I am still concerned about using expression objects for this; this proposed alternative avoids creating expression objects but still requires a multitude of objects on a similar order of magnitude, particularly in the whole-namespace scenario.  Bulk mapping is substantially more memory-efficient.

> > Going back to the use case:
> > > **Use Case**: I want to relocate / fallback to a different configuration name in case I change my configuration name or another configuration name is introduced with the same business behaviour (#268). Also, I want to be able to prioritise such relocations / fallbacks.
> > 
> > 
> > This can be done as well using a more blanket interception. I think this use case broadly falls under one of two categories:
> > 
> > * A configuration namespace has relocated to a new namespace
> > * A single configuration property is been moved to a new location
> > 
> > For the first case, I think a prefix-matching property mapper is effective. By using interceptors to access properties at both the old and new locations, any prioritization rule that makes sense can be applied; furthermore, it is very memory efficient, not requiring the retention of the complete set of keys with placeholder values.
> > For the second case, a plain string to string mapping would be effective, similarly via interception.
> 
> Ok. But where do you suggest to write these mappings / rules? For me, the obvious choice is to add them as some sort of config, and these requires to choose some kind of discriminator so we don't clash with configs already into place. Or do you think we should not provide a basic implementation to serve the simple use cases and each project / app should implement their own?

I don't think config is the right place for such rules.  The remappings in both scenarios above are driven by the configuration consumer, not the end user.  Putting mapping rules in the configuration puts it into the end user's arena.

It makes more sense to me for the configuration consumer (i.e. the framework) to have some way to express that the remappings should be done.  Publishing an interceptor is a convenient way to achieve this, in addition to enabling the more efficient possible implementations I mentioned.

SmallRye Config could enable these cases by creating interceptor classes which implement this functionality.  With what we have today, users could extend these classes with variations that populate the mappings as needed upon construction; or, as a more elegant solution, we could introduce a `ConfigSourceInterceptorProvider` in the same spirit as `ConfigSourceProvider` which can construct and return configured instances of these classes.

> > As a final note: Quarkus has explicitly avoided supporting any kind of relocation of properties, generally speaking, since it is presumed that it would cause some usability difficulties.
> 
> I understand. But I think this is a clear use case and we are experiencing it now with SmallRye JWT. It is very likely that is going to happen a few more times in the future for other specs, while we keep innovating at the implementations and pushing stuff for the API's. Do you think we should drop this and let projects deal with this as they see fit?

I think the use case is valid.  I do think that the projects need to have firsthand control over what gets mapped, since that's strictly in their domain.  Providing convenience interceptors could be an effective way to achieve this without polluting the API.

radcortez: > I am still concerned about using expression objects for this; this proposed alternative avoids creating expression objects but still requires a multitude of objects on a similar order of magnitude, particularly in the whole-namespace scenario. Bulk mapping is substantially more memory-efficient.
> 

This was to cover simple cases. It was not the intent to support the whole namespace scenario relocation. That will need to be implemented as you said with some sort of prefix match provider. 

There is another detail that I haven't mentioned. Again, in the case of SmallRye JWT, there are properties in the `smallrye.jwt` namespace that do not have corresponding equivalents in the `mp.jwt` namespace. Either they don't make sense at the API level, or not yet field proven to move the the MP API. In this situation, we cannot blindly remap all the properties in one namespace into the other.

> 
> I don't think config is the right place for such rules. The remappings in both scenarios above are driven by the configuration consumer, not the end user. Putting mapping rules in the configuration puts it into the end user's arena.
> 
> It makes more sense to me for the configuration consumer (i.e. the framework) to have some way to express that the remappings should be done. Publishing an interceptor is a convenient way to achieve this, in addition to enabling the more efficient possible implementations I mentioned.
> 
> SmallRye Config could enable these cases by creating interceptor classes which implement this functionality. With what we have today, users could extend these classes with variations that populate the mappings as needed upon construction; or, as a more elegant solution, we could introduce a `ConfigSourceInterceptorProvider` in the same spirit as `ConfigSourceProvider` which can construct and return configured instances of these classes.

That is fine. I was planning to introduce the `Provider` to the interceptors as well. But I also wanted to have a simple interceptor that you could use with minimal effort. In this cases, what do you think about introducing a `relocate.properties` for a simple one to one mapping case?

Also,  I think end users should be able to use this if they want to use it.

> 
> I think the use case is valid. I do think that the projects need to have firsthand control over what gets mapped, since that's strictly in their domain. Providing convenience interceptors could be an effective way to achieve this without polluting the API.

Ok. That is what I want to achieve as well. Thanks!

dmlloyd: > > I am still concerned about using expression objects for this; this proposed alternative avoids creating expression objects but still requires a multitude of objects on a similar order of magnitude, particularly in the whole-namespace scenario. Bulk mapping is substantially more memory-efficient.
> 
> This was to cover simple cases. It was not the intent to support the whole namespace scenario relocation. That will need to be implemented as you said with some sort of prefix match provider.
> 
> There is another detail that I haven't mentioned. Again, in the case of SmallRye JWT, there are properties in the `smallrye.jwt` namespace that do not have corresponding equivalents in the `mp.jwt` namespace. Either they don't make sense at the API level, or not yet field proven to move the the MP API. In this situation, we cannot blindly remap all the properties in one namespace into the other.

All the more reason to use an interceptor base class for this purpose: it can implement arbitrarily complex remapping decisions with minimal cost.  Example pseudocode:

```java
if (name.startsWith("prefix1.") && ! name.startsWith("prefix1.subprefix2") && name.endsWith(".remapMe")) { ... }
```

This example approach allocates zero additional objects per name lookup and yet can perform complex decisions.

> > I don't think config is the right place for such rules. The remappings in both scenarios above are driven by the configuration consumer, not the end user. Putting mapping rules in the configuration puts it into the end user's arena.
> > It makes more sense to me for the configuration consumer (i.e. the framework) to have some way to express that the remappings should be done. Publishing an interceptor is a convenient way to achieve this, in addition to enabling the more efficient possible implementations I mentioned.
> > SmallRye Config could enable these cases by creating interceptor classes which implement this functionality. With what we have today, users could extend these classes with variations that populate the mappings as needed upon construction; or, as a more elegant solution, we could introduce a `ConfigSourceInterceptorProvider` in the same spirit as `ConfigSourceProvider` which can construct and return configured instances of these classes.
> 
> That is fine. I was planning to introduce the `Provider` to the interceptors as well. But I also wanted to have a simple interceptor that you could use with minimal effort. In this cases, what do you think about introducing a `relocate.properties` for a simple one to one mapping case?

Again I don't agree with doing it this way: the cost is higher and it's not clear what the benefit is.  We have just introduced a flexible interception framework.  It is best to put this into the users' hands.

> Also, I think end users should be able to use this if they want to use it.

Ultimately you have to decide what use case is being met.  So far everything discussed relates to framework providers, not to end users.  But even in the end user case, they can always produce their own interceptor.

radcortez: > Again I don't agree with doing it this way: the cost is higher and it's not clear what the benefit is. We have just introduced a flexible interception framework. It is best to put this into the users' hands.

It would still be implemented in the interceptor, but the idea was to read the simple mappings from that file, so you didn't have to implement your own interceptor or Provider to init a possible base class.

Goal was to not have to implement anything on SmallRye JWT side (since this is a simple use case). If things become less evident, that sure, we implement the interceptor over there with the expected rules.

Do you think we should provide a way with minimal configuration to handle simple cases? If yes, how does the consumer configures the mappings / rules?

dmlloyd: Once you introduce a file-based API, you've adopted a permanent maintenance burden which is beyond a simple API.  The file needs a syntax, and the syntax needs to be concise and also needs to meet the use case(s), which need to be enumerated.  And when new cases arise, you hope that the syntax can be extended to meet them but that's not a given at all.

An interceptor class could trivially be implemented which accepts a name-mapping `Function` which is completely open-ended and flexible, and also easy to use.  But even then you still have a semantic decision relating to whether the configuration is absolutely relocated or whether the original location is also allowed to have a value.  So that again comes down to use cases.

Perhaps a different approach would be, examine the use cases, and see how difficult it is to meet each use case with *just* an interceptor.  If the resultant interceptor is reasonably simple, then I'd say it's just not worth adding more maintenance surface area.

There needs to be sufficient value provided to justify the cost of the feature.  If the problem is trivial to solve with an interceptor then to me the value is not sufficient.

radcortez: That is fine, I'll add the Provider support so it would be easier for consumers to create their interceptors and provide a base mapper interceptor as well.

I think also my main concern is to provide something that users can see how it works (either by a base class and / or documentation) so they know that they can do it in that way.

dmlloyd: Maybe we need some example subprojects?

radcortez: Yes, that already come to my mind multiple times.

The issue is where? Most consumers are not even aware of SmallRye (they know Quarkus or Wildfly). So if you add the samples here, they have a very low chance of being found (you have to promote them, blog them, etc). 

I guess that if we link them from either Wildfly / Quarkus to here it helps. 

What do you think?

dmlloyd: It sounds like examples are nice to have but not likely to be worth the investment (at present).  Maybe that will change in the future.

radcortez: If we do the base class, plus add it to the documentation, at least that will be something we can point people to.

radcortez:
@dmlloyd Let me know if this is more in line with what you had in mind.

I've added 2 interceptors. One for when you just want to relocate a namespace (basically giving priority to the new namespace), and another to fallback (give priority to the original name but fallback to another name).

implementation/src/main/java/io/smallrye/config/RelocateConfigSourceInterceptor.java
```diff
@@ -0,0 +1,29 @@
+package io.smallrye.config;
+
+import java.util.Map;
+import java.util.function.Function;
+
+import javax.annotation.Priority;
+
+@Priority(300)
+public class RelocateConfigSourceInterceptor implements ConfigSourceInterceptor {
```
dmlloyd, line 9: Since these interceptors don't do anything unless they're manually configured, does it make sense to have them be added by default or discovered?  Maybe we need a defaulted `getPriority()` method on `ConfigSourceInterceptor`.

radcortez: I think it makes sense to have a method to set the priority as well like `ConfigSource`. 

BTW, these interceptors are not added by default or get discovered either. You always have to manually configure them.

radcortez: Hum, thinking a little better about this. The changes we made in https://github.com/smallrye/smallrye-config/pull/275 with the `io.smallrye.config.ConfigSourceInterceptorFactory` require us to have the interceptors priority before construction time to order the initialization oder. If we add a method to provide the priority, then we have no way to know the right priority before initialization.

dmlloyd: True, that presents a challenge.  Perhaps the priority would be set by the factory?

radcortez: Yes, I see no other way. We already do that indirectly with the method I added to retrieve the interceptor class.

radcortez: Ok, I've implemented this in https://github.com/smallrye/smallrye-config/pull/275, since I already had the support for the factory and the Interceptor with priority in the builder.

We need to merge https://github.com/smallrye/smallrye-config/pull/275 first so I can clean this up with the work done over there.

radcortez:
Cleaned this up with the changes done in #275.

radcortez:
Thanks @dmlloyd :)

PR #272, opened 2020-03-31
Fixes #271 - Moved CDI related code to its own module.

### LINKED ISSUE #271: Separate CDI Layer from Implementation
An initial thought is that the CDI layer would retain the current artifact name, so as not to break any downstream consumers, and then the pure Java implementation would be in a new artifact named {currentName]-core

The CDI artifact would then depend on the "-core" artifact for the bulk of the implementation and only contain the CDI specific pieces.

radcortez:
The CDI related code was already it a separate package with no relation to the actual implementation code. 

Moved the CDI code to it own module with the current name `smallrye-config` and renamed the implementation module to be `smallrye-config-core`. This should ensure that consumers don't break.

I've tried this with Quarkus and it passed the MP Config TCK. Most likely, when this is released, Quarkus can change to only depend on `smallrye-config-core`.


cdi/pom.xml
```diff
@@ -0,0 +1,87 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+ ~  Copyright 2017 Red Hat, Inc.
```
svenhaag, line 4: 2017?!

cdi/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.smallrye.config</groupId>
+    <artifactId>smallrye-config-parent</artifactId>
+    <version>1.7.1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>smallrye-config</artifactId>
```
svenhaag, line 28: shouldn't it be "smallrye-config-cdi" for instance?

cdi/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+ ~  limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.smallrye.config</groupId>
+    <artifactId>smallrye-config-parent</artifactId>
+    <version>1.7.1-SNAPSHOT</version>
```
svenhaag, line 25: bump to 1.9 ?

cdi/pom.xml
svenhaag: Ah ok, so you basically assume most of the users out there are using CDI anyway?! I would find it cleaner if CDI enabling would be a separate dependency. I think its like that for most libraries?! Cheers!

cdi/pom.xml
svenhaag: Why waiting? If you wait you have to do a pull merge again 😄.

radcortez: It is not a breaking change, but it is a change in the structure, so I was thinking that this should go into a major release. I'll probably do a minor release with what we have in master and then maybe we can do this. Most likely the MP Config 2.0 API is still a couple of months away.

Is this something useful for you?

cdi/pom.xml
radcortez: Well not exactly. We don't really have an idea on who is using or not since everything is bundled together. 

The idea here was to keep the same artifact name for the CDI dependency (which will depend on the standalone implementation, with a new artifact name), so it won't break and if you want to use the non-CDI dependency, you just change it to the new one.

So right now we have:

artifact `smallrye-config` (both CDI / Implementation)

Move to:

artifact `smallrye-core` (Implementation Only)
artifact `smallrye-config` (CDI only code -> depends on smallrye-core)

So this won't break anyone and if you want to use the implementation only without CDI you change the dependency from `smallrye-config` to `smallrye-core`. Does that make sense?

cdi/pom.xml
svenhaag: Absolutely, as I intend to provide my own CDI implementation based on SmallRyes. The variant will include the option to allow null values instead of the NoSuchElement exception. Currently I would only be able to overwrite the producers via the CDI Specializes annotation, but the AfterDeploymentValidation strikes before that.

radcortez: Ok, lets try to push it fast then :)

radcortez: Bumped to 1.8.2-SNAPSHOT, so we can merge it to master and I'll update everything together on release to 1.9.

cdi/pom.xml
radcortez: @svenhaag let me know if this is ok with you.

svenhaag: Hi @radcortez you are awesome! That's just fine. Thank you very much.

radcortez:
Ok, I'll need to check if this works fine with Quarkus. If I don't find any issue, I think we should be good to go.

svenhaag:
Currently the io.smallrye.config.inject.ConfigExtension has a public constructor but only package private @Observer methods. However I'd like to inherit the class or utilize it in my own CDI extension. The methods are currently not used anywhere else so I either would expect them to be private or public (API). Does that make sense for you? Cheers

radcortez:
Nobody can could these methods directly, since they are observer methods from CDI, so not much sense to make them public. Maybe we can make them protected.

What are you looking to inherit and change exactly?

svenhaag:
> What are you looking to inherit and change exactly?

I'd like to provide my own producers so I guess extending on `ConfigProducer` would do the trick.
Same goes for the `ConfigExtension` where I have to overwrite `validate(AfterDeploymentValidation)`.
However, overwriting these classes allone would not help, as the CDI `Extension` SPI still points to the SmallRye implementation. So a cleaner option would be to call another SPI, e.g. `io.smallrye.config.CdiValidation` that allows customization. Anyway, that would be another issue, though.

radcortez:
Depending on you want to do, your own extension can add the same methods and they will run twice, since they are just observer methods in CDI. That should allow you to add additional producers, and additional validations, unless of course you don't really want any of the behaviour from the CDI Extension and in that case, the split and implementing your own is the way to go.

We could make the methods protected, so you can extend and override. Not sure how useful that may be to you. There is some state being collected and use across methods (injection points), so most likely we would need to expose a getter to retrieve the injection points as well.

svenhaag:
Yeah exactly that would do the trick (protected methods and getter for the injection points). I think with this I could at least utilize your classes instead of forking the whole CDI module. Cheers.

radcortez:
@svenhaag I've changed the `ConfigExtension` to be easily extended.

Most likely, that would not be enough, because the jar still packs the service loader file, so by including this dependency this Extension will always run.

svenhaag:
> @svenhaag I've changed the `ConfigExtension` to be easily extended.
> 
> Most likely, that would not be enough, because the jar still packs the service loader file, so by including this dependency the Extension will always run.

This is awesome! I'm back at work next week. I'll try then to utilize it.
Would it also make sense to you to move the `SecuritySupport.java` to `smallrye-config-common`?

radcortez:
> Would it also make sense to you to move the `SecuritySupport.java` to `smallrye-config-common`?

Not so sure about that one :)

We would need to make the class and methods `public`. We cannot use the same package to keep it package protected, since in Java 9, split packages are disallowed.

svenhaag:
> We would need to make the class and methods `public`. We cannot use the same package to keep it package protected, since in Java 9, split packages are disallowed.

Indeed, can we have at least the SecuritySupport methods `protected` too? That would be great. Thanks.

radcortez:
> > We would need to make the class and methods `public`. We cannot use the same package to keep it package protected, since in Java 9, split packages are disallowed.
> 
> Indeed, can we have at least the SecuritySupport methods `protected` too? That would be great.
> Same goes for `ConfigProducer` methods actually - especially `#getConfig()`. Thanks.

Hum, not sure if it is going to be useful. The `ConfigProducer` just wraps methods to produce the required instances via `ConfigProducerUtil`. Extending it it would not help that much, since you do need to override `io.smallrye.config.inject.ConfigExtension#beforeBeanDiscovery` to not register the `ConfigProducer` and add your own.

svenhaag:
> The `ConfigProducer` just wraps methods to produce the required instances via `ConfigProducerUtil`. Extending it it would not help that much, since you do need to override `io.smallrye.config.inject.ConfigExtension#beforeBeanDiscovery` to not register the `ConfigProducer` and add your own.

I can inherit the ConfigProducer and use CDIs @Specializes to overwrite only certain producer methods :P.

radcortez:
> I can inherit the ConfigProducer and use CDIs @Specializes to overwrite only certain producer methods :P.

Sure. You are not going to save that much :P

`ConfigProducer` methods are now protected. Give it a try and see if this is ok now.

radcortez:
@svenhaag did you gave this a try?

svenhaag:
> @svenhaag did you gave this a try?

Indeed, works like a charm so far. Thanks!

svenhaag:
> @svenhaag did you gave this a try?

Huh, now I'm missing ConfigExtension#getInjectionPoints and ConfigProducerUtil#getConfigKey should be public too.

svenhaag:
Dang, it's not working at all :(. Of course if I do
`public class ConfigProducer extends io.smallrye.config.inject.ConfigProducer` and do not provide the dependency a NoClassDefFoundError occurs and I cannot provide the dependency because of the SPI which would load the SmallRye variant. Bummer.

svenhaag:
> 
> 
> @svenhaag did you gave this a try?

It works so far, however I have to use the `maven-shade-plugin` to add the SmallRye classes I don't want to overwrite or utilize. That is something to keep in mind. Cheers

radcortez:
Ok. I've changed the visibility of those methods. Let me know if you need anything else.

radcortez:
@svenhaag I've just moved the relevant commits that provide the extension points to master. I hope this is enough for you.

I think the CDI code shift to another module should be done only in 2.0.

radcortez:
Closing this to apply it in the mpconfig2.0 branch.

PR #275, opened 2020-04-07
Fixes #274. Added interceptor to activate a Configuration Profile.

### LINKED ISSUE #274: Support Configuration Profiles

radcortez:


implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptor.java
```diff
@@ -18,6 +20,15 @@
  * order may be non deterministic.
  */
 public interface ConfigSourceInterceptor extends Serializable {
+    /**
+     * This method initializes the interceptor, with access to the current Config object. This allows to
+     * configure the interceptor using configurations retrieved from the Config.
+     *
+     * @param config the current {@code Config} associated to the interceptor.
+     */
+    default void initialize(Config config) {
```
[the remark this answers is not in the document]
radcortez: Sure. The main idea here was to give a way to configure the interceptor with configs, without having to write some init block at the proceed method (when you have access to the config).

implementation/src/main/java/io/smallrye/config/ProfileConfigSourceInterceptor.java
```diff
@@ -0,0 +1,39 @@
+package io.smallrye.config;
+
+import javax.annotation.Priority;
+
+import org.eclipse.microprofile.config.Config;
+
+@Priority(600)
+public class ProfileConfigSourceInterceptor implements ConfigSourceInterceptor {
+    public static final String SMALLRYE_PROFILE = "smallrye.config.profile";
```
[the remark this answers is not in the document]
radcortez: Hum, isn't it risky? `config.profile` is somehow generic and we may clash with something already out there.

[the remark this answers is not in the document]
dmlloyd: That could be fine, especially if this were a constructor parameter.

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptor.java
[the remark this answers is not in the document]
dmlloyd: I understand.  It's just that if the `Config` object is published before the interceptor is initialized, then it is possible for the config to be accessed before the interceptor is ready.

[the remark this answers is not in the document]
dmlloyd: The init order really should be inner-to-outer, not outer-to-inner, for the exact same reason.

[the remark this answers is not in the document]
radcortez: I did it in a way that init is done in Config constructor so it should be fine.

But no worries, I'll change it to use `ConfigSourceInterceptorContext` instead.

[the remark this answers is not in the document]
dmlloyd: Actually this is not fine: https://stackoverflow.com/a/9851843/2033093

implementation/src/main/java/io/smallrye/config/ProfileConfigSourceInterceptor.java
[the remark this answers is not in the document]
radcortez: I can add the option to set that up as well.

Anyway, in other SmallRye projects, the namespace for config starts with `smallrye` shouldn't be consistent there?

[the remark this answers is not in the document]
dmlloyd: Perhaps we should have a configuration property to set the name of the configuration property... :slightly_smiling_face:

I guess it doesn't matter what the property name is as long as it can be set in the interceptor's constructor.

implementation/src/main/java/io/smallrye/config/ProfileConfigSourceInterceptor.java
```diff
@@ -0,0 +1,39 @@
...
+            profile = config.getOptionalValue(SMALLRYE_PROFILE, String.class).orElse(null);
+        }
+    }
+
+    @Override
+    public ConfigValue getValue(final ConfigSourceInterceptorContext context, final String name) {
+        if (profile != null) {
+            final ConfigValue profileValue = context.proceed("%" + profile + "." + name);
+            if (profileValue != null) {
+                return profileValue;
```
dmlloyd, line 33: This is slightly incorrect unfortunately.

The following cases should be tested.  Given ConfigSource A(ordinal 100) and B(ordinal 200):

* _A_ contains `%profile.foo` and _B_ contains `foo`: _A_ wins
* _A_ contains `foo` and _B_ contains `%profile.foo`: _A_ still wins
* _A_ contains both `%profile.foo` and `foo`: _A_ `%profile.foo` wins

Otherwise we have a situation where a config source of higher ordinal can override one of lower ordinal.

So this means that I was incorrect in my original assumption:  We have to get `ConfigValue` instances for both property names, and then if both are returned, we have to compare the ordinals and return the value for the source with the lower ordinal.  Good thing we put ordinal on `ConfigValue`, right? :wink:

implementation/src/main/java/io/smallrye/config/ProfileConfigSourceInterceptor.java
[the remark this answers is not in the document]
radcortez: Sure, I'll add it :)

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptor.java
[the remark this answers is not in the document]
radcortez: Thanks!

implementation/src/main/java/io/smallrye/config/ProfileConfigSourceInterceptor.java
radcortez: Hehe, yes. 

I guess I just kept that code we discussed some time ago in my mind and didn't think too much about it.

Let me test and fix that.

radcortez: BTW,

> we have to compare the ordinals and return the value for the source with the lower ordinal.

Shouldn't this be:

return the value for the source with the **higher** ordinal?

dmlloyd: Here's my trick to remember ordinal vs priority:

* Higher priority wins (think "this is top priority!")
* Lower ordinal wins (think "1st is more important than 2nd")

dmlloyd: If it were my choice though, I would have used priority for config source as well...

radcortez: One clarification:

In:
> A contains `foo` and B contains `%profile.foo`: A still wins

Why don't we want to have B override A if B has a higher priority and it is the active profile?

radcortez: > * Lower ordinal wins (think "1st is more important than 2nd")

Sure, but in the `getOrdinal` java docs we have:

```java
* If a property is specified in multiple config sources, the value in the config source with the highest ordinal
* takes precedence.
```

If this is the case, shouldn't it be `lowest`?

dmlloyd: Because then if users want to override a value (for example using a system property), they must now know whether the value came from a profile (and which profile might be active) before they can do so.  But what users generally *want* is that if they set a system property or env var, that takes precedence no matter what.  This is consistent with the behavior of Quarkus today.

dmlloyd: > > * Lower ordinal wins (think "1st is more important than 2nd")
> 
> Sure, but in the `getOrdinal` java docs we have:
> 
> ```java
> * If a property is specified in multiple config sources, the value in the config source with the highest ordinal
> * takes precedence.
> ```
> 
> If this is the case, shouldn't it be `lowest`?

Yes, historically this was mixed up in a *number* of places, and I see we've mixed it up again. :-)

radcortez: Well, then it is a huge mix. Because even the default ConfigSources use the notion of priority:

```java
 * <li>{@linkplain System#getProperties() System properties}, with an {@linkplain #getOrdinal() ordinal value} of {@code 400}</li>
 * <li>{@linkplain System#getenv() Environment properties}, with an ordinal value of {@code 300}</li>
 * <li>The {@code /META-INF/microprofile-config.properties} {@linkplain ClassLoader#getResource(String) resource}, 
 * with an ordinal value of {@code 100}</li>
```

And we don't want `microprofile-config.properties` to be the most important :)

radcortez: > Because then if users want to override a value (for example using a system property), they must now know whether the value came from a profile (and which profile might be active) before they can do so. But what users generally _want_ is that if they set a system property or env var, that takes precedence no matter what. This is consistent with the behavior of Quarkus today.

Effectively, it might just be easier to define that profiles are applied per ConfigSource?

dmlloyd: > Well, then it is a huge mix. Because even the default ConfigSources use the notion of priority:

Yeah maybe I have it backwards!  Mixing things up is sadly common for me.  Maybe we should change all incidence of "ordinal" to "priority" and end the confusion for good (even in the specification)!

> > Because then if users want to override a value (for example using a system property), they must now know whether the value came from a profile (and which profile might be active) before they can do so. But what users generally _want_ is that if they set a system property or env var, that takes precedence no matter what. This is consistent with the behavior of Quarkus today.
> 
> Effectively, it might just be easier to define that profiles are applied per ConfigSource?

It would be easier to implement, but not useful to users to do it this way.  The config source priority has to be considered before the profile.

radcortez: Just trying to figure out the best way to then explain to the users. In Quarkus documentation this does not seem clear:
https://quarkus.io/guides/config#configuration-profiles

So we want to say that higher priority (or lower ordinal :)) sources have precedence over a profile configuration.

dmlloyd: Yes, I agree with that.  Sorry if I didn't understand you.

radcortez: No worries. I most likely I didn't explain myself properly :)

implementation/src/main/java/io/smallrye/config/ProfileConfigSourceInterceptor.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Optional;
+
+import javax.annotation.Priority;
+
+@Priority(600)
+public class ProfileConfigSourceInterceptor implements ConfigSourceInterceptor {
+    public static final String SMALLRYE_PROFILE = "smallrye.config.profile";
+
+    private String profile;
+    private String config;
```
dmlloyd, line 12: I guess this is meant to be the configuration key, so its name probably ought to reflect that; also, since it is initialized in the constructor, it should be `final`.

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptor.java
```diff
@@ -18,6 +18,19 @@
...
+    /**
+     * This method initializes the interceptor, with access to the current {@link ConfigSourceInterceptorContext}.
+     *
+     * Interceptors in the chain are initialized in priority order and the current
+     * {@link ConfigSourceInterceptorContext} contains the current interceptor, plus all other interceptors already
+     * initialized.
+     *
+     * @param context the current {@link ConfigSourceInterceptorContext} associated to the Config.
+     */
+    default void initialize(ConfigSourceInterceptorContext context) {
```
dmlloyd, line 30: I wonder if it would be better to either have a contract where interceptor constructors are invoked reflectively passing in a `ConfigSourceInterceptorContext`, or else have users implement an `ConfigSourceInterceptorFactory` interface which accepts that type and returns a `ConfigSourceInterceptor`?

This has some design benefits:
* Anything that would have been done in `initialize` can be done in a constructor and thus written to a `final` field
* Interceptors always have deterministic state
* The interceptor interface is simpler

WDYT?

radcortez: Yes, I think it would be better.

We already use a `Factory` pattern for `SmallRyeConfig`. Might as well do it here as well to keep it consistent.

implementation/src/main/java/io/smallrye/config/ProfileConfigSourceInterceptor.java
radcortez: Oh yeah, this was from the old version.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -163,13 +156,29 @@ private ConfigSourceInterceptorContext buildInterceptorChain(final SmallRyeConfi
                     return null;
                 }, null);
 
-        for (int i = interceptors.size() - 1; i >= 0; i--) {
-            current = new SmallRyeConfigSourceInterceptorContext(interceptors.get(i), current);
+        List<ConfigSourceInterceptor> configSourceInterceptors = initInterceptors(interceptors, current);
+        for (int i = configSourceInterceptors.size() - 1; i >= 0; i--) {
+            current = new SmallRyeConfigSourceInterceptorContext(configSourceInterceptors.get(i), current);
```
dmlloyd, line 161: Why recreate the contexts here?  You could reuse the same context instances and do the init in a single loop.

implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
```diff
@@ -93,11 +96,16 @@ public SmallRyeConfigBuilder addDiscoveredInterceptors() {
         return discoveredConverters;
     }
 
-    List<ConfigSourceInterceptor> discoverInterceptors() {
-        List<ConfigSourceInterceptor> interceptors = new ArrayList<>();
+    List<InterceptorWithPriority> discoverInterceptors() {
```
dmlloyd, line 99: If we're bundling our interceptors, I think we should separate `addDefaultInterceptors` from `addDiscoveredInterceptors` so that we can replace the default ones (for example, configuring them differently) while still discovering user interceptors.

radcortez: This is something we should decide. At this time, we are not bundling any of our interceptors by default, you need to register them via the ServiceLoader.

Do you think we should bundle them by default?

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
radcortez: I guess I can return the context instances instead. The idea was to avoid reusing the context passed to the factory, since it wouldn't be the final chain.

implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
dmlloyd: Yes I think so.  Otherwise it will be very hard for users to have discovered interceptors while also excluding these initial ones.

radcortez: They don't need to exclude them. They are only discovered if you register them via ServiceLoader and right now we are not registering any of the interceptors we added. Right now the user will need to add the ServiceLoader file to they app with any of our interceptors reference to register them. So, even if they call `addDiscoveredInterceptors` none of ours will be added.

BTW, we don't have a `addDefaultInterceptors` for the same reason, but I can add it.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
radcortez: Another reason why I've done it that way is because it is easier to build the chain from backwards to the first, since we want our last interceptor to be the one that actual lookups the value, so we can link them together to the last one by least priority to highest. Start by `lookup` then `lowest->lookup`, `highest->lowest->lookup`.

On constructor, we want to initialize them on priority as well, so it means that we need to build the chain `highest->lookup`, meaning that we would need to break the chain in the middle to insert less priority interceptors `highest->low->lookup`

Do you thing we should change this?

dmlloyd: I think we should change it so that interceptors are initialized in last-to-first order.  That makes everything consistent and it means that an interceptor will never "see" a subsequent interceptor on initialization that it wouldn't normally "see".

dmlloyd: IOW build it like this:

* `lookup`
* `ctxt->lookup`
* `interceptor(last)->ctxt->lookup`
* `ctxt->interceptor(last)->ctxt->lookup`

etc., initializing at the same time that it is built.

radcortez: Sure.

implementation/src/main/java/io/smallrye/config/ProfileConfigSourceInterceptor.java
```diff
@@ -0,0 +1,40 @@
...
+        this.profile = Optional.ofNullable(context.proceed(profileConfigName)).map(ConfigValue::getValue).orElse(null);
+    }
+
+    @Override
+    public ConfigValue getValue(final ConfigSourceInterceptorContext context, final String name) {
+        if (profile != null) {
+            final ConfigValue profileValue = context.proceed("%" + profile + "." + name);
+            if (profileValue != null) {
+                final ConfigValue originalValue = context.proceed(name);
+                return originalValue.getConfigSourceOrdinal() > profileValue.getConfigSourceOrdinal() ? originalValue
```
dmlloyd, line 33: `originalValue` could be `null`.

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptorFactory.java
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Gets the Class of the Interceptor.
+     *
+     * This is required, because the interceptor priority needs to be sorted before doing initialization.
+     *
+     * @return the Class of the {@link ConfigSourceInterceptor}
+     */
+    Class<? extends ConfigSourceInterceptor> getInterceptorClass();
+
+    final class Delegate implements ConfigSourceInterceptorFactory {
```
dmlloyd, line 30: This should be static, I think.

dmlloyd: But interface members are also (almost) always public, and this is no exception.  It probably shouldn't be part of the API.

radcortez: Ok, let me move that to a separate class.

implementation/src/main/java/io/smallrye/config/ProfileConfigSourceInterceptor.java
radcortez: True. Let me fix that.

implementation/src/main/java/io/smallrye/config/ProfileConfigSourceInterceptor.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public ConfigValue getValue(final ConfigSourceInterceptorContext context, final String name) {
+        if (profile != null) {
+            final ConfigValue profileValue = context.proceed("%" + profile + "." + name);
+            if (profileValue != null) {
+                final ConfigValue originalValue = context.proceed(name);
+                if (originalValue != null) {
+                    return originalValue.getConfigSourceOrdinal() > profileValue.getConfigSourceOrdinal() ? originalValue
```
dmlloyd, line 34: One more tiny detail... if the ordinals are equal, we then disambiguate based on the name of the config source as per https://github.com/smallrye/smallrye-config/blob/master/implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java#L56-L65 so we probably want a simlar check here.

radcortez: Ok!

implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
```diff
@@ -245,4 +286,38 @@ int getPriority() {
...
+                @Override
+                public ConfigSourceInterceptor getInterceptor(final ConfigSourceInterceptorContext context) {
+                    return interceptor;
+                }
+
+                @Override
+                public int getPriority() {
+                    return Optional.ofNullable(interceptor.getClass().getAnnotation(Priority.class))
+                            .map(Priority::value)
+                            .orElse(ConfigSourceInterceptorFactory.super.getPriority());
```
dmlloyd, line 305: It's unusual that the factory priority is lower than the annotation.  Doesn't that seem backwards?  As a user I would probably expect that the annotation would be used only if there was no priority given... of course that would mean that it has to be *possible* for there to be no priority.  Some ways I can think of to do that would be to define a `ConfigSourceInterceptorFactoryWithPriority` subinterface, or change `getPriority()` to return an `OptionalInt`.

radcortez: Yes, maybe we want the method to have priority :) 

Probably is better to just change it to an OptionalInt, so we don't create yet another interface.

PR #276, opened 2020-04-11
Properties from file system should be search with the same rules used for environment variables

kifj:
**Statement of problem**

given a interface for the MicroProfile REST client

```
package com.example;
@Produces(APPLICATION_JSON)
@RegisterRestClient
public interface QuickQuoteService {
  ...
}
```

which will retrieve its URL from MP Config with the key

`com.example.QuickQuoteService/mp-rest/url`

and a ConfigMap in Kubernetes with entries

```apiVersion: v1
kind: ConfigMap
data:
 com_example_QuickQuoteService_mp_rest_url: https://quote.cnbc.com/quote-html-webservice
```

A ConfigMap defines rules on the key, which forbids '/', because it cannot be mapped to an environment variable or a file name. 

According to specification the replacement rules are defined for environment variables, all non-alphanumeric characters are replaced with '_'; and if that doesn't resolve, all alphabetical characters can be upper-case. 

This ConfigMap will be mounted in a volume, resulting in a file named  com_example_QuickQuoteService_mp_rest_url with the value as content.

The FileSystemConfigSource should pick up that file (which it does) and the Config should return its content as value for the key `com.example.QuickQuoteService/mp-rest/url` (which is doesn't)

**Solution**

The mapping rules for environment variables (see  EnvConfigSource) should apply to file names as well, even if the specification does not define this behavior.

### DIFF
--- a/sources/file-system/src/main/java/io/smallrye/config/source/file/FileSystemConfigSource.java
+++ b/sources/file-system/src/main/java/io/smallrye/config/source/file/FileSystemConfigSource.java
@@ -23,6 +23,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Collections;
 import java.util.Map;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -59,6 +60,7 @@ import io.smallrye.config.common.MapBackedConfigSource;
 public class FileSystemConfigSource extends MapBackedConfigSource {
     private static final long serialVersionUID = 654034634846856045L;
 
+    private static final Pattern PATTERN = Pattern.compile("[^a-zA-Z0-9_]");
     private static final Logger LOG = Logger.getLogger("io.smallrye.config");
 
     FileSystemConfigSource(File dir) {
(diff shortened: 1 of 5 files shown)


radcortez:
Hi @kifj 

Thank yo for the PR! 

Just a clarification: from my understanding, the spec only mandates the name mapping to happen for Environment Variables: https://github.com/eclipse/microprofile-config/blob/cafee0a79fd36dfaa4f7b30bc77efa75eefa589b/spec/src/main/asciidoc/configsources.asciidoc#L53

Now, that doesn't mean we shouldn't support this in the `FileSystemConfigSource` :)

Do you mind adding a few tests?

sources/file-system/src/main/java/io/smallrye/config/source/file/FileSystemConfigSource.java
```diff
@@ -96,4 +98,30 @@ private static String readContent(Path file) {
             throw new UncheckedIOException(e);
         }
     }
+
+    @Override
+    public String getValue(String name) {
```
radcortez, line 103: Add tests.

kifj:
You're right regarding the specification so I have updated the comment and added a unit test

radcortez:
Thanks!

Can you also document the new feature for `FileSystemConfigSource` here:
https://github.com/smallrye/smallrye-config/blob/9a7716257d73881f44ab1d9dda466135967df278/doc/modules/ROOT/pages/config-sources/filesystem-config-source.adoc?

PR #278, opened 2020-04-15
Fixes #269. Initial work to support secrets.

### LINKED ISSUE #269: Hide secrets in Configuration
Original issue from Quarkus:
https://github.com/quarkusio/quarkus/issues/7442

But I think we should support this at the SmallRye Config level.

radcortez:
This is the initial work to implement #269 .

A new class `SecretKeys` can be configured with the key names to be considered secrets and access to these keys is protected unless it is explicitly asked for.

There are still some open questions:
- How do we protect `ConfigSource#getPropertyNames`? We could wrap it in our own source, but then it will become a different type for the user. Is this acceptable?
- We may need to expose the `SecretKeys` in the interceptor chain, so we can control access from the interceptors as well. This requires to add a getter method in `ConfigSourceInterceptorContext`.
- Users would need to cast Config to SmallRyeConfig to use this feature. Unless we give them some other direct API.


radcortez:
@dmlloyd please, let me know if you think this is going into the right direction :)

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptorContext.java
```diff
@@ -14,4 +14,6 @@
      * @return a {@link ConfigValue} with information about the key, lookup value and source ConfigSource.
      */
     ConfigValue proceed(String name);
+
+    SecretKeys getSecretKeys();
```
dmlloyd, line 18: I think adding special purpose features like this one on to general purpose interfaces risks a large growth.

From the end user perspective, IMO ideally they should only "see" the API concept of secret keys when the config source is being built, and when (globally) turning on and off secrets.  The config source builder can build an interceptor which has the key set right on it, so the intermediate abstraction of `SecretKeys` should not be needed.

implementation/src/main/java/io/smallrye/config/SecretKeys.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public boolean isSecret(String propertyName) {
+        return secrets.apply(propertyName);
+    }
+
+    public boolean isSecretAccessible(String propertyName) {
+        return !isSecret(propertyName) || !LOCKED.get();
+    }
+
+    public void accessSecrets(Runnable runnable) {
```
dmlloyd, line 34: These four methods should be `static`, and not dependent on the property name.  Rather I think it would make sense to have a query (`isLocked()`) that the interceptor checks.  If the name is in the set, and `isLocked()` is `true`, then it will simply throw `SecurityException`.

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptorContext.java
radcortez: I've added this here so an interceptor could find is a particular key is a secret, in the case of a logging interceptor, to exclude it from logging.

For instance, when you open up the secrets to perform injection, it would open to the entire interceptor chain, so a logging interceptor has access to the secret being injected and print the result.  We would need to disable access in a middle of an enable access and I think it might be a little confusing.

We can certainly pass in the Set to the logging interceptor as well, and make the check as we would do in the security one and skip the log. On the other hand, I think it is useful to have some kind of API that tells you is a certain key `isSecret`. Or we can just expose the Set directly.

radcortez:
@dmlloyd This should be more in line with your review. 

I think there are still 2 thing we need to figure out:
- How to protect `ConfigSource#getPropertyNames`?
- How do interceptors "exclude" or run in a protected context (in the case for a LoggingInterceptor) when the lock is open before the chain.

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptorContext.java
dmlloyd: TBH I imagined loggers just catching the exception.  Otherwise there's yet another piece of information that has to be propagated along with the configuration value.

implementation/src/main/java/io/smallrye/config/SecretKeys.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.function.Supplier;
+
+class SecretKeys implements Serializable {
+    private static final ThreadLocal<Boolean> LOCKED = ThreadLocal.withInitial(() -> Boolean.TRUE);
+
+    static boolean isLocked() {
+        return LOCKED.get();
+    }
+
+    static void accessSecrets(Runnable runnable) {
```
dmlloyd, line 13: This API should be public IMO.

```java
public static void doUnlocked(Runnable r) { ... }
public static <T> T doUnlocked(Supplier<T> s) { ... }
public static <T, R> R doUnlocked(Function<T, R> f) { ... }
```

And as mentioned in my other comment, the opposite:
```java
public static void doLocked(Runnable r) { ... }
public static <T> T doLocked(Supplier<T> s) { ... }
public static <T, R> R doLocked(Function<T, R> f) { ... }
```

dmlloyd:
Using the exception model, we don't need to filter the property names in any way: the key is not secret, and anyway property expressions could reference the key so the existence of the key can't really be hidden.

To "re-lock" we could have opposite methods which run a function of some sort with the lock flag re-set.

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptorContext.java
radcortez: Ah no, it is the logger to log the config name, source and value, so you know what was loaded. It seems to be a common request, at least seen some people asking about it. In that case, you need to lock the keys again in the case you run in an unlocked context so you don't print the value accidentally.

implementation/src/main/java/io/smallrye/config/SecretKeys.java
radcortez: Ok. I was also delegating this to SmallRyeConfig. I think it is easier as an API perspective to be "found" since you see the method there, but we can have both.

radcortez:
Yes, it would be handled by the exception model.

In my mind, to be extra safe, I thought that we could configure all the secret keys plus any expansion variables that may contain sensitive information and exclude them in the property names.

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptorContext.java
dmlloyd: Yes, that is understood.  I mean, I expect that things which may log locked properties will surround accesses with try/catch, and skip entries for which an exception was thrown.

dmlloyd:
We don't need to go to that extent.  If we rely solely on an exception model, and a user accesses a property with a variable in it which expands to a secret, we get behavior like this:

|   | Locked | Unlocked |
| -: | ------ | -------- |
| **Expanding** | `SecurityException` | OK (secret is in expansion) |
| **Not expanding** | OK (no secret in expansion) | OK (no secret in expansion) |

So we don't need to do any complex analysis (which is always dangerous when talking about security anyways) for this case.  There doesn't seem to be any need to exclude the names of secret properties from this perspective: we have one unified protection mechanism, so that's only one thing that can potentially fail.

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptorContext.java
radcortez: Yes, but in the case where the lock is unlocked before the chain, you still want the logger to be locked.

radcortez:
Yes, my idea was not to do any complex analysis. It would on the user to add those names into the secret list and we could just exclude them from everywhere (include the list properties). Nevermind, lets not touch property names.

dmlloyd:
Furthermore I'd imagine a logger would capture exceptions anyway, to be able to handle a variety of situations.  Consider in this example that `property2` is a secret, and `property3` cannot be accessed (due to a transient problem in the config source perhaps).  As a user I'd expect something like this:

```
property1 = "foo"
property2 = <secret>
property3 = <cannot access>
property4 = "bar ${property2}"
```

Catching exceptions for the exceptional cases really makes this easy.  The logger doesn't really need any special knowledge, it just needs to catch `RuntimeException`.  It doesn't even have to discriminate TBH; that's purely an extra-value feature.

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptorContext.java
radcortez: So, the general purpose logger, should always run its code in a new lock.

radcortez:
Yes, that is ok. Unless I'm missing something, the case that I wanted to prevent is when you have the lock open and you have the logger in the middle, but you don't want to log the access, right?

implementation/src/main/java/io/smallrye/config/SecretKeys.java
dmlloyd: It all depends on how many things you want to put on the root API.  Either way works for me I guess.

radcortez: For now just `accessSecrets(Runnable)` and `accessSecret(Supplier)`. We can add the `Function` as well. These look to me to be the most common ones that you will need. If you need something more specialised, you use `SecretKeys`.

dmlloyd:
If the question is "do we log accesses to secret keys", I think the answer is probably "why not?": if you're logging config accesses then you probably want that information as well.  But whether or not it is actually logged, you still have to figure out whether it's secret, so you'd eventually need something like:

```java
    try {
        ConfigValue maybeLocked = doLocked(() -> context.proceed(name));
        log(name, maybeLocked.getValue());
    } catch (SecurityException e) {
        log(name, "<locked>");
    } catch (RuntimeException e) {
        log(name, "<access error>");
    }
    return context.proceed(name);
```

radcortez:
Yes, I think we can log secret config access, but omitting the value of course.

That is why we need a `doLocked` that we can do in the middle of a unlocked context, because we cannot trust the lock value at that point.

radcortez:
I think it should be better now :)

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -254,6 +273,22 @@ public String getRawValue(String name) {
         return configSourcesRef.get();
     }
 
+    public void unlockSecrets(Runnable runnable) {
+        SecretKeys.doUnlocked(runnable);
+    }
+
+    public <T> T unlockSecrets(Supplier<T> supplier) {
+        return SecretKeys.doUnlocked(supplier);
+    }
```
dmlloyd, line 282: I'm sorry I've been very unclear about all of this: it is my opinion that the doLocked/doUnlocked should only be available from one place (either `SmallRyeConfig` or `SecretKeys`) and should be `static` (because the underlying mechanism is global, so making it appear that it is bound to a configuration instance may mislead the user into thinking that only the configuration in question should be affected).

radcortez: Hum, that raises another interesting question. Since the secret keys are defined by config instance, should we consider the lock / unlock also per config instance?

dmlloyd: I would say "no", because from a security standpoint, it's either OK to access secrets (generally speaking) or it is not OK.  From a technical standpoint, having many thread local variables is expensive.  So I lean towards "no", it should remain global.

radcortez: Ok, let's leave that way. I'll fix the previous comment.

implementation/src/main/java/io/smallrye/config/SecretKeys.java
```diff
@@ -0,0 +1,52 @@
+package io.smallrye.config;
+
+import java.io.Serializable;
+import java.util.function.Supplier;
+
+public class SecretKeys implements Serializable {
```
dmlloyd, line 6: Since this class comprises only static members, it should be `final` and have a `private` constructor.

implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
```diff
@@ -150,6 +153,17 @@ public OptionalInt getPriority() {
...
         }));
+        interceptors.add(new InterceptorWithPriority(new ConfigSourceInterceptorFactory() {
+            @Override
+            public ConfigSourceInterceptor getInterceptor(final ConfigSourceInterceptorContext context) {
+                return new SecretKeysConfigSourceInterceptor(secretKeys);
+            }
+
+            @Override
+            public OptionalInt getPriority() {
+                return OptionalInt.of(-Integer.MAX_VALUE);
```
dmlloyd, line 164: I'm not sure this interceptor should be absolutely first.  It might then miss lookups on keys that are relocated or otherwise manipulated by the user.

I wonder if we should document our interceptor priorities with notes about where user interceptors should be placed.  IIRC some specifications define useful integer constants for users to reference.

radcortez:
Thanks, let me fix that.

implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
radcortez: Sure, we can add some integer constants for the interceptors. That may also help when we add support to interleave interceptors and config sources.

PR #287, opened 2020-04-21
Fixes #285. Initial work to intercept ConfigSources.

### LINKED ISSUE #285: Interceptors need to be able to intercept iteration
As I stated in #196, the profile interceptor needs to be able to intercept iteration so that it can add properties found only in profiles.

Additionally, the profile interceptor needs the ability to filter out profile names, so that the set of configuration properties appears to be only the union of the set of properties present with no profile and the current profile.

It seems like a good idea to simultaneously attempt to solve the potential performance issue with iteration mentioned on the MP Config project.  The issue is that it is sometimes the case that iteration of config properties intends to get the value as well as the key, because lookup of individual keys might be somewhat slower than iterating the full key-value entries of the configuration for some configuration sources.  So it might be a good idea to have two iteration mechanisms: one for names (String) and one for name+value (ConfigValue).

radcortez:


implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -136,28 +134,12 @@ private ConfigSourceInterceptorContext buildInterceptorChain(final SmallRyeConfi
...
-                                        .withConfigSourceOrdinal(configSource.getOrdinal())
-                                        .build();
-                            }
-                        }
-                    }
-                    return null;
-                }, null);
+                new ConfigSourceInterceptor.Empty(), null);
+        // TODO - radcortez - This needs to play with addConfigSource
+        List<ConfigSource> configSources = configSourcesRef.get();
```
radcortez, line 139: We need to rebuild the interceptor chain if the ConfigSources list change.

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptor.java
```diff
@@ -32,4 +33,29 @@
      * @return a ConfigValue with information about the key name, value, config source origin and ordinal.
      */
     ConfigValue getValue(ConfigSourceInterceptorContext context, String name);
+
+    default Iterable<String> getNames(ConfigSourceInterceptorContext context) {
```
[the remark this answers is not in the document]
dmlloyd: Streams are significantly more difficult to intercept AFAICT.  You're limited to a much smaller number of operations.  Streams also are likelyt to perform worse with many more allocations.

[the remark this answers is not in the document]
dmlloyd: I would use `Iterator` TBH.  It's one less object to construct.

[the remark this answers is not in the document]
radcortez: I had Iterator before but Config uses Iterable so I've changed it to be consistent. What do you think?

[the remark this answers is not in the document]
dmlloyd: Config is the front end interface; its `Iterable` can trivially use the interceptor chain's `Iterator` to materialize it.  It's better to do this at the front end because otherwise every interceptor that needs to wrap will have to wrap both the `Iterable` and the `Iterator` that it creates.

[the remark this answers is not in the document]
radcortez: Ok. I'll change that.

radcortez:
@dmlloyd let me know what you think about the direction.

I had to rewrite `SmallRyeConfig` a bit to be able to cope with `addConfigSource`, because we also need to recreate the Interceptor chain now.

There is still documentation missing and some tests (and I'll work on that), just don't want to invest too much without us agreeing in the API and implementation.

implementation/src/main/java/io/smallrye/config/SmallRyeConfigSourceInterceptorContext.java
```diff
@@ -15,4 +17,14 @@
     public ConfigValue proceed(final String name) {
         return interceptor.getValue(next, name);
     }
+
+    @Override
+    public Iterator<String> names() {
```
dmlloyd, line 22: The `proceed` method is a verb.  Maybe this should be `iterateNames` so that it is also a verb?

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptor.java
```diff
@@ -32,4 +34,29 @@
      * @return a ConfigValue with information about the key name, value, config source origin and ordinal.
      */
     ConfigValue getValue(ConfigSourceInterceptorContext context, String name);
+
+    default Iterator<String> getNames(ConfigSourceInterceptorContext context) {
```
dmlloyd, line 38: Maybe these should be `iterateNames` for consistency with the context (see other comment)?

radcortez:
> Overall it seems OK. I wonder if we should consider interleaving interceptors with sources though; that could be interesting (and maybe simpler).

Yes, I'm also planning to do that, but I think we should do it a separate PR. 

> 
> We probably should deprecate the `addSource` method.

I was going to propose that as well. I've seen that you added it for Quarkus, but I couldn't find any usage of that in Quarkus at all. 

I'll finish this up so we can move forward. Thank you!

radcortez:
Ok, I think it is good to go :)

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptor.java
```diff
@@ -1,35 +1,86 @@
...
 import java.io.Serializable;
+import java.util.Collections;
+import java.util.Iterator;
 
 /**
- * The ConfigSourceInterceptor allows you to intercept the resolution of a configuration key name before the
- * configuration value is resolved by the Config and before any conversion taking place.
+ * The ConfigSourceInterceptor allows to intercept the resolution of a configuration name before the
+ * configuration value is resolved by the Config and before any conversion taking place. It can also intercept
+ * iteration of co
```
dmlloyd, line 10: Looks like this didn't get completed.

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptor.java
```diff
@@ -1,35 +1,86 @@
...
  * <p>
  *
- * This is useful to provide logging, transform the key or substitute the value.
+ * This is useful to provide logging, transform names or substitute values.
  * <p>
  *
  * Implementations of {@link ConfigSourceInterceptor} are loaded via the {@link java.util.ServiceLoader} mechanism and
  * can be registered by providing a resource named {@code META-INF/services/io.smallrye.config.ConfigSourceInterceptor},
  * which contains the fully qualified {@code ConfigSourceInterceptor} implementation class name as its content.
+ *
```
dmlloyd, line 19: Missing paragraph tag.

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptor.java
```diff
@@ -1,35 +1,86 @@
...
+ * Alternatively, a {@link ConfigSourceInterceptor} can also be loaded with a {@link ConfigSourceInterceptorFactory}.
  * <p>
  *
+ *
```
dmlloyd, line 23: Extra line I guess?

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptor.java
```diff
@@ -1,35 +1,86 @@
...
+     * custom {@link ConfigValue} built by the interceptor. Calling
+     * {@link ConfigSourceInterceptorContext#proceed(String)} will continue to execute the interceptor chain. The chain
+     * can be short-circuited by returning another instance of {@link ConfigValue}.
      *
      * @param context the interceptor context. See {@link ConfigSourceInterceptorContext}
-     * @param name the configuration key name being intercepted.
+     * @param name the configuration name being intercepted.
      *
-     * @return a ConfigValue with information about the key name, value, config source origin and ordinal.
+     * @return a {@link ConfigValue} with information about the name, value, config source and ordinal or {@code null}.
```
dmlloyd, line 39:
```suggestion
     * @return a {@link ConfigValue} with information about the name, value, config source and ordinal, or {@code null} if the value isn't present
```

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptor.java
```diff
@@ -1,35 +1,86 @@
...
+     *
+     * @param context the interceptor context. See {@link ConfigSourceInterceptorContext}
+     *
+     * @return an Iterator of {@link ConfigValue} with information about the name, value, config source and ordinal.
+     */
+    default Iterator<ConfigValue> iterateValues(ConfigSourceInterceptorContext context) {
+        return context.iterateValues();
+    }
+
+    class Empty implements ConfigSourceInterceptor {
```
dmlloyd, line 70:
```suggestion
    ConfigSourceInterceptor EMPTY = new ConfigSourceInterceptor() {
```

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptorContext.java
```diff
@@ -1,17 +1,32 @@
 package io.smallrye.config;
 
 import java.io.Serializable;
+import java.util.Iterator;
 
 /**
- * Exposes contextual information about the intercepted invocation of {@link ConfigSourceInterceptor}. This allows you
- * to control the behavior of the invocation chain.
+ * Exposes contextual information about the intercepted invocation of {@link ConfigSourceInterceptor}. This allows to
```
dmlloyd, line 7:
```suggestion
 * Exposes contextual information about the intercepted invocation of {@link ConfigSourceInterceptor}. This allows implementers to
```

implementation/src/main/java/io/smallrye/config/ConfigSourceInterceptor.java
radcortez: Ops, not sure what happened here.

implementation/src/main/java/io/smallrye/config/ConfigValueConfigSourceWrapper.java
```diff
@@ -0,0 +1,70 @@
+package io.smallrye.config;
+
+import java.io.Serializable;
+import java.util.Map;
+import java.util.Set;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+public class ConfigValueConfigSourceWrapper implements ConfigValueConfigSource, Serializable {
```
dmlloyd, line 9: This could probably be `final` and non-`public`, could it not?

implementation/src/main/java/io/smallrye/config/ConfigValueMapStringView.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.AbstractCollection;
+import java.util.AbstractMap;
+import java.util.AbstractSet;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+
+public class ConfigValueMapStringView extends AbstractMap<String, ConfigValue> {
```
dmlloyd, line 12: Could this one also be non-public?

implementation/src/main/java/io/smallrye/config/ConfigValueConfigSourceWrapper.java
radcortez: Yes.

implementation/src/main/java/io/smallrye/config/ConfigValueMapStringView.java
radcortez: Yes.

PR #288, opened 2020-04-21
Add i18n messages and logging

kenfinnigan:


kenfinnigan:
If this looks ok, I will do it for the other modules as well

PR #296, opened 2020-04-27
Support recursive expansion.

radcortez:


dmlloyd:
Very nice that we don't have to use a thread local here!

radcortez:
Thanks :)

PR #298, opened 2020-05-04
Fixes to interceptors.

radcortez:


implementation/src/main/java/io/smallrye/config/FallbackConfigSourceInterceptor.java
```diff
@@ -1,16 +1,15 @@
...
 import javax.annotation.Priority;
 
 @Priority(Priorities.LIBRARY + 400)
 public class FallbackConfigSourceInterceptor implements ConfigSourceInterceptor {
-    private final Function<String, String> mapping;
+    private final NameMapper mapping;
 
-    public FallbackConfigSourceInterceptor(final Function<String, String> mapping) {
-        this.mapping = mapping != null ? mapping : Function.identity();
+    public FallbackConfigSourceInterceptor(final NameMapper mapping) {
```
dmlloyd, line 11: In general I don't think it's a good idea to introduce specific types for the purpose of serializability.  It's better to document that the serializability of a given configuration is contingent on its components also being serializable.  The default function for example could instead be a non-public singleton class instance.

radcortez: The issue is us having and instance of Function in the interceptor, and having to keep a copy of the interceptors in the Config (which needs to be Serializable), because of the feature that allows us to add ConfigSources after an instance is already built.

dmlloyd: The function instances that we define can definitely be serializable (I recommend defining an actual class rather than a lambda, since classes have more stable serialization characteristics than lambdas do).

radcortez: Weird. I think I've done that, but it was still failing in the Serialization TCK test. Let me give another look.

radcortez: Ah also one thing is that we cannot force the user to pass in a class. He can pass the lambda and break the Config Serialization, unless we don't expose the API that receives the function.

dmlloyd: True, but the critical fact here is that it's the *user* doing it.  As long as we document that the configuration is only serializable when all of its components are, then I think we're OK.  Note for example that `ConfigSource` doesn't extend `Serializable`, so the user could break serialization today simply by creating a non-serializable configuration source.  This doesn't impact our ability to pass the TCK however.

radcortez: Sure. I'll retract this part and let the user handle it.

PR #313, opened 2020-05-16
Initial work to mix sources and interceptors priorities.

radcortez:


radcortez:
@dmlloyd any thoughts?

PR #317, opened 2020-05-27
Initial work to bridge with Bean Validator.

radcortez:
A lot of rough edges, but it was mostly to prove the concept.

- Introduced a ConfigValidator to hookup on validations methods.
- These methods are either called on CDI injection or when looking up a single property via the programmatic API.
- We are free to use any validation. We can implement our own, or hook up into BVal, so you can use BVal annotations

If this is interesting and we feel that the direction is right, I'll finish the missing pieces.


radcortez:
Replaced by https://github.com/smallrye/smallrye-config/pull/488.

Issue #321, opened 2020-06-05
SmallRyeConfigBuilder.getDefaultSources() should be `protected`


Joseph-Cass:
The following method should be `protected` rather than having no modifier:
https://github.com/smallrye/smallrye-config/blob/9c767a9a5053b64e97c389297aec8da4593e0c1d/implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java#L123
so that embedders can optionally extend the default sources.

Joseph-Cass:
Made the following PR with a potential solution: https://github.com/smallrye/smallrye-config/pull/322

radcortez:
Hi @Joseph-Cass. Thank you for the issue.

What are you trying to achieve exactly? I'm not so sure if we should change the visibility modifier to `protected`, at least without considering doing it for other methods as well.

If you want to add additional sources, you can just call `withSources(ConfigSource... configSources)` and add the sources you want to register, plus the default ones. If you don't want the default ones, just omit that method and register each source again with `withSources(ConfigSource... configSources)`

Joseph-Cass:
Thanks for the reply @radcortez. 

This looks like what we were after. We will try out your suggestion, thanks!

radcortez:
Great! So I guess we don't need this then? :)

tevans78:
I don't think that calling `withSources` from an overriden `build` method will work. We need to _only_ add these extra default sources if the `addDefaultSources` method has been called. That means either overriding the `getDefaultSources` method or being able to check the `addDefaultSources` flag (which is currently private).

radcortez:
That is already handled for you when `.build()` is called and the `SmallRyeConfig` object is created.

Might be easier if you can explain me what are you trying to achieve? Thanks!

tevans78:
The spec defines some default sources. Joe is trying to add some additional default sources as well. Sources that are added if `addDefaultSources` has been called on the builder. These are sources provided by Open Liberty, not by the user.

radcortez:
I see. So these are additional defaults sources from Liberty to the default ones specified by the spec (environment, system, file).

Yes, in this case, there is not clear way to do it. Maybe you could override `addDefaultSources` as well and track if it was called, so you call `withSources` to add the additional sources.

Anyway, to make this more consistent, other methods should be changed to protected. Maybe only the methods related with the `ConfigBuilder` API.

tevans78:
[...]

PR #322, opened 2020-06-05
Added protected

Joseph-Cass:
Made the method protected


Joseph-Cass:
Resolves  #321

Joseph-Cass:
I've added protected to the other methods as discussed @radcortez 

Could you please double check they are on the correct methods? Thanks

PR #333, opened 2020-06-17
Configuration Mapper

radcortez:
@dmlloyd let me know if you think this is going into the right direction. 

There is still a lot of things to take care and to try. Here are a few things that I did:
- Optional registration of class mappers in the SmallRyeBuilder for proper validation.
- ~~When a root cannot be found, one is created on the fly (experimental, I need to think better about this. We may want to merge with the current one).~~
- ~~Support JavaBeans getter style method names~~
- Validation is only done for properties that have a registered root. For instance if we register the root `server`, we only validate the `server` namespace. 
- Moved standalone API's to main package and rename annotations for consistency.
- Support empty root prefix.
- Use SmallRyeConfigs Default Values API.
- Fixes around Optional / Map mapping.

There is still a lot of things to work on (which I'm happy to do):
- More tests
- ~~Improve SmallRyeConfig entry point APIs to access the Mapping~~
- Additional CDI Integration (after #369)


radcortez:
@dmlloyd I've tried to fix some of the issues that I've found and write some additional tests. I believe this is going into the right track, but please let me know what do you think.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -211,6 +230,31 @@ public String getRawValue(String name) {
...
+        return getConfigProperties(klass, "");
+    }
+
+    @Experimental("TODO")
+    public <T> T getConfigProperties(Class<T> klass, String prefix) {
+        try {
+            final ConfigMappingProvider.Result result = configMappingProvider.mapConfiguration(this);
+            final T configRoot = result.getConfigRoot(prefix, klass);
+            // Build one on the fly
+            // Think if we want to merge the unknown one with the current configMapping.
```
dmlloyd, line 244: This I still don't agree with; it defeats the purpose of much of this API which was designed to uniformly validate a part of the configuration space.  Mapped configuration should be established while the configuration is being built IMO; after all, there is still a 1:1 mapping between "application" and configuration.

implementation/src/main/java/io/smallrye/config/mapper/ConfigMapping.java
```diff
@@ -636,6 +639,16 @@ public Result mapConfiguration(SmallRyeConfig config) throws ConfigurationValida
         return new Result(context.getRootsMap());
     }
 
+    private boolean isPropertyInRoot(String propertyName) {
+        final Set<String> registeredRoots = roots.keySet();
+        for (String registeredRoot : registeredRoots) {
+            if (propertyName.startsWith(registeredRoot)) {
```
dmlloyd: A prefix check is not sufficient here.  Roots might not cover a whole namespace - for example, in some use cases, it might be necessary to match a property name that is outside of the root in a space that isn't directly controlled by the author.  It should be a pattern which can be built into a `KeyMap`, and specified on the root as an optional configuration (e.g. `quarkus.**`).

implementation/src/main/java/io/smallrye/config/mapper/MappingContext.java
```diff
@@ -195,7 +194,8 @@ public NoSuchElementException noSuchElement(Class<?> type) {
         return new NoSuchElementException("A required configuration group of type " + type.getName() + " was not provided");
     }
 
-    public void unknownConfigElement(final ConfigValue configValue) {
+    public void unknownConfigElement(final String propertyName) {
+        problems.add(new Problem(propertyName + " does not map to any root"));
```
dmlloyd, line 198:
```suggestion
        problems.add(new Problem(propertyName + " is not a recognized configuration property name"));
```

implementation/src/main/java/io/smallrye/config/mapper/ConfigMapping.java
```diff
@@ -94,6 +94,10 @@ static String skewer(String camelHumps) {
 
     static String skewer(String camelHumps, int start, int end, StringBuilder b) {
         assert !camelHumps.isEmpty() : "Method seems to have an empty name";
+        if (camelHumps.startsWith("get")) {
```
dmlloyd: I disagree with this for two reasons:

1. I don't like it :)  I think naming the properties exactly is the best, especially since you could potentially have a boolean property which is a verb: `get-file-size-before-uploading` or some such.
2. The implementation allows two properties with the same name to exist, which is inherently problematic and necessitates more validation.  Not really worth the headache to support this convention.

implementation/src/main/java/io/smallrye/config/DefaultValuesConfigSource.java
```diff
@@ -1,14 +1,14 @@
-package io.smallrye.config.mapper;
+package io.smallrye.config;
 
 import java.util.Collections;
 import java.util.Map;
 
 import org.eclipse.microprofile.config.spi.ConfigSource;
 
-final class DefaultValuesConfigSource implements ConfigSource {
+public final class DefaultValuesConfigSource implements ConfigSource {
```
dmlloyd: I would rather not see this exposed in this way; rather I think it would be better to introduce an API to `SmallRyeConfigBuilder` to add default values directly, and let the mapper use that API.  I believe this is similar to a proposal I sent to the MP Config project.

implementation/src/main/java/io/smallrye/config/mapper/ConfigMapping.java
```diff
@@ -602,10 +600,8 @@ public static Builder builder() {
         return new Builder();
     }
 
-    public <B extends SmallRyeConfigBuilder> B registerDefaultValues(B builder) {
-        Assert.checkNotNullParam("builder", builder);
-        builder.withSources(new DefaultValuesConfigSource(defaultValues));
-        return builder;
+    public Map<String, String> getDefaultValues() {
+        return Collections.unmodifiableMap(defaultValues.toMap());
```
dmlloyd: This is incorrect.  The default values cannot be represented by a flat map because they are pattern-backed.  For example a map value of `foo.*.bar` could have a default value of `baz`, and so `baz` would have to be returned for any *matching* key.

implementation/src/main/java/io/smallrye/config/KeyMap.java
```diff
@@ -250,4 +251,22 @@ public StringBuilder toString(StringBuilder b) {
     public String toString() {
         return toString(new StringBuilder()).toString();
     }
+
+    public Map<String, V> toMap() {
```
dmlloyd, line 255: This will tend to result in a wrong answer because it leaves out every "any" value.

implementation/src/main/java/io/smallrye/config/DefaultValuesConfigSource.java
```diff
@@ -1,30 +0,0 @@
-package io.smallrye.config;
-
-import java.util.Collections;
-import java.util.Map;
-
-import org.eclipse.microprofile.config.spi.ConfigSource;
-
-public final class DefaultValuesConfigSource implements ConfigSource {
```
dmlloyd: Disagree: default values have to be represented by patterns thus `KeyMap`, and the set of properties it defines must be empty.

mapper/src/main/java/io/smallrye/config/mapper/CompareWith.java
```diff
@@ -1,23 +0,0 @@
...
-import java.lang.annotation.Target;
-import java.util.Comparator;
-
-/**
- * Specify the comparator to use to compare the annotated type for range.
- */
-@Documented
-@Retention(RetentionPolicy.RUNTIME)
-@Target(ElementType.TYPE_USE)
-public @interface CompareWith {
```
dmlloyd: I'd rather implement these than remove them :-)

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
radcortez: Forgetting about this, what about the integration with the builder and having `ConfigMappingProvider` as part of `SmallRyeConfig` in here: https://github.com/smallrye/smallrye-config/blob/b46864f662f9434f66d67ab4b833902dcf24f81f/implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java#L79

Is this something that you agree?

implementation/src/main/java/io/smallrye/config/mapper/ConfigMapping.java
radcortez: We can enhance this. The issue here was about properties that may not be mapped (and you don't want to map) for instance, all the properties that you get com System and Env that you normally don't use.

implementation/src/main/java/io/smallrye/config/mapper/ConfigMapping.java
radcortez: It's fine, I'll remove it. My idea here is the you may already have a POJO that you could reuse.

implementation/src/main/java/io/smallrye/config/DefaultValuesConfigSource.java
radcortez: This was just am intermediary step with the integration, since mapping had its own package.

Right now, this was already removed in favour of:
https://github.com/smallrye/smallrye-config/commit/b0b61c2f44a84d84275faa740198a234207ab2f8

Let me know if this lines up better with what you were thinking for default values.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
dmlloyd: I think it could make sense to be able to define the config interface mapping in the main builder, sure.

implementation/src/main/java/io/smallrye/config/mapper/ConfigMapping.java
radcortez: We can back this up with the `KeyMap` work, but I did implement the Default API separately first, to not depend yet on the Mapping.

implementation/src/main/java/io/smallrye/config/KeyMap.java
radcortez: I'll check.

mapper/src/main/java/io/smallrye/config/mapper/CompareWith.java
radcortez: Sure, it was just to reduce the noise and focus on what was actually implemented :)

implementation/src/main/java/io/smallrye/config/DefaultValuesConfigSource.java
radcortez: It's ok. Let me bridge this with the API I've added in `SmallRyeConfigBuilder`.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
radcortez: Ok, I'll remove the "on the fly build" and keep the builder interface.

BTW, I didn't change the Mapping Entry API, so it can still be used like you proposed originally.

implementation/src/main/java/io/smallrye/config/NameIterator.java
```diff
@@ -0,0 +1,348 @@
...
+    private static final int FS_BACKSLASH = 2;
+    private static final int FS_QUOTE_BACKSLASH = 3;
+
+    /*
+     * Iteration cookie format
+     *
+     * Bit: 14...12 11 ... 0
+     * ┌───────┬────────────┐
+     * │ state │ position │
+     * │ │ (signed) │
```
dmlloyd, line 82: I guess this lost formatting somewhere along the way :)

radcortez: I'll fix it :)

implementation/src/main/java/io/smallrye/config/mapper/ConfigurationInterface.java
```diff
@@ -0,0 +1,1063 @@
...
+import io.smallrye.config.WithDefault;
+import io.smallrye.config.WithName;
+import io.smallrye.config.inject.InjectionMessages;
+import sun.misc.Unsafe;
+
+/**
+ * Information about a configuration interface.
+ */
+public final class ConfigurationInterface {
+    static final boolean usefulDebugInfo = true;
```
dmlloyd, line 47: This should be set to the result of some system property like `io.smallrye.config.mapper.useful-debug-info` which defaults to `false` (since there's a perf cost).

implementation/src/main/java/io/smallrye/config/mapper/ConfigurationValidationException.java
```diff
@@ -0,0 +1,57 @@
+package io.smallrye.config.mapper;
+
+import io.smallrye.common.constraint.Assert;
+
+/**
+ *
```
dmlloyd, line 6:
```suggestion
 * An exception which is thrown when a configuration validation problem occurs.
```

implementation/src/main/java/io/smallrye/config/KeyMapBackedConfigSource.java
```diff
@@ -0,0 +1,38 @@
...
+    @Override
+    public Map<String, String> getProperties() {
+        return properties.toMap();
+    }
+
+    @Override
+    public String getValue(final String propertyName) {
+        final String value = properties.findRootValue(propertyName);
+        if (value == null) {
+            // In case of searching for an empty root.
```
dmlloyd, line 33: An empty root should not mean an empty initial segment, otherwise at a minimum we're creating a useless extra map.  There's nothing special about "root" values from the perspective of a key map, so if values are getting added like `".foo"` then the problem should be addressed on the add side, not this side.

implementation/src/main/java/io/smallrye/config/mapper/ConfigurationInterface.java
```diff
@@ -0,0 +1,1063 @@
...
+import io.smallrye.config.WithConverter;
+import io.smallrye.config.WithDefault;
+import io.smallrye.config.WithName;
+import io.smallrye.config.inject.InjectionMessages;
+import sun.misc.Unsafe;
+
+/**
+ * Information about a configuration interface.
+ */
+public final class ConfigurationInterface {
```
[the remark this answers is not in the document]
radcortez: I think we can make it package private. At least at first sight I don't see any restriction. Am I missing something?

radcortez:
@dmlloyd do you think we are ready to merge this and work the remaining pieces in subsequents PR's?

implementation/src/main/java/io/smallrye/config/KeyMapBackedConfigSource.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public KeyMapBackedConfigSource(final String name, final int ordinal, final KeyMap<String> properties) {
+        super(name, ordinal);
+        this.properties = properties;
+    }
+
+    @Override
+    public Map<String, String> getProperties() {
+        return properties.toMap();
```
dmlloyd, line 26: This is not correct for default values.  The way I originally implemented it was correct: backed by `KeyMap`, but `getProperties` (and thus `getPropertyNames`) should be empty.

implementation/src/main/java/io/smallrye/config/KeyMap.java
dmlloyd: IMO this is still a very deceptive method.  There is no equivalency between KeyMap and Map, the former being hierarchical and pattern based, and the latter being flat.  This will only lead to temptation to implement things incorrectly (like my comment about the incorrectness of the KeyMap-backed config source).

implementation/src/main/java/io/smallrye/config/KeyMap.java
```diff
@@ -0,0 +1,280 @@
...
+            b.append("no value");
+        } else {
+            b.append("value=").append(rootValue);
+        }
+        b.append(") {");
+        final Iterator<Entry<String, KeyMap<V>>> iterator = entrySet().iterator();
+        KeyMap<V> any = this.any;
+        if (iterator.hasNext()) {
+            Entry<String, KeyMap<V>> entry = iterator.next();
+            b.append(entry.getKey()).append("=>").append(entry.getValue());
```
dmlloyd, line 235: This is needlessly expensive.  It would be significantly better (faster and less allocation-intensive) to append the `"=>"`, and then recursively call the value's `toString(StringBuilder)` method.

implementation/src/main/java/io/smallrye/config/KeyMap.java
```diff
@@ -0,0 +1,280 @@
...
+            while (iterator.hasNext()) {
+                entry = iterator.next();
+                b.append(',').append(entry.getKey()).append("=>").append(entry.getValue());
```
dmlloyd, line 238: Same here...

implementation/src/main/java/io/smallrye/config/KeyMap.java
```diff
@@ -0,0 +1,280 @@
...
+            }
+            if (any != null) {
+                b.append(',').append("(any)=>").append(any);
```
dmlloyd, line 241: And here...

implementation/src/main/java/io/smallrye/config/KeyMap.java
```diff
@@ -0,0 +1,280 @@
...
+            }
+        } else {
+            if (any != null) {
+                b.append("(any)=>").append(any);
```
dmlloyd, line 245: And here.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -211,6 +222,26 @@ public String getRawValue(String name) {
...
 
+    @Experimental("TODO")
+    public <T> T getConfigMapping(Class<T> klass) {
+        return getConfigMapping(klass, "");
+    }
+
+    @Experimental("TODO")
+    public <T> T getConfigMapping(Class<T> klass, String prefix) {
+        try {
+            final ConfigMappingProvider.Result result = configMappingProvider.mapConfiguration(this);
```
dmlloyd, line 233: We shouldn't be re-mapping the configuration on every request; it should only be done once when the configuration is built.  If validation fails, the configuration build should fail.

implementation/src/main/java/io/smallrye/config/KeyMapBackedConfigSource.java
radcortez: Hum, why do you think that `getProperties` shouldn't also return the property if a default value is set?

dmlloyd:
I think after the last few comments are addressed it should be OK to go.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
radcortez: Hum, and how about dynamic sources? Your mapping may be valid / invalid on built time vs lookup time depending on the sources state.

radcortez:
> I think after the last few comments are addressed it should be OK to go.

Ok :)

implementation/src/main/java/io/smallrye/config/KeyMapBackedConfigSource.java
dmlloyd: Multiple reasons:

* Because the property isn't really present (that's why the default value would be returned).
* Because the value returned in this case doesn't actually match the property names that it corresponds to (for example if I have a default value for `foo.*.bar`, then the default value config source does not contain `foo.baz.bar` even though it would return a value for it; OTOH if I have a default value for `foo.bar`, the default value config source *does* contain that property - inconsistency is a form of incorrectness)
* Because it will cause confusion when determining which values are actually present in the configuration, which Quarkus presently relies upon.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
dmlloyd: This is why I made the mapper separate to begin with.  It is only valid at a point in time.  It is fundamentally incompatible with dynamic sources, and must always be, because an important part of the validation checking involves determining if there are required properties that are missing.  Allowing dynamicity in mapped configurations makes this impossible, because a property that is missing may appear at any time in a dynamic configuration environment.  Dynamic configuration mapping also makes map-backed mappings highly impractical, because every time any property is read, the entire configuration would have to be analyzed to determine what is present or missing at that point in time.

mapper/src/main/java/io/smallrye/config/mapper/CompareWith.java
radcortez: We will implement this in a subsequent PR.

implementation/src/main/java/io/smallrye/config/SmallRyeConfigMappingProvider.java
```diff
@@ -0,0 +1,38 @@
...
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import io.smallrye.config.mapper.ConfigMappingProvider;
+import io.smallrye.config.mapper.ConfigurationObject;
+import io.smallrye.config.mapper.ConfigurationValidationException;
+
+public class SmallRyeConfigMappingProvider implements Serializable {
```
dmlloyd, line 11: This class probably should not exist.  The `ConfigMappingProvider` does not need to exist once the corresponding configuration is mapped; the roots should then be associated directly with the configuration.  Otherwise this holds on to a potentially significant amount of memory for an indefinite amount of time.

Also calling this class `SmallRyeConfigMappingProvider` when it does not extend `ConfigMappingProvider` is somewhat confusing.

implementation/src/main/java/io/smallrye/config/mapper/ConfigMappingProvider.java
```diff
@@ -0,0 +1,703 @@
...
+import io.smallrye.common.function.Functions;
+import io.smallrye.config.ConfigValidationException;
+import io.smallrye.config.KeyMap;
+import io.smallrye.config.NameIterator;
+import io.smallrye.config.SmallRyeConfig;
+
+/**
+ *
+ */
+public final class ConfigMappingProvider implements Serializable {
```
dmlloyd, line 36: Given that the config mapping is now so strongly integrated within the config, I'm not sure it makes sense to have this still be public - it really looks like it's just an implementation detail of `SmallRyeConfigBuilder` at this point.

implementation/src/main/java/io/smallrye/config/mapper/ConfigurationInterface.java
```diff
@@ -0,0 +1,1063 @@
...
+    static final ClassValue<ConfigurationInterface> cv = new ClassValue<ConfigurationInterface>() {
+        protected ConfigurationInterface computeValue(final Class<?> type) {
+            return createConfigurationInterface(type);
+        }
+    };
+    static final boolean usefulDebugInfo;
+    static final Unsafe unsafe;
+
+    static {
+        usefulDebugInfo = Boolean.parseBoolean(System.getProperty("io.smallrye.config.mapper.useful-debug-info"));
```
dmlloyd, line 58: The getProperty here needs to be wrapped in an AccessController.doPrivileged.

implementation/src/main/java/io/smallrye/config/SmallRyeConfigMappingProvider.java
radcortez: I was thinking about that (I'm terrible for naming :D).

implementation/src/main/java/io/smallrye/config/mapper/ConfigMappingProvider.java
radcortez: We need to get rid of the `mapper` package to change the modifier.

radcortez:
@dmlloyd good to go? :)

implementation/src/main/java/io/smallrye/config/KeyMapBackedConfigSource.java
dmlloyd, line 26: This is still not correct.  The "toMap" method on keyMap is not meaningful in this way.  KeyMap is pattern-based, not key-value based.  Therefore `getProperties()` can never be meaningful, and `toMap()` on that method probably should not exist (it is only going to be misused).

implementation/src/main/java/io/smallrye/config/KeyMap.java
```diff
@@ -0,0 +1,285 @@
...
+        }
+        b.append('}');
+        return b;
+    }
+
+    public String toString() {
+        return toString(new StringBuilder()).toString();
+    }
+
+    public Map<String, V> toMap() {
```
dmlloyd, line 261: I would just remove this method.

implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
```diff
@@ -314,7 +329,33 @@ boolean isAddDiscoveredInterceptors() {
...
+        ConfigMappingProvider mappingProvider = mappingsBuilder.build();
+        if (!defaultValues.isEmpty() || !mappingProvider.getDefaultValues().isEmpty()) {
+            final KeyMap<String> mappingProviderDefaultValues = mappingProvider.getDefaultValues();
+            defaultValues.forEach((key, value) -> mappingProviderDefaultValues.findOrAdd(key).putRootValue(value));
+            withSources(new KeyMapBackedConfigSource("DefaultValuesConfigSource", mappingProviderDefaultValues) {
+                private static final long serialVersionUID = 7847969724478280439L;
+
+                @Override
+                public Map<String, String> getProperties() {
+                    return new HashMap<>();
```
dmlloyd, line 341: Use `Collections.emptyMap()` instead of constructing a new instance every time the method is called.

implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
```diff
@@ -314,7 +329,33 @@ boolean isAddDiscoveredInterceptors() {
 
     @Override
     public SmallRyeConfig build() {
-        return new SmallRyeConfig(this);
+        ConfigMappingProvider mappingProvider = mappingsBuilder.build();
+        if (!defaultValues.isEmpty() || !mappingProvider.getDefaultValues().isEmpty()) {
+            final KeyMap<String> mappingProviderDefaultValues = mappingProvider.getDefaultValues();
+            defaultValues.forEach((key, value) -> mappingProviderDefaultValues.findOrAdd(key).putRootValue(value));
+            withSources(new KeyMapBackedConfigSource("DefaultValuesConfigSource", mappingProviderDefaultValues) {
```
dmlloyd, line 336: This no longer needs to be an anonymous class AFAICT; the ordinal can be passed in, and the properties map should already be empty.

radcortez: Sort of. `KeyMapBackedConfigSource` is abstract. Do we want to make an implementation for this case?

dmlloyd: It probably can be made concrete then, I would think.  I don't imagine any possible subclass might do more than what the base class already does.

radcortez: Ok!

radcortez:
🎉

PR #341, opened 2020-06-29
Replace regexp on getValue with linear search and replace

pcasaes:
This work is brought on by https://github.com/smallrye/smallrye-reactive-messaging/pull/632

We saw a situation were the regexp in _EnvConfigSource#getValue_ had a severe performance impact. The issue was resolved by calling config only once but it would be nice to get rid of this regexp. On my local machine I was able to get an order of magnitude improvement on the throughput by using a simple linear search and replace.

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
+++ b/implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
@@ -20,7 +20,6 @@ import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.Collections;
 import java.util.Map;
-import java.util.regex.Pattern;
 
 import io.smallrye.config.common.AbstractConfigSource;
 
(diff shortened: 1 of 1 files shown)


radcortez:
Hi @pcasaes. Thank you.

Out of curiosity, do you have some performance numbers to share?

pcasaes:
Sure,

The original code has an average throughput of 550K r/s. The new code has an average throughput of 3M r/s. This running on a 6 core 2018 Intel i9 2.9 GHz.

This is the code 

```
        String envProp = System.getenv("SMALLRYE_MP_CONFIG_PROP");
        assertNotNull(envProp);

        ConfigSource cs = new EnvConfigSource();

        int total = 0;
        for (int i = 0; i < 10; i++) {
            long s = System.nanoTime();
            int count = 0;
            while (System.nanoTime() - s < 1_000_000_000L) {
                assertEquals(envProp, cs.getValue("smallrye.mp.config.prop"));
                assertFalse(cs.getPropertyNames().contains("smallrye.mp.config.prop"));
                count++;
            }
            total += count;
            System.out.println("ran " + count);
        }
        System.out.println("avg " + (total / 10.));
```

PR #369, opened 2020-07-30
Allow consumers of SmallRyeConfigFactory to participate in the Config building.

radcortez:
Main motivation is to be able to register via the builder ConfigMapping interfaces https://github.com/smallrye/smallrye-config/pull/333.

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/inject/ConfigExtension.java
+++ b/implementation/src/main/java/io/smallrye/config/inject/ConfigExtension.java
@@ -36,19 +36,22 @@ import javax.enterprise.inject.spi.AfterDeploymentValidation;
 import javax.enterprise.inject.spi.AnnotatedType;
 import javax.enterprise.inject.spi.BeanManager;
 import javax.enterprise.inject.spi.BeforeBeanDiscovery;
+import javax.enterprise.inject.spi.BeforeShutdown;
 import javax.enterprise.inject.spi.Extension;
 import javax.enterprise.inject.spi.InjectionPoint;
 import javax.enterprise.inject.spi.ProcessInjectionPoint;
 import javax.inject.Provider;
 
 import org.eclipse.microprofile.config.Config;
-import org.eclipse.microprofile.config.ConfigProvider;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
 import org.eclipse.microprofile.config.spi.Converter;
 
 import io.smallrye.config.ConfigValue;
 import io.smallrye.config.SecretKeys;
 import io.smallrye.config.SmallRyeConfig;
+import io.smallrye.config.SmallRyeConfigFactory;
+import io.smallrye.config.SmallRyeConfigProviderResolver;
 
 /**
  * CDI Extension to produces Config bean.
@@ -56,9 +59,18 @@ import io.smallrye.config.SmallRyeConfig;
  * @author <a href="http://jmesnil.net/">Jeff Mesnil</a> (c) 2017 Red Hat inc.
  */
 public class ConfigExtension implements Extension {
+    private final Config config;
+
     private final Set<InjectionPoint> injectionPoints = new HashSet<>();
 
     public ConfigExtension() {
+        SmallRyeConfigProviderResolver configProviderResolver = (SmallRyeConfigProviderResolver) SmallRyeConfigProviderResolver
+                .instance();
+        SmallRyeConfigFactory configFactory = configProviderResolver.getFactoryFor(getContextClassLoader(), false);
+        Config config = configFactory.getConfigFor(configProviderResolver, getContextClassLoader());
+        configProviderResolver.registerConfig(config, getContextClassLoader());
+
+        this.config = config;
     }
 
     protected void beforeBeanDiscovery(@Observes BeforeBeanDiscovery bbd, BeanManager bm) {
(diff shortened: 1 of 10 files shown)


implementation/src/main/java/io/smallrye/config/inject/ConfigExtension.java
```diff
@@ -36,29 +36,41 @@
...
  *
  * @author <a href="http://jmesnil.net/">Jeff Mesnil</a> (c) 2017 Red Hat inc.
  */
 public class ConfigExtension implements Extension {
+    private final Config config;
+
     private final Set<InjectionPoint> injectionPoints = new HashSet<>();
 
     public ConfigExtension() {
+        SmallRyeConfigProviderResolver configProviderResolver = (SmallRyeConfigProviderResolver) SmallRyeConfigProviderResolver
```
dmlloyd, line 67: This would concern me a little bit because the extension isn't necessarily the thing that should have control over instantiating the configuration.  Really the extension just provides additional information.  Otherwise you're going to have trouble in container and appserver environments which themselves set up the configuration.  Remember the spec's funny rules about what an "application" is, and how class loaders map to applications...

A better approach might be to provide a new SPI like "SmallRyeConfigConfigurator" or something like that, which allows implementations to provide extra setup.  The CDI extension could register the config objects and default values and whatever else needs to be registered at that point.

radcortez: > This would concern me a little bit because the extension isn't necessarily the thing that should have control over instantiating the configuration. 

In practice this is what usually happens, since it is the first one to call `ConfigProvider.getConfig`, which translated into the constructor block. There is indeed the possibility that someone may call `ConfigProvider.getConfig()` before the extension and register the `Config` to the same Classloader. In that case, our hands are tied, because the object is already built. Quarkus is fine here, since it doesn't use the Extension to load the `Config`. I do have to check with either Wildly and Thorntail.

> A better approach might be to provide a new SPI like "SmallRyeConfigConfigurator" or something like that, which allows implementations to provide extra setup. The CDI extension could register the config objects and default values and whatever else needs to be registered at that point.

We could provide that option, yes. It still won't fix the problem if there is already a `Config` instance registered for the Classloader. Either we have to allow the Config to mutate 🤮 or replace it 🤮. Any other bright ideas? :)

dmlloyd: If the config is created via `ConfigProvider.getConfig()`, then each configurator will have a chance to modify the config as it is built, and the configuration is created only one time.  If there is already a config registered, it would have been modified by any configurators, so there's no problem there.

radcortez: But in here, if I understood correctly, you are assuming that all the configurators are already available when the `Config` needs to be built. 

In the CDI Extension this may not be practical, since you want to scan for stuff to pass to the configurators and then build the `Config`. For instance, the case where we want to scan the Config Mappings, to register them via the builder. 

While unlikely, a `Config` object may already exist before getting into the CDI Extension and in that case, a configurator won't have any impact. Am I missing something?

dmlloyd: No, you are pointing out a real problem and it's valid AFAICT.  But I think that the answer cannot be to have the extension directly instantiate the configuration.  Somehow there has to be coordination so that there is only one configuration.

It isn't immediately clear why CDI needs to be involved at this stage.  Is this a precursor to having CDI contribute default values?

radcortez: > It isn't immediately clear why CDI needs to be involved at this stage. Is this a precursor to having CDI contribute default values?

The idea was to use CDI to scan for the Config Mapping interfaces, so we can register them in the Builder and do the validation.

radcortez: Another alternative is to do it like this:
https://github.com/radcortez/smallrye-config/commit/8e41cff0cc6e27a47bdd83dc061d0218b4269562

Is still messing with the initial internal state of Config, not directly, but still...

radcortez: @dmlloyd What do you think about something along the lines of https://github.com/radcortez/smallrye-config/commit/8e41cff0cc6e27a47bdd83dc061d0218b4269562?

dmlloyd: I guess it's moving back towards my original design where the mappings had an independent lifecycle from the config.  The config gets instantiated early, and the mappings get instantiated later when we know their topology.

radcortez: Sort of. It was a 3 way step:

1. Build mappings to retrieve the defaults and topology
2. Pass defaults to SmallRyeConfig and build
3. Validate and instantiate the mappings

As we have discussed previously, in CDI, we don't control when the Config is instantiated, so we would always have the problem to pass in the default to a possible already built Config. I guess that if we think that the DefaultsConfigSource is a dynamic ConfigSource during CDI initialisation, we are not really changing the Config object :)

dmlloyd: Yeah, I think that's probably the only choice, though it leaves a problem where property expansion might not work as expected at all stages.  But it seems unavoidable unless we can change how CDI bootstrap works (like we do in Quarkus).

radcortez: I'll check that out. Ok, I'll close this PR and open a new one with the other code.

PR #382, opened 2020-08-25
Provide a way to load the ConfigMapping implementation classes in separate steps.

radcortez:
This is mostly to help with Quarkus integration.

Prototype for Quarkus (still missing the CDI bits):
https://github.com/radcortez/quarkus/commit/9a8f24a872c72510b0b6197ea0ff9b1f49460801

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/ConfigMappingContext.java
+++ b/implementation/src/main/java/io/smallrye/config/ConfigMappingContext.java
@@ -9,7 +9,6 @@ import static io.smallrye.config.ConfigMappingInterface.rawTypeOf;
 import static io.smallrye.config.ConfigMappingInterface.typeOfParameter;
 import static io.smallrye.config.ConfigValidationException.Problem;
 
-import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.UndeclaredThrowableException;
 import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


radcortez:
@dmlloyd any thoughts?

dmlloyd:
At a glance, it looks pretty far outside of my original design parameters.  In particular, you're making a lot of private things public; surely that's a red flag?

radcortez:
I'll check if I can narrow down the public stuff.

In here, the main idea was to split the bytecode generation and the loading of the class, so we can produce it with a `GeneratedClassBuildItem`.  How were you planning to integrate it in Quarkus?

dmlloyd:
Under my original design, the mapper was separate and the API points were designed exactly to fit in to the Quarkus stages.  It would have integrated cleanly by running the generation stages during static init.

radcortez:
Hum... maybe I'm missing something. I did not change the design around the the bytecode generation (only now with this PR, with the split).

Can't remember exactly what was going on, since it was a couple of weeks ago (I can revert to check the issues), but as far as I remember, it was related with `MethodHandles` not being fully supported in Graal and our `defineClass` not working. So, I've worked around it by using the `GeneratedClassBuildItem` and removing the `defineClass` with a Substitution.

Let me see if I can find another way. Feel free to send some other ideias :)

radcortez:
I've revised this and adjusted some of the visibilities.

For now, I believe we are not able to use static init recorders. I've tried and got stuck with https://github.com/oracle/graal/issues/2734. It is fixed, but not released yet.

The only way I could find was to split the byte code generation and push it as a `GeneratedClassBuildItem`. Registration happens at runtime for validation purposes.

radcortez:
@dmlloyd let me know what you think now. Thanks!

PR #383, opened 2020-08-27
Fix thread local leak

stuartwdouglas:
As the thread local was set and not cleaned up
threads end up with a reference to the lambda
used to supply the initial value. This can
result in a ClassLoader leak.

Signed-off-by: Stuart Douglas <sdouglas@redhat.com>

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/SecretKeys.java
+++ b/implementation/src/main/java/io/smallrye/config/SecretKeys.java
@@ -4,14 +4,15 @@ import java.util.function.Supplier;
 
 @SuppressWarnings("squid:S5164")
 public final class SecretKeys {
-    private static final ThreadLocal<Boolean> LOCKED = ThreadLocal.withInitial(() -> Boolean.TRUE);
+    private static final ThreadLocal<Boolean> LOCKED = new ThreadLocal<>();
 
     private SecretKeys() {
         throw new UnsupportedOperationException();
     }
 
     public static boolean isLocked() {
-        return LOCKED.get();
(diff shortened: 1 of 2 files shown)


implementation/src/main/java/io/smallrye/config/SecretKeys.java
```diff
@@ -4,14 +4,15 @@
...
+    private static final ThreadLocal<Boolean> LOCKED = new ThreadLocal<>();
 
     private SecretKeys() {
         throw new UnsupportedOperationException();
     }
 
     public static boolean isLocked() {
-        return LOCKED.get();
+        Boolean result = LOCKED.get();
+        return result == null ? true : result;
```
dmlloyd, line 15: You probably will want to fix doUnlocked too.

stuartwdouglas: From my POV the current code is fine, it's not leaking the value that is the problem, but leaking the value attached to a ThreadLocal that has a reference to the deployment ClassLoader. 

Is it intentional that both doLocked and doUnlocked do not check the current state, and will always reset to the opposite value?

dmlloyd: The current state is checked via `isLocked` afaict.

stuartwdouglas: ah, of course. I have pushed a fix

kenfinnigan:
@dmlloyd is this good with you?

PR #402, opened 2020-09-23
Add way to iterate config properties in ConfigSourceContext.

radcortez:


vsevel:
hello @radcortez. I checked vault's code. I only need to iterate over properties names. so we should be fine with your improvement.

implementation/src/main/java/io/smallrye/config/ConfigSourceContext.java
```diff
@@ -8,4 +10,6 @@
 @Experimental("ConfigSource API Enhancements")
 public interface ConfigSourceContext {
     ConfigValue getValue(String name);
+
+    Iterator<String> iterateNames();
```
Emily-Jiang, line 14: The previous method is about one property but the new method is about a collection of the properties. Since it is contetual info for a config source, what is ConfigValue about?

Emily-Jiang (commented):
What issue is this PR trying to fix?

radcortez:
> What issue is this PR trying to fix?

This is a part of our feature to provide a way to construct a `ConfigSource` with contextual information from other sources. We are not fixing anything in particular. We are just adding a way for implementors to access the current list of properties from other sources. This will allow to filter sources by a prefix.

implementation/src/main/java/io/smallrye/config/ConfigSourceContext.java
radcortez: For the contextual property, we just return `ConfigValue` to provide additional info. In fact, internally we changed SR Config to always work with `ConfigValue` so we can have contextual information everywhere.

PR #420, opened 2020-10-13
Ensure injected Config is serializable

### LINKED ISSUE #405: An Injected Config instance should be serializable
The MP Config spec requires that an injected Config instance must be serializable.
https://github.com/eclipse/microprofile-config/issues/89

If a non-serializable ConfigSource is added to a SmallRye Config then it is no longer serializable and it fails the TCK test.
https://github.com/eclipse/microprofile-config/blob/2c42cc9f86112aef975ee5e06e90aef1dd6379df/tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigProviderTest.java#L146

The Open Liberty implementation "solved" this by always injecting a serializable proxy instance instead of the real Config. I don't particularly like this solution either but putting it here for reference;
https://github.com/OpenLiberty/open-liberty/blob/integration/dev/com.ibm.ws.microprofile.config.1.1.cdi/src/com/ibm/ws/microprofile/config/cdi/DelegatingConfig.java

Azquelt:
Ensure that an injected `Config` object is serializable, regardless of
whether the config sources it uses are serializable or not.

Fixes #405


radcortez (changes requested):
Hi @Azquelt,

Can you please rebase your commit with master? Since this change is not specific to MP Config 2.0, we can push it to master first and later rebase to mpconfig20 branch.

Thanks!

Azquelt:
Rebased to `master`. Since there's no `Config.unwrap` method on the master branch, I made `InjectedConfig.getInstance` package private and renamed it to `getDelegate` so that `ConfigProducerUtil` can still get the `SmallRyeConfig` when it needs it.

Emily-Jiang:
@radcortez does master depends on MP Config 2.0 API? Probably it is best to push mpconfig20 branch to master. Thoughts?

radcortez:
I think we need to find another way to do this.

I've just realised that anyone that is casting the instance from the Producer to `(SmallRyeConfig)` is going to break. Even in the Quarkus codebase this is a reality.

A possible solution is to move `SmallRyeConfig` to its own interface and have `InjectedConfig` to implement it as well.

Azquelt:
Sounds reasonable, I can make `SmallRyeConfig` an interface though that would still break binary compatibility with anything that currently uses `SmallRyeConfig` directly. Is that a problem?

radcortez:
Maybe. I guess we need to try it out.

Another option might be to add the behaviour as a configuration. 

Or go back to the original idea of wrapping the sources, but again you will have the same issue when calling `Config#getConfigSources()`, but maybe with less impact.

Anyway, users casting directly to an instance, is not something I want. That is why I've pushed to add `unwrap` to `Config`.  At some point we are going to need to break this, but we cannot do it now without some warning and adjusting Quarkus code.

Azquelt:
@radcortez Fair enough, here's an implementation to try out. I think existing source code will compile and run without changes, but existing compiled code will have problems.

Making it a configuration option should be straightforward, I'd need to think a bit longer about the other suggestion of wrapping the sources. There's also the somewhat nasty option of having the injected config actually extend the `SmallRyeConfig` but override all of its methods.

dmlloyd:
If the goal is to have the serialized configuration deserialize to be a new configuration, rather than preserving the old configuration as-is (the previous behavior), then the best approach is probably to use `writeReplace()` and `readResolve()` to store an object which replaces itself with a local configuration on deserialize.

Azquelt:
> If the goal is to have the serialized configuration deserialize to be a new configuration, rather than preserving the old configuration as-is...

Sort of. For a general `Config` object, I don't think we do want that. It could have been created by the user with a builder with a different set of sources or converters.

However, the `Config` bean that gets injected is always whatever `Config` object is registered for the current context classloader (a default one is created if there is none registered), so I think it's sufficient to just look that up again.

The spec only requires that the injected `Config` bean is `Serializable`, not that all `Config` objects are serializable. It also doesn't require that config sources are serializable.

radcortez:
Thinking a bit more about this, is the only reason to support `SessionScoped` beans? 

I wonder if we should relax the spec. With the addition of `@ConfigProperties`, it should simplify injection at `SessionScoped` beans, making that the preferred way to inject configurations. 

@Azquelt regarding the current state of the PR, I think we also going to run into issues with this. I'm wondering if we should convert `SmallRyeConfig` to the current `InjectedConfig` and then delegate to a new class that has the current code of `SmallRyeConfig`?

Azquelt:
> @Azquelt regarding the current state of the PR, I think we also going to run into issues with this. I'm wondering if we should convert `SmallRyeConfig` to the current `InjectedConfig` and then delegate to a new class that has the current code of `SmallRyeConfig`?

I think that would work, I'll give it a go. If we're creating a config manually, we can fill in the delgate at creation time, otherwise we can create it with no delegate and it'll be looked up later. Should also probably have a flag to say how it gets serialized in that case.

Azquelt:
@radcortez I've changed it so that `SmallRyeConfig` is now a class which delegates to `SmallRyeConfigImpl`. Can you see if this works for you?

implementation/src/main/java/io/smallrye/config/inject/SecuritySupport.java
```diff
@@ -24,11 +24,11 @@
 /**
  * @author <a href="http://jmesnil.net/">Jeff Mesnil</a> (c) 2018 Red Hat inc.
  */
-class SecuritySupport {
+public class SecuritySupport {
```
kenfinnigan, line 27: I don't think this class is intended to be public. Probably needs to be copied to another package if it's needed, or moved.

@dmlloyd ??

dmlloyd: This definitely should *not* be public, as that will cause a CVE.

Azquelt: Ah sorry, I changed that when I was refactoring and forgot to go back and investigate what should really be done with it. Looks like there is a `SecuritySupport` class in the package it's moved to and it should be using that one.

radcortez:
I think with this approach with are able to keep compatibility with our consumers. I've tried it with Quarkus and run some of the most significant tests around config without issues, but I want to full the full test build.

Anyway, this is a somehow delicate change. @dmlloyd @kenfinnigan do you have any thoughts overall?

dmlloyd:
It still feels over-complicated to me.  I just don't think there needs to be an entire alternative implementation class.  At *most* it should be delegating to the original class for all operations and otherwise using `writeReplace/readResolve` to relocate the original, I would think.

kenfinnigan:
Can anyone clarify what the use case for `Config` being Serializable is?

Is it actually necessary?

radcortez:
> Can anyone clarify what the use case for `Config` being Serializable is?
> 
> Is it actually necessary?

When injecting a `Config` instance in a `SessionScopedBean`.

radcortez:
Actually in most cases this will work, the issue in when a custom `ConfigSource` is parte of the `Config` which is not `Serializable`.

dmlloyd:
The significant question seems to revolve around whether it is expected for the configuration to deserialize to a copy of the configuration that existed on the serializing side, or whether it should deserialize as the actual current configuration that's already in place on the deserializing side.  And, I suppose, whether we want this behavior to be uniform for all circumstances in Quarkus or we want this behavior to only apply to CDI usage.

kenfinnigan:
If this is purely for `@SessionScoped` usage, then I'd say we don't want it.

radcortez:
@Azquelt was there any other case were you required the `Config` to be `Serializable`?

Azquelt:
It's an explicit requirement in the spec, just stated as
> The `Config` instances created via CDI are `Serializable`.

However, the rationale was to support injection into session scoped beans: https://github.com/eclipse/microprofile-config/issues/89.

More generally, I think serializability would be required for injection into a bean from any passivating scope, the only built-in ones are SessionScoped and ConversationScoped.

dmlloyd:
> More generally, I think serializability would be required for injection into a bean from any passivating scope, the only built-in ones are SessionScoped and ConversationScoped.

In which case we almost certainly want deserialized configurations to represent the current configuration, not the one that was on the serializing side.  In fact it's hard to imagine a case where you wouldn't want this to be the behavior; I almost think we should just do something like this:

```java
public class SmallRyeConfig ... implements ... Serializable {
    ...
    Object writeReplace() { return new Ser(); }
}

final Ser implements Serializable {
    Object readResolve() {
        return ConfigProvider.getConfig();
    }
}
```

Maybe even in the specification.  What good is it to serialize the entire configuration *ever*?  Not to mention having that burden on converters and config sources... it doesn't really make sense to do it that way.

Azquelt:
@dmlloyd That would certainly work though the downside would be that any `Config` object would be serialized as the current configuration.

Users can use the `ConfigBuilder` API to create their own `Config` object with a different set of config sources, converters and interceptors. It would seem odd that if I built my own `Config`, serialized it and then deserialized it, I would get back the current configuration, rather than the one I had built.

In practise this might be fairly rare, but this was why my initial PR added a new object which would be used when the `Config` was injected as bean. By using a different class, we get the desired serialization behaviour for serializing an injected `Config` without applying it to all `Config` objects since they might not represent the current config.

dmlloyd:
If you wanted to ship a configuration to another node, serializing it is almost certainly the worst possible way to do it.  It's too fragile and represents a lot of configuration that won't actually be properly serialized anyway, like environment variables and system properties, and possibly other config sources that are relative to the local environment anyway, so what you get back will already be a hodgepodge mix of configurations.  More likely what you'd want to do is produce e.g. a properties listing of some sort, or serialize an individual configuration source.

I don't think the overall serialization mechanism is useful in any case.

Azquelt:
I think I agree with all of that. I'd prefer a solution where:
* `SmallRyeConfig` is not `Serializable`
* When a `Config` is injected, we use a different class which is serializable and delegates to whatever is the current config

Then we would be saying that you cannot serialize an arbitrary `Config` and that deserialising an injected `Config` is the same as just getting the current `Config`.

However, that runs up against the following constraints of the current code:
* `SmallRyeConfig` _is_ currently `Serializable`
* Existing code assumes that any `Config` can be safely cast to a `SmallRyeConfig`, so an injected `Config` must be a `SmallRyeConfig`.

As you suggested, we could make it so that when you deserialize a `SmallRyeConfig` you always get the current config, regardless of the object which was serialized. This would give strange behaviour if a user creates and serializes their own `Config` object, but would give correct behaviour if an injected `Config` is serialized.

Possibly we could work around the strange behaviour by setting a flag to say whether a given `SmallRyeConfig` has been registered as the default config for any classloader and preventing serialization if that isn't the case? Technically `ConfigProviderResolver.registerConfig()` is part of the API and would allow users to change the default config, but it is documented that they shouldn't call it.

radcortez:
Maybe we could simplify this at the spec level and drop the requirement of `Config` being Serializable. In most cases this is not useful and now with `@ConfigProperties` coming in MP Config 2.0, this should be the preferred approach to inject multiple properties in a bean.

Azquelt:
I've pushed a new change which does the following:
* If a config has been registered as the config for a classloader, then we serialize a placeholder object and look up the current config when we deserialize.
* Otherwise we serialize the config as before.

@dmlloyd was right, this is a lot simpler :slightly_smiling_face: 

This meets the spec requirements, maintains the existing serialization behaviour for a custom config and passes the existing tests.

We could also consider:
* Not supporting serialization of custom configs
* Always serizalizing a placeholder object and looking up the current config when deserializing.

In either of these cases, we'd have to remove `ConfigSerializationTest`

radcortez:
@Azquelt Thanks! Sorry for the delay in the response.

> * Always serizalizing a placeholder object and looking up the current config when deserializing.
> 
> In either of these cases, we'd have to remove `ConfigSerializationTest`

I think we should go with that approach and drop the register check. It is fine to change the `ConfigSerializationTest`. The only reason I added it was to be able to catch serialization issues before getting into the TCK and to be able to debug it.

Azquelt:
Thanks, I've pushed an update to make that change.

radcortez:
Thanks. Looks good. Sorry for all the back and forward.

radcortez (approved):

PR #421, opened 2020-10-13
Fix #417 - Added test to check config_ordinal override in microprofile-config.properties.

### LINKED ISSUE #417: Setting `config_ordinal` property in META-INF/microprofile-config.properties should change the ordinal of the PropertiesConfigSource
Hi @Joseph-Cass  and @radcortez,

The fix for issue https://github.com/smallrye/smallrye-config/issues/401 does not completely close this issue https://github.com/eclipse/microprofile-config/issues/573.

Actually setting 'config_ordinal' in `META-INF/microprofile-config.properties` file still does not override the `PropertiesConfigSource` object ordinal.

radcortez:


Emily-Jiang (approved):
LGTM!

PR #423, opened 2020-10-14
Trimming converter should throw NPE

### LINKED ISSUE #422: Builtin converters must throw NPE if asked to convert null
eclipse/microprofile-config#542 clarified that it's invalid to pass `null` to a converter and added a requirement that all built in converters must throw a `NullPointerException` if this occurs.

Currently smallrye-config doesn't do this because the `TrimmingConverter` will return `null` if passed `null`.

I'm happy to make a PR to fix this, but I'm not sure if it should be against `master` or `mpconfig20` - both branches have the issue, but the requirement was only added to the spec in 2.0.

Note the test for this was broken in the TCK and wasn't running but has been fixed by eclipse/microprofile-config#628

Azquelt:
The correct behaviour for a converter which is passed null is to throw an NPE rather than handling it gracefully.

Fixes #422

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/Converters.java
+++ b/implementation/src/main/java/io/smallrye/config/Converters.java
@@ -919,7 +919,7 @@ public final class Converters {
         }
 
         public T convert(final String value) {
-            return value == null ? null : getDelegate().convert(value.trim());
+            return getDelegate().convert(value.trim());
         }
     }
 }


Azquelt:
There is a test for this in the TCK, but I can add another if desired.

radcortez:
Closing this, since it is already implemented in `mpconfig20` branch. Thank you!

Issue #409, opened 2020-10-02
Certain validation failures cause further injection points not to be checked


Azquelt:
In `ConfigExtension.validate`, there is clearly an attempt to detect all validation issues and add each of them as a deployment problem.

However, there are cases which will cause the validate method to throw an exception or return, meaning that it won't validate further injection points.

In particular:
* [Line 140 can `return`](https://github.com/smallrye/smallrye-config/blob/mpconfig20/implementation/src/main/java/io/smallrye/config/inject/ConfigExtension.java#L140)
* [Line 144, `getPropertyKey()` can throw an exception](https://github.com/smallrye/smallrye-config/blob/mpconfig20/implementation/src/main/java/io/smallrye/config/inject/ConfigExtension.java#L144)

(line numbers based on the `mpconfig20` branch)

radcortez:
https://github.com/smallrye/smallrye-config/pull/412/ Should fix this.

On `io.smallrye.config.inject.ConfigProducerUtil#getConfigKey`, I believe it should be fine. Did you have any trouble with it?

Azquelt:
`io.smallrye.config.inject.ConfigProducerUtil#getConfigKey` will throw an exception if the `name` isn't set for the `ConfigProperty` annotation and it can't work out a default name (e.g. because the injection point is a method parameter and the parameter name is not available).

However, `ConfigExtension.validate` won't catch this exception and therefore won't validate any further injection points.

radcortez:
Ah, right! Forgot about the method parameter injection. I'll fix it. Thank you!

radcortez:
Actually, it should be fine. Any exception thrown in the `AfterDeploymentValidation` observer event is added as a deployment problem.

radcortez:
Here is the test: https://github.com/smallrye/smallrye-config/blob/master/implementation/src/test/java/io/smallrye/config/inject/ValidateInjectionTest.java#L62-L72

Azquelt:
This is true, and the error will be reported, but no further validation will be done. I think it would be preferable if we could report all the validation issues, rather than stopping if we hit this one. I'll make a PR for this.

radcortez:
Sure.

PR #424, opened 2020-10-15
Keep validating if injection point name missing

Azquelt:
If an injection point name is missing, record the error and continue
validating, rather than allowing the exception to be thrown.

Enhancement to #409

### DIFF
--- a/implementation/src/test/java/io/smallrye/config/inject/ValidateInjectionTest.java
+++ b/implementation/src/test/java/io/smallrye/config/inject/ValidateInjectionTest.java
@@ -71,6 +71,23 @@ public class ValidateInjectionTest {
                         "")));
     }
 
+    @Test
+    void methodUnnamedProperties() {
+        JupiterTestEngine engine = new JupiterTestEngine();
+        LauncherDiscoveryRequest request = request().selectors(selectClass(MethodUnnamedPropertiesTest.class)).build();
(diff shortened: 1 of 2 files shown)


radcortez (approved):

implementation/src/test/java/io/smallrye/config/inject/ValidateInjectionTest.java
```diff
@@ -71,6 +71,24 @@ void constructorUnnamedProperties() {
                         "")));
     }
 
+    @Test
+    void methodUnnamedProperties() {
+        JupiterTestEngine engine = new JupiterTestEngine();
+        LauncherDiscoveryRequest request = request().selectors(selectClass(MethodUnnamedPropertiesTest.class)).build();
+        EngineExecutionResults results = EngineTestKit.execute(engine, request);
+        System.out.println("Hello!");
```
radcortez, line 79: You may want to remove this sout :)

radcortez (changes requested):

implementation/src/test/java/io/smallrye/config/inject/ValidateInjectionTest.java
Azquelt: Rats. Thanks

radcortez: Thanks!

radcortez (approved):

PR #431, opened 2020-10-25
Prototype to support loading additional configurations locations.

radcortez:
A lot of users have been requesting the ability to load external configuration files:
https://github.com/quarkusio/quarkus/issues/12419
https://github.com/quarkusio/quarkus/issues/11273
https://github.com/quarkusio/quarkus/issues/9869

Also it seems to be a recurring topic in Zulip:
https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/application.2Eproperties.20file.20from.20file.20system
https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/External.20properties
https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Shared.2Fglobal.20application.2Eproperties.20in.20Maven.20multimodule
https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Additional.20config.20properties.20at.20build.20time
https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Quarkus.20configuration

The configuration `smallrye.config.locations` sets additional locations to load configurations. These can be a file, a directory contents, a class path resource or an http resource. Additional locations are only added if the config is set. If not, there is no change on how the config behaves. Since this requires the config to be the factories are auto registered with the Config object, so no additional setup is required.

This is just a prototype and requires some additional polishing.


radcortez:
@xtaixe, @hrstoyanov @mklueh any chance for any of you to try this out and give me some feedback?

You should be able to use this with Quarkus 1.9.0 and with updated to the following dependencies:

```xml
<dependency>
  <groupId>io.smallrye.config</groupId>
  <artifactId>smallrye-config</artifactId>
  <version>1.9.3-SNAPSHOT</version>
</dependency>

<dependency>
  <groupId>io.smallrye.common</groupId>
  <artifactId>smallrye-common-classloader</artifactId>
  <version>1.4.1</version>
</dependency>

<dependency>
  <groupId>io.smallrye.config</groupId>
  <artifactId>smallrye-config-source-additional-properties</artifactId>
  <version>1.9.3-SNAPSHOT</version>
</dependency>
```

Thanks!

xtaixe:
@radcortez tested. It works for me with a `@QuarkusTest` and with the runner jar, but I couldn't test with native because when trying to build the native image I get

```
Error: could not find target method: static java.lang.Class io.quarkus.runtime.configuration.Substitutions$Target_ConfigMappingLoader.loadClass(java.lang.String,byte[])
com.oracle.svm.core.util.UserError$UserException: could not find target method: static java.lang.Class io.quarkus.runtime.configuration.Substitutions$Target_ConfigMappingLoader.loadClass(java.lang.String,byte[])
	at com.oracle.svm.core.util.UserError.abort(UserError.java:68)
	at com.oracle.svm.hosted.substitute.AnnotationSubstitutionProcessor.findOriginalMethod(AnnotationSubstitutionProcessor.java:637)
```

which I'm not getting without the changes to test this. Just FYI.

As a feature, this would work for us.

Thanks for starting the work on this.

radcortez:
@xtaixe Thanks for trying this out. 

Yes, the latest master also changed a few things in the API that affect the native image generation, but these is already handled here: https://github.com/quarkusio/quarkus/pull/12501

So no worries about the native image, this will be fully working once released.

radcortez:
@dmlloyd any thoughts?

dmlloyd:
It seems OK in principle.  We would absolutely have to document the exact rules for which configuration files take precedence and in what order though.  There could be a lot of user confusion in this regard.

radcortez:
I agree. I think that loading additional sources this way should never take precedence over system properties or environment variables. It should be ok to haver an higher priority than `application.properties` or `microprofile-config.properties`.

For order, it may be a little trickier. If you set `config_ordinal` in the source it would be very well defined. If not, it falls in our rule of ordering by ordinal and then by name (which gives you some predictability, but maybe not what you want). We could add a extra rule to take into account the definition order in the configuration.

Emily-Jiang:
There is a similar issue in the MicroProfile Config github. I think it is a good idea. I agree that we can only process .properties files to start with. 

I am thinking this property `smallrye.config.location` could be specified in a config source. Basically, you can put `smallrye.config.location` in your microprofile-config.properties. In this case, the ordinal in the properties files should inherit the ordinal of the config source, where the property is specified if the property `config_ordinal` is not specified in the property files.

radcortez:
> I am thinking this property `smallrye.config.location` could be specified in a config source. 

Yes, I added support to set it up with any other valid source.

> Basically, you can put `smallrye.config.location` in your microprofile-config.properties. In this case, the ordinal in the properties files should inherit the ordinal of the config source, where the property is specified if the property `config_ordinal` is not specified in the property files.

This might be a little tricky. If the additional file has the same ordinal of the defining source, what happens if both files define the same property name with different values? Right now we do a second sort by the `ConfigSource` name, but it is not ideal.

Emily-Jiang:
> This might be a little tricky. If the additional file has the same ordinal of the defining source, what happens if both files define the same property name with different values? Right now we do a second sort by the `ConfigSource` name, but it is not ideal.

If they have the same ordinal, the order is undefined. In the impl, we could say the additional file overrides the parent one as the additional file is somewhere and can be seen to override the property value in the parent.

dmlloyd:
> > This might be a little tricky. If the additional file has the same ordinal of the defining source, what happens if both files define the same property name with different values? Right now we do a second sort by the `ConfigSource` name, but it is not ideal.
> 
> If they have the same ordinal, the order is undefined. In the impl, we could say the additional file overrides the parent one as the additional file is somewhere and can be seen to override the property value in the parent.

A reasonable implementation rule would have been to order first by ordinal, and then by load order (our container implementations have always had a stable class loading order).  The second sort by config source name is really not *functionally* useful, being intended only to bring stability to the sort, but I think that it is actually detrimental in practice: it *removes* the much more useful stability of the initial load order that otherwise would have taken effect.

radcortez:
Sounds reasonable. I don't think anyone is relying on the name sort, so I'll make some changes to use the loading order sort.

radcortez:
Requires release of SR Commons 1.4.2.

Emily-Jiang:
The PR sounds reasonable to me!

radcortez:
@dmlloyd are you ok with this? :)

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
```diff
@@ -0,0 +1,237 @@
...
+            }
+        }
+
+        return configSources;
+    }
+
+    private void tryFileSystem(final URL url, final List<ConfigSource> configSources, final ConfigSourceContext context) {
+        if (url.getProtocol().startsWith("file")) {
+            // Load single file and profiles
+            final Path urlPath = Paths.get(url.getFile()).toAbsolutePath().normalize();
```
dmlloyd, line 80: I don't think this is necessary. The path should simply be the path of the URL, which is already absolute.  Otherwise you get into problems: calling `toAbsolutePath()` actually accesses the filesystem recursively trying to locate the root path, so if you're under a symlinked directory you'll get a different path than expected.

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
```diff
@@ -0,0 +1,237 @@
...
+    @Override
+    public Iterable<ConfigSource> getConfigSources(final ConfigSourceContext context) {
+        final ConfigValue value = context.getValue(SMALLRYE_LOCATIONS);
+
+        final List<ConfigSource> configSources = new ArrayList<>();
+        if (value.getValue() != null) {
+            List<URL> resources = newCollectionConverter(new URLConverter(), ArrayList::new).convert(value.getValue());
+            if (resources != null) {
+                for (URL url : resources) {
+                    tryFileSystem(url, configSources, context);
```
dmlloyd, line 67: Wouldn't it make more sense to have an `if` tree or string `switch` on the URL scheme?

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
```diff
@@ -0,0 +1,237 @@
...
+        return this.loadConfigSource(url, ConfigSource.DEFAULT_ORDINAL);
+    }
+
+    @Override
+    public Iterable<ConfigSource> getConfigSources(final ConfigSourceContext context) {
+        final ConfigValue value = context.getValue(SMALLRYE_LOCATIONS);
+
+        final List<ConfigSource> configSources = new ArrayList<>();
+        if (value.getValue() != null) {
+            List<URL> resources = newCollectionConverter(new URLConverter(), ArrayList::new).convert(value.getValue());
```
dmlloyd, line 64: I think it's probably better to use the more modern `URI` here.  URIs support relative paths correctly for example, and they're more compatible with `Path`.  Also they don't have the equals/hashCode issues associated with URL.  I can't imagine users wanting to put a full blown URL for every filesystem location.

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
```diff
@@ -0,0 +1,237 @@
...
+    protected abstract String[] getFileExtensions();
+
+    protected abstract ConfigSource loadConfigSource(final URL url, final int ordinal) throws IOException;
+
+    protected ConfigSource loadConfigSource(final URL url) throws IOException {
+        return this.loadConfigSource(url, ConfigSource.DEFAULT_ORDINAL);
+    }
+
+    @Override
+    public Iterable<ConfigSource> getConfigSources(final ConfigSourceContext context) {
```
dmlloyd, line 59: When did we make the switch to returning multiple config sources here?  It means that it is impossible to assemble a chain in priority order because the config sources would have to be reshuffled by priority after the chain is built - meaning that each factory will ultimately not be "looking at" lower ordinal config sources as they are being built, which was a big part of the intent of the suggestion.  If this was my idea, then slap me. :-)

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
```diff
@@ -0,0 +1,237 @@
...
+                            .forEach(path -> addConfigSource(toURL(path.toUri()), configSources));
+                } catch (IOException e) {
+                    throw ConfigMessages.msg.failedToLoadResource(e);
+                }
+            }
+        }
+    }
+
+    private void tryClassPath(final URL url, final List<ConfigSource> configSources, final ConfigSourceContext context) {
+        if (url.getProtocol().startsWith("file")) {
```
dmlloyd, line 97: Using relative URIs, you can check to see if the URI has a `null` scheme _or_ the scheme is equal to `file`; in the former case, it's a relative URI.

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
```diff
@@ -0,0 +1,237 @@
...
+            } else {
+                return new URL(fileName + "-" + profile);
+            }
+        } catch (MalformedURLException e) {
+            // The original URL was already validated, so this shouldn't happen.
+            throw new IllegalStateException(e);
+        }
+    }
+
+    private static class URLConverter implements Converter<URL> {
```
dmlloyd, line 202: This is a general :-1: from me; we should strongly prefer `URI`.

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
```diff
@@ -0,0 +1,237 @@
...
+
+import io.smallrye.common.annotation.Experimental;
+
+/**
+ * This {@code AbstractLocationConfigSourceFactory} allows to initialize additional config locations with the
+ * configuration {@link AbstractLocationConfigSourceFactory#SMALLRYE_LOCATIONS}.
+ * <p>
+ *
+ * Locations set in {@link AbstractLocationConfigSourceFactory#SMALLRYE_LOCATIONS} are loaded in order and from the
+ * following resources:
```
dmlloyd, line 32: I'm not sure this doc is very clear.  It's probably better to say something like "the locations comprise a list of URIs which are loaded in order.  The following URI schemes are supported:"

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
```diff
@@ -0,0 +1,237 @@
...
+
+        return configSources;
+    }
+
+    private void tryFileSystem(final URL url, final List<ConfigSource> configSources, final ConfigSourceContext context) {
+        if (url.getProtocol().startsWith("file")) {
+            // Load single file and profiles
+            final Path urlPath = Paths.get(url.getFile()).toAbsolutePath().normalize();
+            if (Files.isRegularFile(urlPath)) {
+                consumeAsPath(toURL(urlPath.toUri()), new ConfigSourcePathConsumer(this, configSources, context));
```
dmlloyd, line 82: All of this juggling between URI and URL is an additional indication that we should strongly prefer URI.

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
```diff
@@ -0,0 +1,237 @@
...
+            } else if (Files.isDirectory(urlPath)) { // Load directory contents
+                try (Stream<Path> paths = Files.walk(urlPath, 1)) {
```
dmlloyd, line 84: This is less critical, but using DirectoryStream is preferred here just because it's much simpler and uses _far_ fewer objects in the heap - especially if you are only recursing one level deep.  You can wrap it in try-with-resources and for-each constructs, and are less likely to be bitten by potential bugs and leaks.

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
```diff
@@ -0,0 +1,237 @@
...
+            }
+        }
+        return false;
+    }
+
+    private static URL addProfileToFileName(final URL fileName, final String profile) {
+        final int dot = fileName.toString().lastIndexOf(".");
+        try {
+            if (dot != -1) {
+                return new URL(fileName.toString().substring(0, dot) + "-" + profile + fileName.toString().substring(dot));
```
dmlloyd, line 192: Building URLs like this is strongly discouraged (think "URL injection").  This is another reason to prefer URI: it will ensure that the encoding of the URL is done correctly, something that is definitely not done here.

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
radcortez: We ended up changing the priority on the factory to be used only for initialization purposes. The main reason is to be able to honour the `config_ordinal` property if set in the `ConfigSource`. So, yes we do initialize with the initial chain and then reshuffle all the sources in the right order.

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
radcortez: I've changed to use `URI`. Still some juggling is needed since the `ClassPathUtils` uses `URL` and the `toURL` is just to avoid catching the Exception on `URI.toURL`, since we know it is valid.

radcortez: Also, we need to keep the abstract method to use an `URL`, since we don't know if the classloader resources set a custom URL handler, and we will loose if we convert it to an `URI`. Everything else is using `URI` now.

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
```diff
@@ -0,0 +1,247 @@
...
+        return configSources;
+    }
+
+    private void tryFileSystem(final URI uri, final ConfigSourceContext context, final List<ConfigSource> configSources) {
+        final Path urlPath = Paths.get(uri.getPath());
+        if (Files.isRegularFile(urlPath)) {
+            consumeAsPath(toURL(urlPath.toUri()), new ConfigSourcePathConsumer(this, context, configSources));
+        } else if (Files.isDirectory(urlPath)) {
+            try (DirectoryStream<Path> paths = Files.newDirectoryStream(urlPath, this::validExtension)) {
+                paths.forEach(path -> addConfigSource(path.toUri(), configSources));
```
dmlloyd, line 91: By for-each I was referring to [this](https://docs.oracle.com/javase/8/docs/technotes/guides/language/foreach.html), but this is OK too (basically the same overhead though in one case we have an iterator and in the other we have a lambda).

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
```diff
@@ -0,0 +1,247 @@
...
+
+        final int dot = fileName.lastIndexOf(".");
+        final String fileNameProfile;
+        if (dot != -1) {
+            fileNameProfile = fileName.substring(0, dot) + "-" + profile + fileName.substring(dot);
+        } else {
+            fileNameProfile = fileName + "-" + profile;
+        }
+
+        return URI.create(uri.toString().replace(fileName, fileNameProfile));
```
dmlloyd, line 208: This doesn't really solve any of the quoting issues and in fact can blow up in various humorous ways.  You probably want something safer like this:
```suggestion
        return new URI(uri.getScheme(), uri.getAuthority(), uri.getHost(), uri.getPort(), fileNameProfile, uri.getQuery(), uri.getFragment());
```

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
```diff
@@ -0,0 +1,247 @@
...
+            }
+        }
+
+        return configSources;
+    }
+
+    private void tryFileSystem(final URI uri, final ConfigSourceContext context, final List<ConfigSource> configSources) {
+        final Path urlPath = Paths.get(uri.getPath());
+        if (Files.isRegularFile(urlPath)) {
+            consumeAsPath(toURL(urlPath.toUri()), new ConfigSourcePathConsumer(this, context, configSources));
```
dmlloyd, line 88: Why not this here:

```suggestion
            consumeAsPath(uri.getPath(), new ConfigSourcePathConsumer(this, context, configSources));
```

dmlloyd: If you're always passing `this` in as the first parameter of a nested class constructor, it probably should just be an inner class.

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    private static class URIConverter implements Converter<URI> {
+        private static final long serialVersionUID = -4852082279190307320L;
+
+        @Override
+        public URI convert(final String value) {
+            try {
+                // TODO - validate path
+                return new URI(value);
```
dmlloyd, line 218: Just as an FYI, you can use `URI.create()` instead here, in which case instead of throwing `URISyntaxException` it will throw an `IllegalArgumentException`.  In fact, using this method, your converter *could* simply be `URI::create`.

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    private void addConfigSource(final URI uri, final List<ConfigSource> configSources) {
+        addConfigSource(toURL(uri), configSources);
+    }
+
+    private void addConfigSource(final URL url, final List<ConfigSource> configSources) {
+        try {
+            final ConfigSource configSource = loadConfigSource(url);
+            if (!configSource.getProperties().isEmpty()) {
```
dmlloyd, line 162: This might be an expensive operation since getProperties() isn't always supported, right?  I'd expect it to be *more likely* that `getPropertyNames().isEmpty()` would be performant - in the worst case it's just a view on `getProperties()`.

dmlloyd:
There are probably more cleanups possible, because it still seems to me that we're doing a lot of unneeded hopping back and forth between forms of URI and other APIs, but I didn't see anything else that is actually *problematic*.

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
radcortez: Sorry, for some reason I've got use to just call `.forEach`. I guess I need to rewind that :)

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
radcortez: Forgot to add a comment here or a TODO for us to discuss. I've just added this to move forward.

Your suggestion was my first implementation, but it doesn't work for the `jar` scheme, because the path is stored in the  `schemeSpecificPart`, so we need some additional logic to cover that case.

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
radcortez: > Why not this here:

Because we don't have a `consumeAsPath` in the `ClassPathUtils` that takes a `String`. We can probably add it. The other methods that do take a `String` handle it as a classpath resource and search the classloader. In this case, we just want to load this single resource.

implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceFactory.java
radcortez: Sure, but since this `Converter` is applied to user input, we probably want to catch and throw the exception with our own custom message.

radcortez:
> There are probably more cleanups possible, because it still seems to me that we're doing a lot of unneeded hopping back and forth between forms of URI and other APIs, but I didn't see anything else that is actually _problematic_.

I'll try to have another look, but in the load of the `ConfigSource` itself we need to keep `URL` because a classloader resource may contain a custom `URLHandler`.

radcortez:
Also, it could probably be improved by adjusting the API in `ClassPathUtils`.

hrstoyanov:
Is this going to be part of 1.10 release?

radcortez:
> Is this going to be part of 1.10 release?

Unlikely, since Quarkus 1.10 is already in the final preparation stage, we usually refrain to introduce major updates to dependencies.

Still, SR Config should be completely compatible once we release, so you can probably update the dependency manually in your project to use the feature until it is GA in Quarkus itself.

Sorry for taking so much time :(

radcortez:
@dmlloyd are you ok with this? :)

radcortez:
In the meanwhile, I've made some additional changes: I've moved the loading code to its own class, so it can be used anywhere and not just in the Factory.

hrstoyanov:
@radcortez  Now that Quarkus 1.11.CR1 is out, how can I use this in Quarkus? Is there a better version of [this tutorial](https://quarkus.io/guides/config)? Does this work with YAML configurations? A simple example for alternative YAML file location?

I am also interested in how to point Quarkus to alternative configuration during unit testing

radcortez:
Hi @hrstoyanov, please refer to this:
https://smallrye.io/docs/smallrye-config/config/config.html#locations

Just use `smallrye.config.locations` to set alternate locations. This works as any other configuration property, so you can use it as an env variable, system property, etc. YAML should also be supported.

Let me know if you need any help.

PR #432, opened 2020-10-26
Do not generate class mapping bytes if not required.

radcortez:
For https://github.com/quarkusio/quarkus/issues/12960

### DIFF
--- a/implementation/src/test/java/io/smallrye/config/ConfigMappingLoaderTest.java
+++ b/implementation/src/test/java/io/smallrye/config/ConfigMappingLoaderTest.java
@@ -26,8 +26,7 @@ class ConfigMappingLoaderTest {
     void loadManually() {
         List<ConfigMappingMetadata> configMappingsMetadata = ConfigMappingLoader.getConfigMappingsMetadata(ServerManual.class);
         configMappingsMetadata.forEach(
-                mappingMetadata -> ConfigMappingLoader.loadClass(ServerManual.class,
-                        mappingMetadata.getClassName(), mappingMetadata.getClassBytes()));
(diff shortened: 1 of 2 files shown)


implementation/src/test/java/io/smallrye/config/ConfigMappingLoaderTest.java
```diff
@@ -26,8 +26,7 @@ void multipleLoads() {
     void loadManually() {
         List<ConfigMappingMetadata> configMappingsMetadata = ConfigMappingLoader.getConfigMappingsMetadata(ServerManual.class);
         configMappingsMetadata.forEach(
-                mappingMetadata -> ConfigMappingLoader.loadClass(ServerManual.class,
-                        mappingMetadata.getClassName(), mappingMetadata.getClassBytes()));
+                mappingMetadata -> ConfigMappingLoader.loadClass(ServerManual.class, mappingMetadata));
```
gsmet, line 29: This is typically a case where a lambda is totally useless and costs memory for nothing.

gastaldi: Nice catch, but this is a test and will never go into production ;)

radcortez: Yes, but I guess the comment was more general, since we are abusing lambdas on the runtime code :)

I'm refactoring some pieces to remove them.

gsmet: @gastaldi sure but it was to illustrate this issue: https://github.com/quarkusio/quarkus/issues/12960#issuecomment-716733836

radcortez:
This also requires an additional substations on Quarkus side. I'll open a PR with that change once the SR Config is out.

PR #441, opened 2020-11-03
Cache null values in EnvConfigSource.

### LINKED ISSUE #440: Should EnvConfigSource cache results which are not found?
I saw that #396 added a cache for entries in the env config source to avoid doing the character replacement on the key on every lookup and to align with Quarkus.

However, I note that it doesn't add anything to the cache if the key is not found in the config source, meaning that repeated lookups for the same key which isn't present anywhere will result in the character replacement on the key being repeated on every lookup. This happens relatively frequently because other microprofile specs allow the user to change the default behaviour by setting config properties and so implementations must check to see if they're set.

Our performance testing picked this up as being slower than other config sources and I notice that [EnvConfigSource][env-config-source] in Quarkus does cache missing values using a `NULL_SENTINAL`.

Is there a reason that missing values are not cached in the smallrye-config `EnvConfigSource`?

[env-config-source]:https://github.com/quarkusio/quarkus/blob/master/core/runtime/src/main/java/io/quarkus/runtime/configuration/ConfigUtils.java#L126

radcortez:
* Fixes #440

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
+++ b/implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
@@ -141,14 +141,14 @@ public class EnvConfigSource extends AbstractConfigSource {
         return DEFAULT_ORDINAL;
     }
 
-    protected Object writeReplace() {
+    Object writeReplace() {
         return new Ser();
     }
 
     static final class Ser implements Serializable {
         private static final long serialVersionUID = 6812312718645271331L;
 
-        protected Object readResolve() {
+        Object readResolve() {
             return new EnvConfigSource();
         }
     }
--- a/implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
+++ b/implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
@@ -20,6 +20,7 @@ import static io.smallrye.config.common.utils.ConfigSourceUtil.CONFIG_ORDINAL_KE
 import static java.security.AccessController.doPrivileged;
 import static java.util.Collections.unmodifiableMap;
 
+import java.io.Serializable;
 import java.security.PrivilegedAction;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
@@ -31,9 +32,11 @@ import io.smallrye.config.common.AbstractConfigSource;
  */
 public class EnvConfigSource extends AbstractConfigSource {
     private static final long serialVersionUID = -4525015934376795496L;
-    private static final int DEFAULT_ORDINAL = 300;
 
-    private final Map<String, String> cache = new ConcurrentHashMap<>(); //the regex match is expensive
+    private static final int DEFAULT_ORDINAL = 300;
+    private static final Object NULL_VALUE = new Object();
+
+    private final Map<String, Object> cache = new ConcurrentHashMap<>();
 
     protected EnvConfigSource() {
         super("EnvConfigSource", getEnvOrdinal());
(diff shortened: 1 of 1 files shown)


implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
```diff
@@ -31,9 +31,11 @@
  */
 public class EnvConfigSource extends AbstractConfigSource {
     private static final long serialVersionUID = -4525015934376795496L;
+
     private static final int DEFAULT_ORDINAL = 300;
+    private static final String NULL_VALUE = "io.smallrye.config.EnvConfigSource.nullValue";
```
dmlloyd, line 36: Rather than rely on a unique string value - which could (at least theoretically) be a legitimate value - what about using an object with unique identity instead?

```suggestion
    private static final String NULL_VALUE = new String("");
```

Then compare by identity:
```java
if (cachedValue == NULL_VALUE) {
```

Alternatively, a unique object of type `Object` may be used (this works best if the value type of the `Map` be `Object`, which would require that the result of the `get` method must be cast on return - but this cast happens internally anyway so it doesn't impact performance or anything of that sort).

radcortez: I actually had that code before, but it fails if you serialize the source. 

I didn't want to make the cache transient, but maybe we need to do it. What do you think?

radcortez: Well probably a Serializable object should work.

dmlloyd: For serialization we probably want a different strategy.  Serializing the cache from one system and deserializing it on another doesn't make any sense; we probably want replacement objects (again).  I'll add a suggestion down below.

implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
```diff
@@ -80,6 +85,7 @@ public String getValue(String name) {
             return value;
         }
 
+        cache.put(name, NULL_VALUE);
         return null;
     }
```
dmlloyd, line 91: Something like this:

```suggestion
    Object writeReplace() {
        return new Ser();
    }

    static final class Ser implements Serializable {
        private static final long serialVersionUID = something;

        Object readResolve() {
            return new EnvConfigSource();
        }
    }
```

radcortez:
Have a look. I think I prefer the `Object` approach to the empty `String` because you can't tell on the debugger which one is actually null or a real empty. That was also another reason I've used a complex `String`.

implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
```diff
@@ -133,4 +140,16 @@ private static int getEnvOrdinal() {
...
+    protected Object writeReplace() {
+        return new Ser();
+    }
+
+    static final class Ser implements Serializable {
+        private static final long serialVersionUID = 6812312718645271331L;
+
+        protected Object readResolve() {
+            return new EnvConfigSource();
+        }
```
dmlloyd, line 153:
```suggestion
    Object writeReplace() {
        return new Ser();
    }

    static final class Ser implements Serializable {
        private static final long serialVersionUID = 6812312718645271331L;

        Object readResolve() {
            return new EnvConfigSource();
        }
```

Serialization replacement methods should not be visible to subclasses or anyone outside of the API.

dmlloyd (approved):

PR #442, opened 2020-11-04
Support compound names in YAML Config Source.

radcortez:
Proposal to support https://github.com/quarkusio/quarkus/issues/11744.


sources/yaml/src/main/java/io/smallrye/config/source/yaml/YamlConfigSource.java
```diff
@@ -117,6 +117,9 @@ private static void flattenYaml(String path, Map<Object, Object> source, Map<Str
 
             if (value instanceof String) {
                 target.put(key, (String) value);
+                if (key.contains("\"")) {
```
Emily-Jiang, line 120: I might be wrong here... I am wondering why not just removing \" on line 109. I don't see the significance of adding the quotes in the first place.

PR #478, opened 2021-01-07
Support .env sources.

radcortez:


radcortez:
Closing, because GH doesn't not seem to detect that some of these commits are already in master. Opening a new PR.

PR #489, opened 2021-01-22
Support Indexed Properties

### LINKED ISSUE #436: SmallRye Config to cater for multi nested List of properties
Description
SmallRye Config to cater for multi nested List of properties

Implementation Idea
Example of how the application.properties/application.yaml file can be set up with nested properties:
eg application.properties:
```
application.environments[0].name=dev
application.environments[0].service[0].name=service1
application.environments[0].service[1].name=service2
application.environments[0].service[2].name=service3
```
eg application.yaml:
```
application:
  environments:
    - name: dev
      services:
        - name: service1
        - name: service2
        - name: service3
```
We then can have a Java ConfigMapping Class as follows:

```
@ConfigMapping(prefix = "application.environments")
public interface AppConfiguration {
  @WithParentName
  List<Environment> environments();

  interface Environment {
    
    @WithParentName
    String name();

    @WithParentName
    List<Service> services();

    interface Service {

      @WithParentName
      String name();
    }
  }
}
```

### LINKED ISSUE #447: Support for Map<String, T> and List<T>
It would useful to support custom converters for `Map<String, T>` and `List<T>`, specifically when using the yaml format.
For instance:
```
    deploy:
      entities:
        - name: one
        - name: two
          option: foo
```

with:

```
public class DeployConfig {
    public List<DeployEntityConfig> entities;
}
public class DeployEntityConfig {

    public String name;
    public Optional<String> option;
}
```

if I try to run this as-is, I get: `SRCFG00013: No Converter registered for class ....deploy.config.DeployEntityConfig`

If I create a custom converter, the convert method gets called on `{"entities": [{"name": "one"}`, then on ` {"name": "two"`, then on ` "option": "foo"}]}`, which is not appropriate to create `DeployEntityConfig`.

If I change the design to be:
```
    deploy:
      entities:
        one: {}
        two:
          option: foo
```
and
```
public class DeployConfig {
    public Map<String, DeployEntityConfig> entities;
}
public class DeployEntityConfig {

    public Optional<String> option;
}
```

I get the error `SRCFG00013: No Converter registered for interface java.util.Map`.

The closest I can get is encoding the structure in a property string. For instance a json document: 
```
    deploy:
      entities:
        - {"name": "one"}
        - {"name": "two", "option": "foo"}
```
But I think we can do better, especially when using the yaml configuration flavor.

radcortez:
* Fixes #436
* Fixes #447

Mappings of Groups in Collections is done by using the `[I]` index in the property to map. Example:

```properties
server.environments[0].name=dev
```

Maps to:

```java
@ConfigMapping(prefix = "server")
public interface ServerCollectionsSet {
    Set<Environment> environments();

    interface Environment {
        String name();
    }
}
```

In the case of comma separated values that map directly to a Collection type by the MP Config spec, it works as usual, but these can also be defined with indexed properties, but the MP Config way has priority:

```properties
server.environments=dev,prod
server.environments[0]=dev
server.environments[1]=prod
```

This also works for the `YamlConfigSource`, since the `yaml` content is serialized in this format. So, in practice this will work for any `ConfigSource` that provides their values in an indexed fashion.

A cool side effect is that these indexed properties can be overridden as usual in higher ordinal sources, per element.

### DIFF
--- a/sources/yaml/src/test/resources/example-collections.yml
+++ b/sources/yaml/src/test/resources/example-collections.yml
@@ -1,0 +1,14 @@
+application:
+  environments:
+    - name: dev
+      services:
+        - name: batch
+        - name: rest
+    - name: prod
+      services:
+        - name: web
+        - name: batch
+        - name: rest
+  images:
+    - base
+    - jdk
--- a/doc/modules/ROOT/pages/config/config.adoc
+++ b/doc/modules/ROOT/pages/config/config.adoc
@@ -5,9 +5,12 @@ include::../attributes.adoc[]
 
 = Config
 
+* <<indexed-properties>>
 * <<profiles>>
 * <<locations>>
 * <<secret-keys>>
+
+include::indexed-properties.adoc[]
 
 include::profiles.adoc[]
 
--- a/doc/modules/ROOT/pages/config/indexed-properties.adoc
+++ b/doc/modules/ROOT/pages/config/indexed-properties.adoc
@@ -1,0 +1,31 @@
+[[indexed-properties]]
+== Indexed Properties
+
+In Microprofile Config, a config value which contains unescaped commas may be converted to `Collection`. This works
+for simple cases, but it becomes cumbersome and limited for more advanced cases.
+
+Indexed Properties provide a way to use indexes in config property names to map specific elements in a `Collection`
+type. Since the indexed element is part of the property name and not contained in the value, this can also be used to
+map complex object types as `Collectionª elements. Consider:
+
+[source,properties]
+----
+# MicroProfile Config - Collection Values
+my.collection=dog,cat,turtle
+
+# SmallRye Config - Indexed Property
+my.indexed.collection[0]=dog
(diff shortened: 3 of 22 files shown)


sources/yaml/src/test/resources/example-collections.yml
```diff
@@ -0,0 +1,11 @@
+application:
```
vsevel, line 1: do you support also collection of native types?
```
application:
  elements:
    - foo
    - bar
...
```

radcortez: Not yet. That is something I need to figure out. The  issue is that any `Collection` type with a type supported natively by MP Config with a `Converter` will have priority and that is not what we want to do here.

radcortez: I've pushed some new code that should provide the foundation to support this as well. There are still some details I'll need to figure out about the YAMLSource.

vsevel:
excellent @radcortez

Issue #486, opened 2021-01-18
Certain Exceptions block other Exceptions being logged


Joseph-Cass:
Similarly to https://github.com/smallrye/smallrye-config/issues/409, when certain `Exceptions` occur, they can block other `Exceptions` being returned to the user.

In particular, I'm seeing a Property Expression Exception, `SRCFG00011`, blocking some messages from other invalid `@ConfigProperty` Injections.

Joseph-Cass:
I'm looking into this now

radcortez:
OK, let me know.

Joseph-Cass:
@radcortez Some other things have come up so I don't think I'll be able to have a look at this for a day or two at least. I'm still happy to look at it though.

radcortez:
We are not in a rush, so take your time :)

PR #493, opened 2021-01-27
Add catch block to Injection validation

Joseph-Cass:
I tried to add an Injection Message (e.g. SRCFG02008) for bad Property Expression Injections (i.e. Injecting a Config Property where the Property Expression value `${x}` doesn't exist) though I couldn't figure out how to propagate the missing value (`${missing.prop}` in the test) through, so thought throwing `SRCFG00011` would be better. Would be happy to hear what others think and to see whether this is possible and worth it.


Joseph-Cass:
Resolves issue #486

radcortez (approved):

PR #507, opened 2021-02-15
Backport from 2.x

radcortez:


implementation/src/main/java/io/smallrye/config/ConfigMapping.java
```diff
@@ -40,4 +40,23 @@
...
+
+    /**
+     * The naming strategy to use for the config mapping. This only matters for method names that contain both
+     * lower case and upper case characters.
+     *
+     * @return the config mapping naming strategy.
+     */
+    NamingStrategy namingStrategy() default NamingStrategy.KEBAB_CASE;
+
+    enum NamingStrategy {
```
svenhaag, line 52: Hi, I think this should be expandable with custom naming strategies. We are using underscores for instance.

radcortez: You mean you provide your own custom strategy? Or just add one that supports underscore?

svenhaag: I meant it would be nice if the code would allow for custom providers (e.g. via SPI?).
However, a default UNDERSCORE variant would be nice too ;).

sources/yaml/src/main/java/io/smallrye/config/source/yaml/YamlConfigSourceProvider.java
```diff
@@ -1,41 +1,36 @@
...
-        final List<ConfigSource> configSources = new ArrayList<>(2);
-        getConfigSource(classLoader, META_INF_MICROPROFILE_CONFIG_RESOURCE, YamlConfigSource.ORDINAL + 10)
-                .ifPresent(configSources::add);
-        getConfigSource(classLoader, WEB_INF_MICROPROFILE_CONFIG_RESOURCE, YamlConfigSource.ORDINAL)
-                .ifPresent(configSources::add);
-        return Collections.unmodifiableList(configSources);
+        final List<ConfigSource> sources = new ArrayList<>();
+        sources.addAll(loadConfigSources("META-INF/microprofile-config.yaml", classLoader));
+        sources.addAll(loadConfigSources("META-INF/microprofile-config.yml", classLoader));
+        sources.addAll(loadConfigSources("WEB-INF/classes/META-INF/microprofile-config.yaml", classLoader));
```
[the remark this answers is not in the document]
radcortez: Hum, yes. I guess we need to do the same for the YAML source. Thanks for noticing it.

implementation/src/main/java/io/smallrye/config/ConfigMapping.java
radcortez: Ok, let me see what I can do :)

Issue #485, opened 2021-01-18
More detailed Error messages


Joseph-Cass:
Could we include some more detailed logging messages to help users debug exceptions? Particularly for `Exception`s caused by `@Injection`s, and also MicroProfile Config 2.0 specific features.

For example, Injecting a `@ConfigProperties` bean where invalidPrefix.akey doesn't exist, e.g.:
```
@Inject
@ConfigProperties(prefix = "invaldPrefix")
ConfigPropertiesPOJO details;
```
Leads to: `NoSuchElementException: SRCFG00014: Property invaldPrefix.akey is required but the value was not found or is empty`

Could this include `@ConfigProperties` specific information? e.g. `No Config Property exists for the prefix <prefix name> for the field <field name>`

Also, I think we should ensure any MicroProfile Config key terms are used and capitalised where appropriate, e.g. the following could be `Required Config Property %s not found`:
https://github.com/smallrye/smallrye-config/blob/bb3d4ab40c720a34b11b6112dabf56690137fe1d/cdi/src/main/java/io/smallrye/config/inject/InjectionMessages.java#L34

Joseph-Cass:
I also think all the following should say which `Config Property` caused the `Exception`:
https://github.com/smallrye/smallrye-config/blob/bb3d4ab40c720a34b11b6112dabf56690137fe1d/implementation/src/main/java/io/smallrye/config/ConfigMessages.java#L105-L115

radcortez:
Sure. Can you one a PR with a proposal?

Joseph-Cass:
This is still on my todo list, but it keep getting pushed down :( 

I'd like to go through all the messages thoroughly rather than just change the ones I've highlighted

Joseph-Cass:
Another potential logging message improvement has been highlighted:
```
@Inject
    @ConfigProperty(name = "some.properties", defaultValue = "")
    private List<String> someProps;
```
Returns an error that some.properties wasn't set, which is somewhat confusing because a defaultValue was "set". We should probably try to improve that error message to state that empty strings are treated as missing values.

radcortez:
No worries, take your time :)

Joseph-Cass:
[...]

PR #513, opened 2021-02-26
Add more detail to Injection error messages

Joseph-Cass:
Resolves most of #485, implementing the suggested approach here: https://github.com/smallrye/smallrye-config/issues/485#issuecomment-784017775.

There are a few bits I was unsure about, which I'll add comments for

### DIFF
--- a/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
+++ b/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
@@ -149,7 +149,9 @@ public class ConfigExtension implements Extension {
             try {
                 name = ConfigProducerUtil.getConfigKey(injectionPoint, configProperty);
             } catch (IllegalStateException e) {
-                adv.addDeploymentProblem(InjectionMessages.msg.retrieveConfigFailure(null, e.getLocalizedMessage(), e));
+                adv.addDeploymentProblem(
+                        InjectionMessages.msg.retrieveConfigFailure(null, injectionPoint.getMember().getName(),
+                                e.getLocalizedMessage(), null, e));
                 continue;
             }
 
@@ -159,7 +161,8 @@ public class ConfigExtension implements Extension {
             if ((!configNames.contains(name) && ConfigProducerUtil.getRawValue(name, config) == null)
                     && !isIndexed(type, name, config)) {
                 if (configProperty.defaultValue().equals(ConfigProperty.UNCONFIGURED_VALUE)) {
-                    adv.addDeploymentProblem(InjectionMessages.msg.noConfigValue(name));
+                    adv.addDeploymentProblem(
+                            InjectionMessages.msg.noConfigValue(name, injectionPoint.getMember().getName(), name));
                     continue;
                 }
             }
@@ -168,7 +171,8 @@ public class ConfigExtension implements Extension {
                 // Check if the value can be injected. This may cause duplicated config reads (to validate and to inject).
                 ConfigProducerUtil.getValue(injectionPoint, config);
             } catch (Exception e) {
-                adv.addDeploymentProblem(InjectionMessages.msg.retrieveConfigFailure(name, e.getLocalizedMessage(), e));
+                adv.addDeploymentProblem(InjectionMessages.msg.retrieveConfigFailure(name, injectionPoint.getMember().getName(),
+                        e.getLocalizedMessage(), name, e));
             }
         }
 
--- a/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
+++ b/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
@@ -149,7 +149,7 @@ public class ConfigExtension implements Extension {
             try {
                 name = ConfigProducerUtil.getConfigKey(injectionPoint, configProperty);
             } catch (IllegalStateException e) {
-                adv.addDeploymentProblem(e);
+                adv.addDeploymentProblem(InjectionMessages.msg.retrieveConfigFailure(null, e.getLocalizedMessage(), e));
                 continue;
             }
 
@@ -167,10 +167,8 @@ public class ConfigExtension implements Extension {
             try {
                 // Check if the value can be injected. This may cause duplicated config reads (to validate and to inject).
                 ConfigProducerUtil.getValue(injectionPoint, config);
-            } catch (IllegalArgumentException e) {
-                adv.addDeploymentProblem(InjectionMessages.msg.illegalConversion(name, type));
-            } catch (NoSuchElementException e) {
-                adv.addDeploymentProblem(e);
+            } catch (Exception e) {
+                adv.addDeploymentProblem(InjectionMessages.msg.retrieveConfigFailure(name, e.getLocalizedMessage(), e));
             }
         }
 
@@ -188,7 +186,8 @@ public class ConfigExtension implements Extension {
         try {
             ConfigMappings.registerConfigMappings((SmallRyeConfig) config, configMappingsWithPrefix);
         } catch (ConfigValidationException e) {
-            adv.addDeploymentProblem(e);
+            adv.addDeploymentProblem(InjectionMessages.msg.retrieveConfigPropertiesFailure(e.getLocalizedMessage(), e));
+
         }
     }
 
--- a/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
+++ b/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
@@ -27,7 +27,6 @@ import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.HashSet;
 import java.util.List;
-import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.OptionalDouble;
 import java.util.OptionalInt;
--- a/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
+++ b/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
@@ -151,7 +151,7 @@ public class ConfigExtension implements Extension {
             } catch (IllegalStateException e) {
                 adv.addDeploymentProblem(
                         InjectionMessages.msg.retrieveConfigFailure(null, formatInjectionPoint(injectionPoint),
-                                e.getLocalizedMessage(), null, e));
+                                e.getLocalizedMessage(), e));
                 continue;
             }
 
@@ -162,7 +162,7 @@ public class ConfigExtension implements Extension {
                     && !isIndexed(type, name, config)) {
                 if (configProperty.defaultValue().equals(ConfigProperty.UNCONFIGURED_VALUE)) {
                     adv.addDeploymentProblem(
-                            InjectionMessages.msg.noConfigValue(name, formatInjectionPoint(injectionPoint), name));
+                            InjectionMessages.msg.noConfigValue(name, formatInjectionPoint(injectionPoint)));
                     continue;
                 }
             }
@@ -172,7 +172,7 @@ public class ConfigExtension implements Extension {
                 ConfigProducerUtil.getValue(injectionPoint, config);
             } catch (Exception e) {
                 adv.addDeploymentProblem(InjectionMessages.msg.retrieveConfigFailure(name, formatInjectionPoint(injectionPoint),
-                        e.getLocalizedMessage(), name, e));
+                        e.getLocalizedMessage(), e));
             }
         }
 
--- a/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
+++ b/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
@@ -25,6 +25,7 @@ import static java.util.stream.Collectors.toSet;
 
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
@@ -176,22 +177,35 @@ public class ConfigExtension implements Extension {
             }
         }
 
-        Set<ConfigMappings.ConfigMappingWithPrefix> configMappingsWithPrefix = new HashSet<>();
         for (AnnotatedType<?> configMapping : configMappings) {
-            configMappingsWithPrefix
-                    .add(configMappingWithPrefix(configMapping.getJavaClass(), getPrefixFromType(configMapping)));
+            registerAndValidateConfigMapping(
+                    configMappingWithPrefix(configMapping.getJavaClass(), getPrefixFromType(configMapping)), adv,
+                    (SmallRyeConfig) config);
         }
 
         for (InjectionPoint injectionPoint : configMappingInjectionPoints) {
-            getPrefixFromInjectionPoint(injectionPoint).ifPresent(prefix -> configMappingsWithPrefix
-                    .add(configMappingWithPrefix((Class<?>) injectionPoint.getType(), prefix)));
+            getPrefixFromInjectionPoint(injectionPoint).ifPresent(prefix -> registerAndValidateConfigMapping(
+                    configMappingWithPrefix((Class<?>) injectionPoint.getType(), prefix), adv, (SmallRyeConfig) config));
         }
 
-        try {
-            ConfigMappings.registerConfigMappings((SmallRyeConfig) config, configMappingsWithPrefix);
-        } catch (ConfigValidationException e) {
-            adv.addDeploymentProblem(InjectionMessages.msg.retrieveConfigPropertiesFailure(e.getLocalizedMessage(), e));
+    }
 
+    private void registerAndValidateConfigMapping(ConfigMappings.ConfigMappingWithPrefix cm, AfterDeploymentValidation adv,
+            SmallRyeConfig config) {
+        try {
+            ConfigMappings.registerConfigMappings(config, Collections.singleton(cm));
+        } catch (ConfigValidationException e) {
+            if (cm.getKlass().getAnnotation(ConfigMapping.class) != null) {
+                adv.addDeploymentProblem(
+                        InjectionMessages.msg.retrieveConfigMappingsFailure(cm.getPrefix(), cm.getKlass().getName(),
+                                e.getLocalizedMessage(), e));
+            } else if (cm.getKlass().getAnnotation(ConfigProperties.class) != null) {
+                adv.addDeploymentProblem(
+                        InjectionMessages.msg.retrieveConfigPropertiesFailure(cm.getPrefix(), cm.getKlass().getName(),
+                                e.getLocalizedMessage(), e));
+            } else {
+                adv.addDeploymentProblem(e);
+            }
         }
     }
 
--- a/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
+++ b/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
@@ -19,6 +19,7 @@ import static io.smallrye.config.ConfigMappings.ConfigMappingWithPrefix.configMa
 import static io.smallrye.config.inject.ConfigMappingInjectionBean.getPrefixFromInjectionPoint;
 import static io.smallrye.config.inject.ConfigMappingInjectionBean.getPrefixFromType;
 import static io.smallrye.config.inject.ConfigProducer.isClassHandledByConfigProducer;
+import static io.smallrye.config.inject.InjectionMessagesUtil.formatInjectionPoint;
 import static io.smallrye.config.inject.SecuritySupport.getContextClassLoader;
 import static java.util.stream.Collectors.toSet;
 
@@ -150,7 +151,7 @@ public class ConfigExtension implements Extension {
                 name = ConfigProducerUtil.getConfigKey(injectionPoint, configProperty);
             } catch (IllegalStateException e) {
                 adv.addDeploymentProblem(
-                        InjectionMessages.msg.retrieveConfigFailure(null, injectionPoint.getMember().getName(),
+                        InjectionMessages.msg.retrieveConfigFailure(null, formatInjectionPoint(injectionPoint),
                                 e.getLocalizedMessage(), null, e));
                 continue;
             }
@@ -162,7 +163,7 @@ public class ConfigExtension implements Extension {
                     && !isIndexed(type, name, config)) {
                 if (configProperty.defaultValue().equals(ConfigProperty.UNCONFIGURED_VALUE)) {
                     adv.addDeploymentProblem(
-                            InjectionMessages.msg.noConfigValue(name, injectionPoint.getMember().getName(), name));
+                            InjectionMessages.msg.noConfigValue(name, formatInjectionPoint(injectionPoint), name));
                     continue;
                 }
             }
@@ -171,7 +172,7 @@ public class ConfigExtension implements Extension {
                 // Check if the value can be injected. This may cause duplicated config reads (to validate and to inject).
                 ConfigProducerUtil.getValue(injectionPoint, config);
             } catch (Exception e) {
-                adv.addDeploymentProblem(InjectionMessages.msg.retrieveConfigFailure(name, injectionPoint.getMember().getName(),
+                adv.addDeploymentProblem(InjectionMessages.msg.retrieveConfigFailure(name, formatInjectionPoint(injectionPoint),
                         e.getLocalizedMessage(), name, e));
             }
         }
--- a/cdi/src/main/java/io/smallrye/config/inject/ConfigInjectionException.java
+++ b/cdi/src/main/java/io/smallrye/config/inject/ConfigInjectionException.java
@@ -22,18 +22,23 @@ package io.smallrye.config.inject;
  * 
  * @author Steve Moyer - The Pennsylvania State University
  */
-public class ConfigException extends Exception {
+public class ConfigInjectionException extends Exception {
 
     private static final long serialVersionUID = 1L;
 
     private final String configPropertyName;
 
-    public ConfigException(String configPropertyName, String message) {
+    public ConfigInjectionException(String message, String configPropertyName) {
         super(message);
         this.configPropertyName = configPropertyName;
     }
 
-    public ConfigException(String configPropertyName, String message, Throwable cause) {
+    public ConfigInjectionException(String message, String configPropertyName, Throwable cause) {
+        super(message, cause);
+        this.configPropertyName = configPropertyName;
+    }
+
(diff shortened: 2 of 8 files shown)


implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -219,6 +205,30 @@
...
+     * 2. {@link NoSuchElementException} - if the property is not defined <br>
+     * 3. {@link NoSuchElementException} - if the property is defined as an empty string <br>
+     * 4. {@link NoSuchElementException} - if the {@link Converter} returns {@code null} <br>
+     * <br>
+     * 
+     * Calls for converting optional values ({@link Config#getOptionalValue} and "Injecting Optional Values")
+     * should only throw an {@link Exception} for #1 ({@link IllegalArgumentException} when the property cannot be converted to
+     * the specified type).
+     */
+    public static <T> T convertValue(String name, String value, Converter<T> converter) {
```
Joseph-Cass, line 230: Could this code live in `smallrye-config-common`?

cdi/src/test/java/io/smallrye/config/inject/ValidateInjectionTest.java
```diff
@@ -304,8 +562,67 @@ void fail() {
...
+        @WeldSetup
+        WeldInitiator weld = WeldInitiator.from(ConfigExtension.class, ServerDetailsBean.class)
+                .addBeans()
+                .activate(ApplicationScoped.class)
+                .inject(this)
+                .build();
+
+        @Inject
+        // Unqualified due to @ConfigProperties missing here
+        ServerDetailsBean server;
```
Joseph-Cass, line 614: I'm not sure this test is actually testing config-injection functionality.

Isn't it just checking how Weld handles a bad Injection?

Joseph-Cass: The current output is just:
```
org.jboss.weld.exceptions.IllegalArgumentException: WELD-001408: Unsatisfied dependencies ...
...
Caused by: org.jboss.weld.exceptions.DeploymentException: WELD-001408: Unsatisfied dependencies ...
...
```

cdi/src/test/java/io/smallrye/config/inject/ValidateInjectionTest.java
```diff
@@ -26,68 +28,212 @@
...
+    void MissingConverter() {
+        DeploymentException exception = getDeploymentException(MissingConverterTest.class);
+        assertThat(exception).hasMessageStartingWith(
+                "SRCFG02001: Failed to Inject @ConfigProperty for key my.prop SRCFG02006:");
+
+        assertThat(exception.getCause()).isInstanceOf(ConfigInjectionException.class);
+
+        assertThat(exception.getCause().getCause()).isInstanceOf(java.lang.IllegalArgumentException.class);
+        assertThat(exception.getCause().getCause()).hasMessage(
+                "SRCFG02006: No Converter registered for class io.smallrye.config.inject.ValidateInjectionTest$MissingConverterTest$MyType");
```
Joseph-Cass, line 142: This message is not a ConfigMessages (SRCFG0000X) but it could be with a fair bit of re-shuffling of the smallrye-config-core code. Although I doubt think it would be worth it...

cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
```diff
@@ -184,7 +181,8 @@ protected void validate(@Observes AfterDeploymentValidation adv) {
         try {
             ConfigMappings.registerConfigMappings((SmallRyeConfig) config, configMappingsWithPrefix);
         } catch (ConfigValidationException e) {
-            adv.addDeploymentProblem(e);
+            adv.addDeploymentProblem(InjectionMessages.msg.retrieveConfigPropertiesFailure(e.getLocalizedMessage(), e));
```
Joseph-Cass, line 184: Is it fair to assume that these validation checks are exclusively for @ConfigProperties property values?

i.e. throwing an exception message saying there was an issue creating a @ConfigProperties bean

cdi/src/main/java/io/smallrye/config/inject/InjectionMessages.java
```diff
@@ -11,35 +10,29 @@
...
-    @Message(id = 2000, value = "No Config Value exists for required property %s")
-    ConfigException noConfigValue(@Param @Pos(1) String key);
+    @Message(id = 2000, value = "Failed to Inject @ConfigProperty for key %s since the config property could not be found in any config source")
+    ConfigInjectionException noConfigValue(@Param @Pos(1) String key);
 
-    @Signature(messageIndex = 1, value = { String.class, String.class, Throwable.class })
-    @Message(id = 2001, value = "Failed to retrieve config for key %s")
-    ConfigException retrieveConfigFailure(@Cause IllegalArgumentException cause, @Param @Pos(1) String key);
+    @Message(id = 2001, value = "Failed to Inject @ConfigProperty for key %s %s")
+    ConfigInjectionException retrieveConfigFailure(@Param @Pos(1) String key, @Param @Pos(2) String cuase, @Cause Exception e);
```
Joseph-Cass, line 22: This could probably include the target type too, if that would be useful?
```
@Message(id = 2001, value = "Failed to Inject @ConfigProperty for key %s to type %s %s")
ConfigInjectionException retrieveConfigFailure(@Param @Pos(1) String key, @Param @Pos(2) String type , @Param @Pos(3) String cuase, @Cause Exception e);
```

Joseph-Cass, line 22: Also just noticed a typo. I'll correct this with any other suggestions.
```suggestion
    ConfigInjectionException retrieveConfigFailure(@Param @Pos(1) String key, @Param @Pos(2) String cause, @Cause Exception e);
```

Azquelt: I'm not sure what you mean by "type" here. If mean the type of the injection point (e.g. are we injecting this property as a `String` or as a `boolean`) then I don't think that's necessary.

However, I think it would be useful to say where we were injecting into (i.e. the field or method parameter annotated with `@ConfigProperty`) because that's where the user will have to look if they're not sure of the purpose of the missing property.

Joseph-Cass: Yes, I meant the type of the injection point :)

Interesting suggestion about the location of the `InjectionPoint`. Is there a way to get a nice concise description of this? I know the InjectionPoint.toString() is a bit nasty and long.

cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
radcortez: Yes. This is valid for @ConfigProperties and @ConfigMapping.

cdi/src/main/java/io/smallrye/config/inject/InjectionMessages.java
radcortez: Try `InjectionPoint.getMember().getName()` and see how that looks like.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
radcortez: Why do we need to slip the method?

Joseph-Cass: So that it can be called from `smallrye-config-inject` to reduce duplicate code: https://github.com/smallrye/smallrye-config/blob/72a1c0bb48a683c2e50f43743b3f23cf98fd3db4/cdi/src/main/java/io/smallrye/config/inject/ConfigProducerUtil.java#L52

Joseph-Cass:
I've added the `InjectionPoint.getMember().getName()` to the `@configProperty` Exception messages. Let me know what you think @radcortez @Azquelt 

I've also removed the `@Param @Pos()` code from the `InjectionMessages.java` since:
1. I thought it was very difficult to read
2. I couldn't get it working when I was trying to add `String location`. Not sure if it was a bug or me doing something wrong, but I thought it would just be easier to remove it.

cdi/src/main/java/io/smallrye/config/inject/InjectionMessages.java
```diff
@@ -10,36 +9,30 @@
...
-    ConfigException noConfigValue(@Param @Pos(1) String key);
+    @Message(id = 2000, value = "Failed to Inject @ConfigProperty for key %s into %s since the config property could not be found in any config source")
+    ConfigInjectionException noConfigValue(String key, String location, @Param String configPropertyName);
 
-    @Signature(messageIndex = 1, value = { String.class, String.class, Throwable.class })
-    @Message(id = 2001, value = "Failed to retrieve config for key %s")
-    ConfigException retrieveConfigFailure(@Cause IllegalArgumentException cause, @Param @Pos(1) String key);
+    @Message(id = 2001, value = "Failed to Inject @ConfigProperty for key %s into %s %s")
+    ConfigInjectionException retrieveConfigFailure(String key, String location, String causeMessage,
+            @Param String configPropertyName, @Cause Exception e);
```
Azquelt, line 22: If I'm reading this correctly, you now have to pass in the key twice (once as `key` and once as `configPropertyName`)? I guess this is because you've removed `@Param @Pos(1)` from `key`?

If you do need to pass in the same property twice, it would be good to note that as it's weird.

cdi/src/test/java/io/smallrye/config/inject/ValidateInjectionTest.java
```diff
@@ -26,68 +28,214 @@
...
+        assertThat(exception.getCause().getCause()).hasMessage(
+                "SRCFG00040: The config property empty.property is defined as the empty String (\"\") which the following Converter considered to be null: io.smallrye.config.Converters$BuiltInConverter");
+    }
+
+    @Test
+    void badProperty() throws Exception {
+        DeploymentException exception = getDeploymentException(BadPropertyTest.class);
+        assertThat(exception)
+                .hasMessageStartingWith(
+                        "SRCFG02001: Failed to Inject @ConfigProperty for key bad.property into badProp SRCFG00041:");
```
Azquelt, line 114: I don't think this is enough information for the user to identify the injection location.

I would use the following to identify the location:
* For a field: `com.example.MyClass.myField`
* For a method parameter: `com.example.MyClass.myMethod(String, int, AnotherClass)`
* For a constructor parameter: `com.example.MyClass(String, int, AnotherClass)`

This might require either a helper method to do the conversion from a `Member` to a string, or a helper class and  [`@FormatWith`](https://jboss-logging.github.io/jboss-logging-tools/apidocs/org/jboss/logging/annotations/FormatWith.html)

cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
radcortez: But probably we should adjust the message, because the message makes a reference to `@ConfigProperties` but the issue might be on a `@ConfigMapping`.

cdi/src/main/java/io/smallrye/config/inject/InjectionMessages.java
radcortez: +1

cdi/src/test/java/io/smallrye/config/inject/ValidateInjectionTest.java
radcortez: +1

radcortez (changes requested):
This also requires a rebase with the main branch. Thanks!

cdi/src/test/java/io/smallrye/config/inject/ValidateInjectionTest.java
Joseph-Cass: I'm going to give this a try hopefully some time next week. Thanks :)

cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
Joseph-Cass: Have you got a suggested message that would cover both cases?

radcortez: A possibility would be to group the issues by type in the `ConfigValidationException`. I've also remembered that the `Problem` array may report issues from multiple mappings or properties, meaning that you don't know exactly where the problem occurred (just that it was in the mappings area).

In the Extension we could try to register each mapping / property at a time. That would give us a list of problems per type. And in the `ConfigMappingWithPrefix` we can guess if it is a `@ConfigProperties` or `@ConfigMapping`, by looking into the klass. Or just add that metadata into `ConfigMappingWithPrefix` when it is being constructed.

Joseph-Cass:
@radcortez 
> This also requires a rebase with the main branch

I've rebased with main in this commit: https://github.com/smallrye/smallrye-config/pull/513/commits/2b2c04e6640ff6ea5606f80ba38622feb0c1e403

It was a bit of a nasty one, so would you like to check that the new `getValues()` method for the new `IndexedProperties` feature still works as intended? I also added a couple of tests to check that `IndexedProperties` play nicely with the new error messages in this PR.

> But probably we should adjust the message, because the message makes a reference to @ConfigProperties but the issue might be on a @ConfigMapping.

I haven't resolved this yet.

@Azquelt 
> I don't think this is enough information for the user to identify the injection location

I added a helper class to format the injection location in this commit: https://github.com/smallrye/smallrye-config/pull/513/commits/9197cc57f34bc5e66791dd9595ddb104b4e1bae2

In the tests this location still looks a little lengthy, but that's because of the nested inner classes, which we'd be less likely to see in practice.

radcortez:
> @radcortez
> 
> > This also requires a rebase with the main branch
> 
> I've rebased with main in this commit: [2b2c04e](https://github.com/smallrye/smallrye-config/commit/2b2c04e6640ff6ea5606f80ba38622feb0c1e403)
> 
> It was a bit of a nasty one, so would you like to check that the new `getValues()` method for the new `IndexedProperties` feature still works as intended? I also added a couple of tests to check that `IndexedProperties` play nicely with the new error messages in this PR.
> 

I had to rebase again and it should be ok now.

Azquelt:
> I added a helper class to format the injection location in this commit: 9197cc5

I think this now looks fine.

cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
Joseph-Cass: @radcortez do you mean something like:
```
for (ConfigMappings.ConfigMappingWithPrefix cm : configMappingsWithPrefix) {
    try {
        ConfigMappings.registerConfigMappings((SmallRyeConfig) config, new HashSet<>(Arrays.asList(cm)));
    } catch (ConfigValidationException e) {
        if (cm.getKlass().getAnnotation(ConfigMapping.class) != null) {
            adv.addDeploymentProblem(
                    InjectionMessages.msg.retrieveConfigMappingFailure(e.getLocalizedMessage(), e));
        } else if (cm.getKlass().getAnnotation(ConfigProperties.class) != null) {
            adv.addDeploymentProblem(InjectionMessages.msg.retrieveConfigPropertiesFailure(e.getLocalizedMessage(), e));
        } else {
            adv.addDeploymentProblem(e);
        }
    }
}
```
instead of:
https://github.com/smallrye/smallrye-config/blob/c57b5eb12a3360ed137015a99b0adddbcda5eee5/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java#L188-L192

radcortez: Yes.

Joseph-Cass: This seems like we would be unnecessarily creating new Sets for each element. Perhaps we could add a new `registerConfigMappings(Config, ConfigMappingWithPrefix)` method to `ConfigMappings`?

radcortez: Sure

Joseph-Cass:
I had a lot of headaches with validating ConfigMappings individually, but think I've managed to sort it :) 

It's worth noting that I discovered that the `@configMapping` ConfigInjectionException messages generate malformed config property names, which I've noted here:
https://github.com/smallrye/smallrye-config/blob/80190a36c3f4bc6fab855696b72789748b4f8793/cdi/src/test/java/io/smallrye/config/inject/ValidateInjectionTest.java#L210

For example:
```
@ConfigMapping(prefix = "server")
interface Server {
    String missingHost();
    }
```
Would return 
```
io.smallrye.config.ConfigValidationException: Configuration validation failed:
	java.util.NoSuchElementException: SRCFG00014: The config property server.missing-host is required but it could not be found in any config source
```

The trace led me [to here](https://github.com/smallrye/smallrye-config/blob/main/implementation/src/main/java/io/smallrye/config/ConfigMappingGenerator.java#L680-L681) though I think this is an "issue" for another time... I can create one If you think it's worth it?

Finally, I've just noticed that the tests have failed on Windows. I _think_ this may be from the newlines in the following assertions: https://github.com/smallrye/smallrye-config/blob/80190a36c3f4bc6fab855696b72789748b4f8793/cdi/src/test/java/io/smallrye/config/inject/ValidateInjectionTest.java#L274-L282

Annoyingly, the failure has occurred in the one test with a huge Error message (for some reason `org.jboss.weld.exceptions.DeploymentException.DeploymentException()` shows the whole stack trace if there was more than one Exceptions... Not sure if this is desirable behaviour?) so the test output is a little hard to read.

I'll work on fixing that now.

Joseph-Cass:
I've just realised that the following Exception should probably append the localised error message to the end: 

https://github.com/smallrye/smallrye-config/blob/main/implementation/src/main/java/io/smallrye/config/ConfigMessages.java#L136-L137

i.e.
```
    @Message(id = 39, value = "The config property %s with the config value \"%s\" threw an Exception whilst being converted %s")
    IllegalArgumentException converterException(@Cause Throwable converterException, String configProperty, String configValue, String causeMessage);
```

If you agree @radcortez would you like me to do that before/after this PR is concluded?

cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
```diff
@@ -159,36 +162,50 @@ protected void validate(@Observes AfterDeploymentValidation adv) {
...
+                    configMappingWithPrefix((Class<?>) injectionPoint.getType(), prefix), adv, (SmallRyeConfig) config));
         }
 
+    }
+
+    private void registerAndValidateConfigMapping(ConfigMappings.ConfigMappingWithPrefix cm, AfterDeploymentValidation adv,
+            SmallRyeConfig config) {
         try {
-            ConfigMappings.registerConfigMappings((SmallRyeConfig) config, configMappingsWithPrefix);
+            ConfigMappings.registerConfigMappings(config, Collections.singleton(cm));
```
radcortez, line 196: Aren't we creating multiple instances of `ConfigMappingProvider`? I'm not sure if this is ok. For example, when validating properties, each instance will report unused properties because they are only looking into their mapping subset and have no idea if unused properties are being mapped in another instance of `ConfigMappingProvider`.

radcortez (changes requested):

radcortez:
> If you agree @radcortez would you like me to do that before/after this PR is concluded?

You can do it later if you prefer.

I've left some additional comment reviews. Sorry if this PR is taking so much time :(

cdi/src/main/java/io/smallrye/config/inject/ConfigInjectionException.java
```diff
@@ -22,22 +22,36 @@
  * 
  * @author Steve Moyer - The Pennsylvania State University
  */
-public class ConfigException extends Exception {
+public class ConfigInjectionException extends Exception {
```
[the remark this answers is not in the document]
Joseph-Cass: That's understandable. I can rename this if needs be?

Joseph-Cass:
> You can do it later if you prefer

@radcortez Yes, let's do it late. I can create an issue if this PR is merged

cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
Joseph-Cass: Yes, that would be the case. I'm not sure exactly what you mean though? The point of commit https://github.com/smallrye/smallrye-config/pull/513/commits/65d869887be15ef447fd2a1dfc69421d60b22065 was to deal with this fact. IIRC correctly, the `ConfigMappingProvider` just gets used to populate `this.mappings` in `ConfigMappings` via the `registerConfigMappings()` method which was updated:
https://github.com/smallrye/smallrye-config/blob/7a0e7685ae67e3e4317602573ed5f4884ccdc531/implementation/src/main/java/io/smallrye/config/ConfigMappings.java#L26-L35

Also, isn't this what you suggested in this comment? https://github.com/smallrye/smallrye-config/pull/513#discussion_r603264073

Have you got an example test that this implementation may break?

Joseph-Cass: Did you see that `registerAndValidateConfigMapping()` is being called as each `configMapping` and `injectionPoint` is being discovered? As opposed to before when they were all being collected, then registered. I think this may have been the resolve to the "unused properties being mapped" issue you mentioned:  https://github.com/smallrye/smallrye-config/blob/7a0e7685ae67e3e4317602573ed5f4884ccdc531/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java#L180-L189

cdi/src/main/java/io/smallrye/config/inject/ConfigProducerUtil.java
```diff
@@ -48,56 +48,29 @@ private ConfigProducerUtil() {
...
         if (hasCollection(injectionPoint.getType())) {
-            return convertValues(name, injectionPoint.getType(), getDefaultValue(injectionPoint), config);
+            return convertValues(name, injectionPoint.getType(), resolvedValue, defaultValue, config);
         }
 
-        return convertValue(name, resolveConverter(injectionPoint, config), getDefaultValue(injectionPoint), config);
+        return SmallRyeConfig.convertValue(name, resolvedValue, resolveConverter(injectionPoint, config));
     }
 
-    public static <T> T getValue(String name, Type type, String defaultValue, Config config) {
```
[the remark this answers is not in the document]
Joseph-Cass: Okay, I can re-introduce it?

cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
tevans78: @radcortez I think it would be useful to have an example of what you think wouldn't work. At the moment Joe's code is passing all of the existing tests.

radcortez: Ok, let me try to implement a test.

radcortez: Please check: https://github.com/smallrye/smallrye-config/pull/552. I think we should make these tests more self contained.

If I apply this change plus the current registration proposal, we get a validation error, because some properties will be reported as not being used (they are, but the check is performed in different instances). Please check https://github.com/radcortez/smallrye-config/commits/multiple-config-mapping-provider and run `ConfigMappingInjectionTest`.

Joseph-Cass: I agree, #552 is a very good change, and makes the tests a lot more clear. Note that with this change, this Javadoc would need updating:
https://github.com/smallrye/smallrye-config/blob/7a0e7685ae67e3e4317602573ed5f4884ccdc531/cdi/src/test/java/io/smallrye/config/inject/ValidateInjectionTest.java#L45-L47

Isn't the error because you're missing this important change?
https://github.com/smallrye/smallrye-config/blob/7a0e7685ae67e3e4317602573ed5f4884ccdc531/implementation/src/main/java/io/smallrye/config/ConfigMappings.java#L26-L35

Joseph-Cass: I've rebased with main locally, and with the new Exception messages in this PR I'm getting
```
org.jboss.weld.exceptions.DeploymentException: SRCFG02004: Failed to create @ConfigMapping bean with prefix client for interface io.smallrye.config.inject.ConfigMappingInjectionTest$Server. Configuration validation failed:
java.util.NoSuchElementException: SRCFG00014: The config property client.port is required but it could not be found in any config source
```
for this line which sounds about right to me?
https://github.com/smallrye/smallrye-config/blob/d7d2b367db5db2125d1d832ff46b80315782f4d3/cdi/src/test/java/io/smallrye/config/inject/ConfigMappingInjectionTest.java#L102

radcortez: Can you push to the PR, even if it is failing so I can have a better look? Thanks!

Joseph-Cass: I'll push what I had, although I have no idea how this is going to work with the merging of https://github.com/smallrye/smallrye-config/pull/561.

The branch currently fails tests with `does not map to any root` Exception messages. I commented out the properties throwing these Exceptions to achieve the Exception messages in https://github.com/smallrye/smallrye-config/pull/513#discussion_r623205463. 

I'm going to try to create a separate PR which adds the improved Injection Exception messages from this PR without touching the Config Mapping and Config Properties code, so that we at least have some improvement to the messages.

Joseph-Cass: > I'm going to try to create a separate PR

https://github.com/smallrye/smallrye-config/pull/562

@radcortez Am I correct in thinking that if a ConfigMapping interface is injected, with lets say `prefix="client"`, then SmallRye Config "validates" every config property that exists in all Config Sources, for the given prefix, and if there exists a config property which isn't used by any ConfigMapping interface, a `<config property> does not map to any root Exception` is thrown? If so, why? And where can I find documentation on this please?

radcortez: Not every config property, but configs in the namespace of the ConfigMapping. The ideia is that you get a validation in case you forgot to map a property in your mapping class.

Joseph-Cass: Is [this documentation](https://smallrye.io/docs/smallrye-config/mapping/mapping.html) the only doc for ConfigMapping? (and of course [the Javadoc](https://github.com/smallrye/smallrye-config/blob/main/implementation/src/main/java/io/smallrye/config/ConfigMapping.java))

radcortez:
Moved to #562.

PR #514, opened 2021-03-02
Support indexed properties CDI injection.

radcortez:


radcortez:
Implement this in Quarkus as well.

PR #515, opened 2021-03-05
Do not leak java.lang.ProcessEnvironment.

radcortez:


gsmet (approved):
As mentioned by @gastaldi you could use `putAll` to get a better default sizing but other than that, looks good.

But I would add a comment for future you.

gastaldi (dismissed):
image

gastaldi:
You could maybe also use `Collections.unmodifiableMap` on the newly created `HashMap` to comply with the semantics of `java.lang.System#getenv()`

implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
```diff
@@ -110,12 +110,13 @@ private static String replaceNonAlphanumericByUnderscores(String name) {
         return sb.toString();
     }
 
+    /**
+     * A new Map with the contents of System.getEnv. Im the Windows implementation, the Map is an extension of
```
gastaldi, line 114:
```suggestion
     * A new Map with the contents of System.getEnv. In the Windows implementation, the Map is an extension of
```

implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
```diff
@@ -110,12 +110,13 @@ private static String replaceNonAlphanumericByUnderscores(String name) {
...
+     * ProcessEnvironment. This causes issues with Graal and native mode, since ProcessEnvironment should not be
+     * instanciated in the heaop.
```
gastaldi, line 116:
```suggestion
     * instanciated in the heap.
```

PR #523, opened 2021-03-10
Add test to check default ConfigSources.

### LINKED ISSUE #522: Duplicate microprofile-config.properties ConfigSoucres from Config.getConfigSources()
For example:
```
System.out.println("Default ConfigSources: "+config.getConfigSources());
```

Produces (split across new lines for readability):
```
Default ConfigSources: [
SysPropConfigSource, 
EnvConfigSource, 
PropertiesConfigSource[source=file:/<path>/META-INF/microprofile-config.properties], 
PropertiesConfigSource[source=file:/<path>/META-INF/microprofile-config.properties], 
DefaultValuesConfigSource
]
```

These 2 microprofile-config.properties ConfigSources contain the same properties (from `.getProperties()`) but are not equal (from `.equals()`)  

The same thing also happens for `microprofile-config-<profile_name>.properties` for Config Profiles on the [Config Source level](https://download.eclipse.org/microprofile/microprofile-config-2.0/microprofile-config-spec-2.0.html#_on_config_source_level).

radcortez:
- Fixes #522

### DIFF
--- a/testsuite/extra/src/test/java/io/smallrye/config/test/source/SmallRyeConfigTest.java
+++ b/testsuite/extra/src/test/java/io/smallrye/config/test/source/SmallRyeConfigTest.java
@@ -1,0 +1,52 @@
+package io.smallrye.config.test.source;
+
+import static java.util.stream.Collectors.toList;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertTrue;
+
+import java.util.List;
+import java.util.stream.StreamSupport;
+
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.testng.Arquillian;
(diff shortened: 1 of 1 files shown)


Joseph-Cass:
Could you try using `.addAsWebInfResource(` instead of `.addAsResource(` please?

radcortez:
So actually I did some tests and they work in the same way. Both place the file in `/WEB-INF/classes/META-INF/microprofile-config.properties`.

radcortez:
It seems that the duplicated resources might be coming from the dual check in the builder. In theory `WEB-INF/classes/META-INF/microprofile-config.properties` shouldn't be loading anything, since a web classloader places all resources from `/classes` and the valid path it `META-INF/microprofile-config.properties`.

How is the layout of the application you are seeing this behaviour?

Joseph-Cass:
> How is the layout of the application you are seeing this behaviour?

Tom discusses this here: https://github.com/smallrye/smallrye-config/issues/522#issuecomment-800468798

radcortez:
Moved to https://github.com/smallrye/smallrye-config/pull/528.

PR #528, opened 2021-03-17
Don't directly search WEB-INF/classes for config properties files

### LINKED ISSUE #522: Duplicate microprofile-config.properties ConfigSoucres from Config.getConfigSources()

tevans78:
fixes #522

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
+++ b/implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
@@ -144,7 +144,6 @@ public class SmallRyeConfigBuilder implements ConfigBuilder {
         defaultSources.add(new EnvConfigSource());
         defaultSources.add(new SysPropConfigSource());
         defaultSources.addAll(classPathSources(META_INF_MICROPROFILE_CONFIG_PROPERTIES, classLoader));
-        defaultSources.addAll(classPathSources(WEB_INF_MICROPROFILE_CONFIG_PROPERTIES, classLoader));
 
         return defaultSources;
     }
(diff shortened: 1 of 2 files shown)


tevans78:
@radcortez I included your new test. Are there any others you think should be added? I was slightly surprised that I didn't have to change anything else.

radcortez:
Thanks, I think the change should be fine. To be on the safe side, we can add an additional test that adds the config file as `.addAsWebInfResource(mpConfig, "classes/META-INF/microprofile-config.properties");` into the Arquillian archive. And this one should also be loaded.

radcortez:
I think it doesn't make any difference. I've tried it myself and Arquillian will always create the file in the same place.

radcortez:
@kabir do you see any issue from the Wildfly side? I've tried this with the config quickstart and it seems to work. There may be some other uses cases that I'm not aware.

kabir:
@radcortez I'll  run some parts of our testsuite locally to see what happens (assuming/checking that SM Config master works with WildFly first before testing this change)

radcortez:
Thanks!

kabir:
Thank you for thinking of us :)

I've run everything I can think of and it all seems to work fine with this fix. If I missed something and end up needing this again, I guess I can provide my own SmallRyeConfigProviderResolver or something and add this back into a builder. But I don't think that is very likely.

radcortez:
Thanks, I'll merge this. If if you find any issue let me know and we figure something else.

radcortez (approved):

PR #530, opened 2021-03-18
Fix validation to verify that a property name is present in a ConfigMapping root.

radcortez:


implementation/src/main/java/io/smallrye/config/ConfigMappingProvider.java
```diff
@@ -649,14 +649,40 @@ private void mapConfiguration(SmallRyeConfig config, ConfigMappings mappings) th
         mappings.registerConfigMappings(context.getRootsMap());
     }
 
-    private boolean isPropertyInRoot(String propertyName) {
+    private boolean isPropertyInRoot(NameIterator propertyName) {
```
someth2say, line 652: In the previous version, `isPropertyInRoot` returns true if *any* registered root contains the property.
But in this version, it returns true only when *all* registered roots contain the property.

Is this behavior change intentional?

radcortez: Probably not :)

radcortez: I guess I've messed up on the branching and pushed and old version when I've renamed `master` to `main`. I took the chance and added a few more tests. Do you mind reviewing it? :)

someth2say (approved):
LGTM
I'm just a bit puzzled about the `serverBoot` entries in the tests.
Do they serve any purpose? 
They seem to not interfere with the mappings, but they are not asserted.

Anyway, its presence/absence does no difference, so no issues leaving them.

radcortez:
Thanks for the review.

These properties are there to make sure they don't interfere with the mapping validation. When you reported the other issue in Quarkus, I've found that that if you have a ConfigMapping with a prefix (for instance `server`) and a property that started with the prefix but without sub properties (for instance `serverSomething`), the validation was reporting that `serverSomething` was not mapped, since the validation was doing a `startsWith`.

PR #540, opened 2021-04-01
Add a FileSystemConfigSourceFactory to bootstrap the FileSystemConfigSource with the configuration property smallrye.config.source.file.location.

radcortez:


henrikcaesar:
A question about the smallrye.config.source.file.location property.
Wouldn't it be good if it could be a list of locations? (smallrye.config.source.file.location**s**)
Just as the smallrye.config.locations property.

I need to read from two sources, one configmap and one secret, and this is not supported as far as I understand. I would still need to create my own ConfigSourceProvider returning two FileSystemConfigSource:es.

Perhaps a separate issue just wanted to see what you think?

radcortez:
Hi @henrikcaesar,

Thank for reaching out. Yes, I think it makes sense. Do you mind creating a separate issue for it? Feel free to implement it as well if you are up for it :)

PR #549, opened 2021-04-14
Default smallrye.config.mapping.validate-unknown to false

### LINKED ISSUE #548: smallrye.config.mapping.validate-unknown should default to false
I pulled version 2.2.0 into OpenLiberty and found that unless `smallrye.config.mapping.validate-unknown` was set to `false`, SmallRye Config 2.2.0 did not pass the TCK. Using the default value of `true`, hundreds of validation errors were produced. e.g.
```
Stack Dump = org.jboss.weld.exceptions.DeploymentException: Configuration validation failed:
	PATH does not map to any root
	java.specification.version does not map to any root
	com.ibm.team.repository.transport.client.protocol does not map to any root
	bvt.prop.security_4_HTTP_default does not map to any root
	bvt.prop.controller_3.http does not map to any root
	wlp.workarea.dir does not map to any root
	com.ibm.vm.bitmode does not map to any root
	sun.arch.data.model does not map to any root
	java.vendor.url does not map to any root
	ldap.3.ssl.port does not map to any root
	mp.tck.prop.dummy does not map to any root
	bvt.prop.batch.endpoint_1_HTTP_default does not map to any root
	sun.boot.library.path does not map to any root
	sun.java.command does not map to any root
	bvt.prop.controller_4.replica does not map to any root
	com.ibm.tools.attach.enable does not map to any root
	java.specification.vendor does not map to any root
	java.version.date does not map to any root
	java.home does not map to any root
	org.apache.aries.blueprint.preemptiveShutdown does not map to any root
```
The full list was much longer. Given that the vast majority of these errors are being triggered by properties from the default config sources, having `validate-unknown` default to `true` does not seem like a good idea.

I also note that the error message is not very helpful and I had to dig through the source to find the undocumented property which was causing the problem.

tevans78:
Signed-off-by: tevans <tevans@uk.ibm.com>

fixes #548

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/ConfigMappingProvider.java
+++ b/implementation/src/main/java/io/smallrye/config/ConfigMappingProvider.java
@@ -769,7 +769,7 @@ final class ConfigMappingProvider implements Serializable {
     public static final class Builder {
         final Map<String, List<Class<?>>> roots = new HashMap<>();
         final List<String[]> ignored = new ArrayList<>();
-        boolean validateUnknown = true;
+        boolean validateUnknown = ConfigMappings.VALIDATE_UNKNOWN_DEFAULT;
 
         Builder() {
         }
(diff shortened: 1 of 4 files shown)


radcortez:
@tevans78 Thanks for the PR. 

If you don't mind, I'm replacing it with #561, which maintains the expected behaviour for SR `@ConfigMapping`, but remove the validation entirely for `@ConfigProperties`

PR #579, opened 2021-05-21
Allow to fetch multiple config parameters as a Map

### LINKED ISSUE #578: Add basic support of Map injection
As MicroProfile Config doesn't yet support map injection, this ticket is meant to propose a basic support of `Map` injection to propose an early implementation of it without taking the risk to diverge too much with the specification.

essobedo:
fixes #578 

## Motivation

When we need a dynamic configuration, it can be very helpful to use a map. For example, let's say that I want to keep in my configuration the custom reason phrases to return to the enduser according to the http status code.

I would like to have:

The expected code

```
@ConfigProperty(name = "reasons")
Map<Integer, String> reasonPhrases; // or eventually Map<String, String>
```

The expected configuration
```
reasons.200=My custom reason phrase for OK
reasons.201=My custom reason phrase for Created
...
```

## Modifications

* Adds a generic producer for Map
* Adds a MapConverter to convert the content of type `<key1>=<value1>;<key2>=<value2>...` into a `Map`.
* Adds map support to the `ConfigMappingGenerator` 
* Adds related unit tests
* Adds new methods `getValues(String name, Class<K> kClass, Class<V> vClass)` and `getOptionalValues(String name, Class<K> kClass, Class<V> vClass)` to support properly the map from the class `SmallRyeConfig`

### DIFF
--- a/cdi/src/main/java/io/smallrye/config/inject/ConfigProducerUtil.java
+++ b/cdi/src/main/java/io/smallrye/config/inject/ConfigProducerUtil.java
@@ -1,6 +1,7 @@
 package io.smallrye.config.inject;
 
 import static io.smallrye.config.Converters.newCollectionConverter;
+import static io.smallrye.config.Converters.newMapConverter;
 import static io.smallrye.config.Converters.newOptionalConverter;
 
 import java.lang.annotation.Annotation;
@@ -12,6 +13,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiFunction;
@@ -27,8 +29,10 @@ import org.eclipse.microprofile.config.ConfigValue;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.eclipse.microprofile.config.spi.Converter;
 
+import io.smallrye.config.Converters;
 import io.smallrye.config.SecretKeys;
 import io.smallrye.config.SmallRyeConfig;
+import io.smallrye.config.common.AbstractConverter;
 import io.smallrye.config.common.AbstractDelegatingConverter;
 
 /**
@@ -51,19 +55,7 @@ public final class ConfigProducerUtil {
      * @return the converted configuration value.
      */
     public static <T> T getValue(InjectionPoint injectionPoint, Config config) {
-        String name = getName(injectionPoint);
-        if (name == null) {
-            return null;
-        }
-
-        String defaultValue = getDefaultValue(injectionPoint);
-        String rawValue = getRawValue(name, config);
-        if (hasCollection(injectionPoint.getType())) {
-            return convertValues(name, injectionPoint.getType(), rawValue, defaultValue, config);
-        }
-
-        return ((SmallRyeConfig) config).convertValue(name, resolveDefault(rawValue, defaultValue),
-                resolveConverter(injectionPoint, config));
+        return getValue(getName(injectionPoint), injectionPoint.getType(), getDefaultValue(injectionPoint), config);
     }
 
     /**
@@ -71,7 +63,7 @@ public final class ConfigProducerUtil {
      *
      * @param name the name of the configuration property.
      * @param type the {@link Type} of the configuration value to convert.
-     * @param defaultValue the defaut value to use if no configuration value is found.
+     * @param defaultValue the default value to use if no configuration value is found.
      * @param config the current {@link Config} instance.
      *
      * @return the converted configuration value.
@@ -80,8 +72,30 @@ public final class ConfigProducerUtil {
         if (name == null) {
             return null;
         }
-        String resolvedValue = resolveValue(name, defaultValue, config);
-        return ((SmallRyeConfig) config).convertValue(name, resolvedValue, resolveConverter(type, config));
+        if (hasCollection(type)) {
+            return convertValues(name, type, getRawValue(name, config), defaultValue, config);
+        } else if (hasMap(type)) {
+            return convertValues(name, type, defaultValue, config);
+        }
+
+        return ((SmallRyeConfig) config).convertValue(name, resolveDefault(getRawValue(name, config), defaultValue),
+                resolveConverter(type, config));
+    }
+
+    /**
+     * Converts the direct sub properties of the given parent property as a Map.
+     *
+     * @param name the name of the parent property for which we want the direct sub properties as a Map.
+     * @param type the {@link Type} of the configuration value to convert.
+     * @param defaultValue the default value to convert in case no sub properties could be found.
+     * @param config the configuration from which the values are retrieved.
+     * @param <T> the expected type of the configuration value to convert.
+     *
+     * @return the converted configuration value.
+     */
+    private static <T> T convertValues(String name, Type type, String defaultValue, Config config) {
+        return ((SmallRyeConfig) config).convertValue(name, null,
+                resolveConverter(type, config, (kC, vC) -> new StaticMapConverter<>(name, defaultValue, config, kC, vC)));
     }
 
     private static <T> T convertValues(String name, Type type, String rawValue, String defaultValue, Config config) {
@@ -98,7 +112,7 @@ public final class ConfigProducerUtil {
             for (String indexedProperty : indexedProperties) {
                 // Never null by the rules of converValue
                 collection.add(
-                        ((SmallRyeConfig) config).convertValue(indexedProperty, resolveValue(indexedProperty, null, config),
+                        ((SmallRyeConfig) config).convertValue(indexedProperty, getRawValue(indexedProperty, config),
                                 itemConverter));
             }
             return collection;
@@ -127,21 +141,17 @@ public final class ConfigProducerUtil {
         return SecretKeys.doUnlocked(() -> config.getConfigValue(name).getValue());
     }
 
-    private static String resolveValue(String name, String defaultValue, Config config) {
-        String rawValue = getRawValue(name, config);
-        return resolveDefault(rawValue, defaultValue);
-    }
-
     private static String resolveDefault(String rawValue, String defaultValue) {
         return rawValue != null ? rawValue : defaultValue;
     }
 
-    private static <T> Converter<T> resolveConverter(final InjectionPoint injectionPoint, final Config config) {
-        return resolveConverter(injectionPoint.getType(), config);
+    private static <T> Converter<T> resolveConverter(final Type type, final Config config) {
+        return resolveConverter(type, config, Converters::newMapConverter);
     }
 
     @SuppressWarnings("unchecked")
-    private static <T> Converter<T> resolveConverter(final Type type, final Config config) {
+    private static <T> Converter<T> resolveConverter(final Type type, final Config config,
+            final BiFunction<Converter<Object>, Converter<Object>, Converter<Map<Object, Object>>> mapConverterFactory) {
         Class<T> rawType = rawTypeOf(type);
         if (type instanceof ParameterizedType) {
             ParameterizedType paramType = (ParameterizedType) type;
@@ -150,11 +160,18 @@ public final class ConfigProducerUtil {
                 return (Converter<T>) newCollectionConverter(resolveConverter(typeArgs[0], config), ArrayList::new);
             } else if (rawType == Set.class) {
                 return (Converter<T>) newCollectionConverter(resolveConverter(typeArgs[0], config), HashSet::new);
+            } else if (rawType == Map.class) {
+                return (Converter<T>) mapConverterFactory.apply(resolveConverter(typeArgs[0], config),
+                        resolveConverter(typeArgs[1], config));
             } else if (rawType == Optional.class) {
-                return (Converter<T>) newOptionalConverter(resolveConverter(typeArgs[0], config));
+                return (Converter<T>) newOptionalConverter(resolveConverter(typeArgs[0], config, mapConverterFactory));
             } else if (rawType == Supplier.class) {
-                return resolveConverter(typeArgs[0], config);
+                return resolveConverter(typeArgs[0], config, mapConverterFactory);
             }
+        } else if (rawType == Map.class) {
+            // No parameterized types have been provided so it assumes that a Map of String is expected
+            return (Converter<T>) mapConverterFactory.apply(resolveConverter(String.class, config),
+                    resolveConverter(String.class, config));
         }
         // just try the raw type
         return config.getConverter(rawType).orElseThrow(() -> InjectionMessages.msg.noRegisteredConverter(rawType));
@@ -206,6 +223,23 @@ public final class ConfigProducerUtil {
         } else {
             throw InjectionMessages.msg.noRawType(type);
         }
+    }
+
+    /**
+     * Indicates whether the given type is a type of Map or is a Supplier or Optional of Map.
+     * 
+     * @param type the type to check
+     * @return {@code true} if the given type is a type of Map or is a Supplier or Optional of Map,
+     *         {@code false} otherwise.
+     */
+    private static boolean hasMap(final Type type) {
+        Class<?> rawType = rawTypeOf(type);
+        if (rawType == Map.class) {
+            return true;
+        } else if (type instanceof ParameterizedType) {
+            return hasMap(((ParameterizedType) type).getActualTypeArguments()[0]);
+        }
+        return false;
     }
 
     private static <T> boolean hasCollection(final Type type) {
@@ -278,7 +312,7 @@ public final class ConfigProducerUtil {
         throw InjectionMessages.msg.noConfigPropertyDefaultName(ip);
     }
 
-    final static class IndexedCollectionConverter<T, C extends Collection<T>> extends AbstractDelegatingConverter<T, C> {
+    static final class IndexedCollectionConverter<T, C extends Collection<T>> extends AbstractDelegatingConverter<T, C> {
         private static final long serialVersionUID = 5186940408317652618L;
 
         private final IntFunction<Collection<T>> collectionFactory;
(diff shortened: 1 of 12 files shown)


essobedo:
Hi @radcortez it is the PR related https://github.com/eclipse/microprofile-config/issues/562#issuecomment-841279200
It is very basic because there are many things unspecified (like for example how to manage default values in case of `@ConfigProperty`) but at least it covers simple uses cases. WDYT of this PR?

cdi/src/main/java/io/smallrye/config/inject/ConfigProducerUtil.java
```diff
@@ -107,6 +111,36 @@ private ConfigProducerUtil() {
...
+     *         type is a type of map, an empty map otherwise.
+     */
+    private static <T> T convertValuesToMap(String name, Type type, Config config) {
+        Map map = new HashMap();
+        final String prefix = name.endsWith(".") ? name : name + ".";
+        if (type instanceof ParameterizedType) {
+            ParameterizedType paramType = (ParameterizedType) type;
+            Type[] typeArgs = paramType.getActualTypeArguments();
+            for (String pName : config.getPropertyNames()) {
+                if (pName.startsWith(prefix)) {
```
radcortez, line 132: What if you have multiple subnamespaces? For instance:

```
my.map.key=key
my.map.value=value
my.map.sub.something=1234
```

cdi/src/test/java/io/smallrye/config/inject/ConfigInjectionTest.java
```diff
@@ -80,6 +83,9 @@ void converters() {
 
     @ApplicationScoped
     static class ConfigBean {
+        @Inject
+        @ConfigProperty(name = "reasons")
+        Map<Integer, String> reasons;
```
radcortez, line 88: I think we must add some more tests. Add more Map combinations, use a custom converter, etc.

radcortez (changes requested):
Thanks for the PR!

> how to manage default values in case of `@ConfigProperty`, 
We could support something like `key=value` and parse the default value if the injection is a map. Alternatively, we could just disallow it all together and throw a validation exception.

> how to deal with `config.getValue` in case of map since we have no way to get the Parameterized types) 
We would need to support something similar to a `GenericType` like JAX-RS to carry type information, but for now we could just assume that the programmatic API for Map returns a `Map<String,String>`.

cdi/src/main/java/io/smallrye/config/inject/ConfigProducerUtil.java
essobedo: @radcortez You mean that in that case you would expect only the keys `key` and `value` in the map with `@ConfigProperty(name = "my.map")`?

radcortez: I guess this was not a good example :) 

My point was about keys after the map that have more that one level? Should the key be included as `x.y`? Should it be ignored because it is composed?

essobedo: I understand your point, for me since there many unspecified  aspects like this one, I wanted to keep it simple by making a "flat map". 
So in your example  If I have the following code:

```
@ConfigProperty(name = "my.map")
Map<String, String> map;
```

I expect to have key/key, value/value and sub.something/1234 in my map as pairs of String. 

Subnamespaces could make sense if we wanted to support complex mapping like:

```
my.map.c1.name=John
my.map.c1.age=33
```

The target class:
```
class Customer {
   String name;
   int age;
}
```

The mapping definition

```
@ConfigProperty(name = "my.map")
Map<String, Customer> map;
```

I could then expect to have one entry inside my map whose key is `c1` but I think that we would be too far from the current logic of the original spec where the value of all fields annotated with `@ConfigProperty` should be the value of the property converted into the target type.

Don't you agree?

cdi/src/test/java/io/smallrye/config/inject/ConfigInjectionTest.java
essobedo: @radcortez I added some tests, let me know if it is not enough. Please note that it brings support of Map of List or Map of Set but only with `@ConfigProperty` but not in case of `@ConfigProperties`, if you expect the same support with `@ConfigProperties`, I think that it should be done in another ticket.

cdi/src/main/java/io/smallrye/config/inject/ConfigProducerUtil.java
radcortez: I don't think we should support complex mapping (well we could, since this is already supported via @ConfigMapping, but I believe it would require a lot of extra work)

What I'm trying to define here is what should be the behaviour in a case we find a subnamespace when injecting a map. I would say that we should ignore these properties.

essobedo: @radcortez Fixed

cdi/src/test/java/io/smallrye/config/inject/ConfigInjectionTest.java
essobedo: @radcortez done

essobedo:
> how to manage default values in case of `@ConfigProperty`,
> We could support something like `key=value` and parse the default value if the injection is a map. Alternatively, we could just disallow it all together and throw a validation exception.

I just added the support of default value with the next format `<key1>=<value1>;<key2>=<value2>...`, knowing that we can use a backslash to escape the characters `=` and `;` if needed

> how to deal with `config.getValue` in case of map since we have no way to get the Parameterized types)
> We would need to support something similar to a `GenericType` like JAX-RS to carry type information, but for now we could just assume that the programmatic API for Map returns a `Map<String,String>`.

I added a new BuiltIn Converter for the Map to be able to support `config.getValue`

essobedo:
@radcortez Hi, all your remarks have been addressed, this PR is ready for review now. Thank you in advance 🙏

essobedo:
> I added a new BuiltIn Converter for the Map to be able to support `config.getValue`

Finally I rather added new dedicated methods as extensions for the map support such that I don't need the Map converter to be part of the Builtin converters

radcortez:
> I expected to have it included into Quarkus 2.0, I guess it is too late now 😢

I wouldn't be too worried about it. Quarkus has a very high release cadence, so it we miss the boat now, we will get the next one soon :)

Anyway, we still need to test this in Quarkus. Did you try it? In the meantime I'll have review the PR.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -191,13 +191,44 @@
...
+     * @param kClass the type into which the keys should be converted
+     * @param vClass the type into which the values should be converted
+     * @param <K> the key type
+     * @param <V> the value type
+     * @return the resolved property value as an instance of the requested Map (not {@code null})
+     * @throws IllegalArgumentException if a key or a value cannot be converted to the specified types
+     * @throws NoSuchElementException if no direct sub properties could be found.
+     */
+    @Experimental("Extension to retrieve mandatory sub properties as a Map")
+    public <K, V> Map<K, V> getMapValue(String name, Class<K> kClass, Class<V> vClass) {
```
radcortez, line 207: Wondering if we should just name this `getValue`. It shouldn't clash with any of the other methods and it will be consistent with `getValue` and `getValues`. What do you think?

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -294,11 +325,68 @@ public String getRawValue(String name) {
...
+     *
+     * @param name The configuration property name
+     * @param kClass the type into which the keys should be converted
+     * @param vClass the type into which the values should be converted
+     * @param <K> the key type
+     * @param <V> the value type
+     * @return the resolved property value as an instance of the requested Map (not {@code null})
+     * @throws IllegalArgumentException if a key or a value cannot be converted to the specified types
+     */
+    @Experimental("Extension to retrieve non mandatory sub properties as a Map")
```
radcortez, line 384: Same here with the naming.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -294,11 +325,68 @@ public String getRawValue(String name) {
...
+     * it will keep all the existing properties whose key starts with the given name. All the
+     * remaining properties will be converted using the format {@code <key1>=<value1>;<key2>=<value2>...}.
+     * <p>
+     * The special characters are automatically escaped with a backslash.
+     *
+     * @param name the name used as a prefix to collect all the properties that match.
+     * @return a String containing the matching properties that have been formatted if any, {@code null}
+     *         otherwise.
+     */
+    public String getMapRawValue(String name) {
```
radcortez, line 351: Wondering if there is any point in flattening the map values to a single String, just so we can use the Converter. Maybe we could just create and populate the map directly?

radcortez (commented):
I think this is going into the right direction. Left a few comments for us to discuss :) Thanks!

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
essobedo: If so I would rather use `getValues` since in a `Map` you have more than one value, don't you agree?

cdi/src/main/java/io/smallrye/config/inject/ConfigProducerUtil.java
```diff
@@ -208,6 +222,28 @@ private static String resolveDefault(String rawValue, String defaultValue) {
...
     }
 
+    /**
+     * Indicates whether the given type is a type of Map or is a Supplier or Optional of Map.
+     * 
+     * @param type the type to check
+     * @return {@code true} if the given type is a type of Map or is a Supplier or Optional of Map,
+     *         {@code false} otherwise.
+     */
+    private static boolean isMapExpected(final Type type) {
```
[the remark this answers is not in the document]
essobedo: You mean `hasCollection` right?

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
essobedo: I know but the api is String based, I mean the converters expect String, the ConfigValue expect String, the default value is expected as a String... If really needed I can check how it can be improved in a different commit

radcortez: It can be done like the indexed property support. Lets leave it for now and come back to it later if it causes any issues.

radcortez:
@essobedo Thank you! I think we are almost done. Just one last detail: can we document the new feature here: https://github.com/smallrye/smallrye-config/blob/fecdcf541a0ad979cc3f83f8e5f4829da61e5be8/doc/modules/ROOT/pages/config/config.adoc. You can follow the indexed properties example.

radcortez:
Also, I recommend to make a build and test it with Quarkus master. Did you try it with Quarkus as is?

essobedo:
> Also, I recommend to make a build and test it with Quarkus master. Did you try it with Quarkus as is?

Not yet but I will test it today and let you know the results.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -191,13 +191,75 @@
...
+     *
+     * @param name The configuration property name
+     * @param keyConverter The converter to use for the keys.
+     * @param valueConverter The converter to use for the values.
+     * @param <K> The type of the keys.
+     * @param <V> The type of the values.
+     * @return the resolved property value as an instance of the requested Map or {@code null} if it could not be found.
+     * @throws IllegalArgumentException if a key or a value cannot be converted to the specified types
+     */
+    public <K, V> Map<K, V> valuesAsMap(String name, Converter<K> keyConverter, Converter<V> valueConverter) {
```
essobedo, line 226: I could not use getValues as name here because it is in conflict with `getValues(String name, Converter<T> converter, IntFunction<C> collectionFactory)`

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
essobedo: @radcortez Done in a different commit to revert it easily if needed. Please tell me if it is more or less what you had in mind, meanwhile I will test it with Quarkus then I will add some doc about it

essobedo:
@radcortez just to let you know that I had to make couple of additional fixes to make it work with Quarkus, knowing that even on Quarkus side there are some little changes to make. I have not yet finished but I could make injection with `@ConfigProperty` work but I have one last bug to fix on Quarkus side with `@ConfigProperties`. Indeed I get the following error at runtime only, it should not be too hard to fix:

```
2021-05-28 20:47:41,177 ERROR [io.qua.run.Application] (main) Failed to start application (with profile worker): java.lang.ClassNotFoundException: org.objectweb.asm.ClassVisitor
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
        at io.quarkus.bootstrap.runner.RunnerClassLoader.loadClass(RunnerClassLoader.java:104)
        at io.quarkus.bootstrap.runner.RunnerClassLoader.loadClass(RunnerClassLoader.java:60)
        at io.smallrye.config.ConfigMappingClass.<init>(ConfigMappingClass.java:35)
        at io.smallrye.config.ConfigMappingClass.createConfigurationClass(ConfigMappingClass.java:27)
        at io.smallrye.config.ConfigMappingClass.access$000(ConfigMappingClass.java:7)
        at io.smallrye.config.ConfigMappingClass$1.computeValue(ConfigMappingClass.java:11)
        at io.smallrye.config.ConfigMappingClass$1.computeValue(ConfigMappingClass.java:8)
        at java.base/java.lang.ClassValue.getFromHashMap(ClassValue.java:228)
        at java.base/java.lang.ClassValue.getFromBackup(ClassValue.java:210)
        at java.base/java.lang.ClassValue.get(ClassValue.java:116)
        at io.smallrye.config.ConfigMappingClass.getConfigurationClass(ConfigMappingClass.java:17)
        at io.smallrye.config.ConfigMappingLoader.getConfigMappingClass(ConfigMappingLoader.java:42)
        at io.smallrye.config.ConfigMappingProvider$Builder.addRoot(ConfigMappingProvider.java:887)
        at io.smallrye.config.ConfigMappings.mapConfiguration(ConfigMappings.java:52)
        at io.smallrye.config.ConfigMappings.registerConfigProperties(ConfigMappings.java:43)
        at io.quarkus.arc.runtime.ConfigRecorder.registerConfigProperties(ConfigRecorder.java:60)
        at io.quarkus.deployment.steps.ConfigBuildStep$registerConfigClasses-2116437784.deploy_0(ConfigBuildStep$registerConfigClasses-2116437784.zig:159)
        at io.quarkus.deployment.steps.ConfigBuildStep$registerConfigClasses-2116437784.deploy(ConfigBuildStep$registerConfigClasses-2116437784.zig:40)
        at io.quarkus.runner.ApplicationImpl.doStart(ApplicationImpl.zig:546)
        at io.quarkus.runtime.Application.start(Application.java:101)
        at io.quarkus.runtime.ApplicationLifecycleManager.run(ApplicationLifecycleManager.java:101)
```

essobedo:
@radcortez After a deeper investigation, it appears that I need to move the method `generateInterfaceName` out of `ConfigMappingGenerator` to avoid loading `ClassVisitor` at runtime for nothing since the method doesn't use it and the byte code has been generated at build time already

implementation/src/main/java/io/smallrye/config/ConfigMappingGenerator.java
```diff
@@ -352,20 +352,6 @@
         return writer.toByteArray();
     }
 
-    static String generateInterfaceName(final Class<?> classType) {
```
essobedo, line 355: Need to move it to prevent getting `ClassNotFoundException` at runtime because `asm` is missing which is wanted as not used and not expected in native mode

essobedo:
So I tested it successfully on Quarkus (modulo a small patch to support map) with the next use case:

The main configuration class:
```
@Singleton
public class GreetingConfiguration {

    @ConfigProperty(name = "root.numbers")
    public Map<Integer, String> numbers;

    @ConfigProperty(name = "root.numbers")
    public Optional<Map<Integer, String>> oNumbers;

    @ConfigProperty(name = "root.numbers")
    public Supplier<Map<Integer, String>> sNumbers;

    @ConfigProperties
    public RootConfig root;
}
```

The class `RootConfig`:
```
@ConfigProperties(prefix = "root")
public class RootConfig {

    public Map<Integer, String> numbers;
}
```

Let me know if it is good enough for you or you have another uses in mind to test, meanwhile I start working on the doc.

essobedo:
FYI here is my test branch for Quarkus https://github.com/quarkusio/quarkus/compare/main...essobedo:add-map-support?expand=1

essobedo:
@radcortez I added the doc, so now all your remarks have been addressed

cdi/src/main/java/io/smallrye/config/inject/ConfigProducerUtil.java
```diff
@@ -128,20 +142,20 @@ static String getRawValue(String name, Config config) {
     }
 
     private static String resolveValue(String name, String defaultValue, Config config) {
```
radcortez, line 144: This seems to be used only by the indexed case and called always with null as a default. We should inline any code and remove the method.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
radcortez: Ok, let's just use `getValuesAsMap`. So we can keep the `get` consistency.

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -191,13 +191,75 @@
...
+     * @param configValue The value of the configuration property
+     * @param converter The converter to use to convert the value into the expected type.
+     * @param <T> the expected type.
+     * @return The value of the property converted into the expected type
+     * @throws IllegalArgumentException if the property cannot be converted to the specified type
+     * @throws NoSuchElementException if the property is not defined or is defined as an
+     *         empty string or the converter returns {@code null}
+     */
+    @SuppressWarnings("unchecked")
+    private <T> T getValue(String name, ConfigValue configValue, Converter<T> converter) {
```
radcortez, line 262: I think we don't need to split the method anymore, since it only has one usage now.

implementation/src/main/java/io/smallrye/config/ConfigMappingGenerator.java
radcortez: Interesting. I wonder why we didn't pick this up before.

radcortez (changes requested):
@essobedo Thank you for your patience. Very good job. Just a few more minor things (leftovers from the other proposals and we should be good to go).

essobedo:
@radcortez Done, please check again. Do you want me to squash my commits?

radcortez:
> @radcortez Done, please check again. Do you want me to squash my commits?

Yes, please. Thank you :)

essobedo:
> > @radcortez Done, please check again. Do you want me to squash my commits?
> 
> Yes, please. Thank you :)

Done

radcortez (approved):

radcortez:
@essobedo Thank you very much for all the work. I hope you can continue to contribute :)

Can you please open a Draft PR for Quarkus with the changes required? It it fine to use the SNAPSHOT (the CI won't build it as draft). We can update the version once we release, so the PR is for us to not forget the changes.

I think we may need some other fixes for Quarkus in other areas, so we need these in before doing the final release.

essobedo:
> @essobedo Thank you very much for all the work. I hope you can continue to contribute :)

Your welcome, it is always with great pleasure that I contribute to open source projects. If you would like me to help you (if I can) on another ticket, please let me know 😄 

> Can you please open a Draft PR for Quarkus with the changes required? It it fine to use the SNAPSHOT (the CI won't build it as draft). We can update the version once we release, so the PR is for us to not forget the changes.
> 
> I think we may need some other fixes for Quarkus in other areas, so we need these in before doing the final release.

Here it is https://github.com/quarkusio/quarkus/pull/17579

rmanibus:
Thanks a lot @essobedo !

PR #581, opened 2021-05-23
Application fails to start in native image mode when some config sour…

mswiderski:
…ces are not available due to dead code elimination

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
+++ b/implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
@@ -28,6 +28,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.OptionalInt;
+import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


radcortez:
@mswiderski can you please provide me with more details? How are you using SR Config? Do you have a reproducer I can look at?

mswiderski:
@radcortez will try to prepare simple reproducer tomorrow morning (on the road today) but there is no specifics just following extensions in my project

````
   <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-funqy-knative-events</artifactId>
    </dependency>
    <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-rest-client</artifactId>
    </dependency>
    <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-rest-client-jackson</artifactId>
    </dependency>
````

and that stacktrace I pasted in the first comment is the only thing that is shown upon start, build of the native image is going through without issues.

radcortez:
Ok, thanks. I can try to do that on my own.

Probably we would need to do the fix in Quarkus.

mswiderski:
> Ok, thanks. I can try to do that on my own.

let me know if you manage to do that by tomorrow otherwise I will work on it first thing in the morning.

> Probably we would need to do the fix in Quarkus.

Do you see something really Quarkus specific here? If I would use small rye config with anything else that is compiled to native image I believe it would result in the same thing. Or am I missing something?

The problem as I understood it was that the class is removed by native image compiler but the service locators files are still declaring them to be loaded and thus the failure.

radcortez:
> > Ok, thanks. I can try to do that on my own.
> 
> let me know if you manage to do that by tomorrow otherwise I will work on it first thing in the morning.
> 

I've tried with code Quarkus and just the sample application with the listed dependencies and it worked fine for me. Maybe some piece of code is triggering the issue.

> > Probably we would need to do the fix in Quarkus.
> 
> Do you see something really Quarkus specific here? If I would use small rye config with anything else that is compiled to native image I believe it would result in the same thing. Or am I missing something?
> 
> The problem as I understood it was that the class is removed by native image compiler but the service locators files are still declaring them to be loaded and thus the failure.

Yes, but this will just hide the error for other situations not related with Quarkus and native images. For instance a change in the classloader may prevent service loader instances to load properly.

mswiderski:
@radcortez it turned out to be only reproducable when there is custom `ConfigSource` in the app as it will then enable service loader files for native image and that's why it fails at boot.

Here is a reproducer https://github.com/mswiderski/funqy-rest-client

just run `man clean package -Pcontainer-native` and it will build the image in container and produce a container with that native image. So then starting it with docker run will fail with the mentioned error.

> Yes, but this will just hide the error for other situations not related with Quarkus and native images. For instance a change in the classloader may prevent service loader instances to load properly.

Agreed that I might have side effects so including log statement about failing service loader instances could be good enough hint. What do you think?
I also opened a thread on quarkus mailing list so if you could chime in and provide your feedback there it would be appreciated.

radcortez:
Thanks! Let me have a look into the reproducer.

radcortez:
@mswiderski if you don't mind, I'm closing this PR, since this needs to be fixed from the Quarkus side. This is a proposal to start with the fix: https://github.com/quarkusio/quarkus/pull/17483.

Thank you for looking into this and trying to fix the issue.

mswiderski:
@radcortez sounds good to me. thanks for working on the fix and let me know if you want me to test it out

PR #584, opened 2021-06-01
Improve way to manage default values in case of a Map

essobedo:
Follow up of #579

## Motivation

The current documentation about the support of Map doesn't describe how to define default values, it needs to be added. Moreover #579 adds a sonar violation , it needs to be fixed. Finally tests on failures are missing.

## Modifications:

* Rewrites the converter to avoid relying on a regular expression with a negative lookbehind to fix the Sonar violation
* Adds a new config message for the MapConverter in case of invalid format
* Adds unit tests to check that errors are properly thrown in case sub properties are missing and default value is not in the expected format.
* Modifies the example for `@ConfigProperty` to set `defaultValue` and describe how to use it properly


essobedo:
@radcortez Hi, here is the missing part of the previous task

radcortez:
Thanks!

radcortez (approved):

PR #588, opened 2021-06-03
Avoid throwing Exception for Optional types and ConfigValue if expression cannot be expanded

### LINKED ISSUE #587: Avoid throwing Exception for Optional types and ConfigValue if expression cannot be expanded
Up to now, when the value of a property could not be expanded, a `NoSuchElementException` is thrown whatever the target type while it should not be thrown for an Optional type as stated into the specification, I quote:

`
If an expression cannot be expanded and does not have a default value, a NoSuchElementException is thrown. In the Optional case, an empty Optional will be returned.
`

Source: https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/property-expressions.asciidoc

essobedo:
fixes #587

## Motivation

Up to now, when the value of a property could not be expanded, a `NoSuchElementException` is thrown whatever the target type while it should not be thrown for an Optional type or a ConfigValue.

## Modifications:

* Adds the new method `getRawConfigValue` to provide a `ConfigValue` without catching the `NoSuchElementException`
* Modifies the method `getConfigValue` to return an empty `ConfigValue` when a `NoSuchElementException` is caught (not stated into the spec but IMHO it makes more sense than throwing the exception) 
* Adds the new method `noSuchElementExceptionForbidden` to know if a target type accepts or not `NoSuchElementException` if the value of the corresponding property cannot be expanded. (`ConfigValue` is part of it even if it is not stated into the spec but IMHO it makes more sense)
* Removes the annotation `@Experimental` from `getConfigValue` as it is now covered by the spec

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
+++ b/implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
@@ -242,12 +242,17 @@ public class SmallRyeConfig implements Config, Serializable {
     }
 
     /**
-     * 
-     * This method handles calls from both {@link Config#getValue} and {@link Config#getOptionalValue}.<br>
+     * This method handles calls from {@link Config#getValue}.
+     */
+    public <T> T getValue(String name, Converter<T> converter) {
+        return getValue(name, getConfigValue(name, false), converter);
+    }
+
+    /**
+     * This method handles calls from both {@link Config#getValue} and {@link Config#getOptionalValue}.
      */
     @SuppressWarnings("unchecked")
-    public <T> T getValue(String name, Converter<T> converter) {
-        final ConfigValue configValue = getConfigValue(name);
+    public <T> T getValue(String name, ConfigValue configValue, Converter<T> converter) {
         if (ConfigValueConverter.CONFIG_VALUE_CONVERTER.equals(converter)) {
             return (T) configValue;
         }
@@ -327,10 +332,32 @@ public class SmallRyeConfig implements Config, Serializable {
         return Objects.equals(expected, getRawValue(name));
     }
 
-    @Experimental("Extension to the original ConfigSource to allow retrieval of additional metadata on config lookup")
     public ConfigValue getConfigValue(String name) {
-        final ConfigValue configValue = configSources.getInterceptorChain().proceed(name);
-        return configValue != null ? configValue : ConfigValue.builder().withName(name).build();
+        return getConfigValue(name, true);
+    }
+
+    /**
+     * Return the {@link org.eclipse.microprofile.config.ConfigValue} for the specified property name from the underlying
+     * {@linkplain ConfigSource
+     * configuration source}. The lookup of the configuration is performed immediately, meaning that calls to
+     * {@link org.eclipse.microprofile.config.ConfigValue} will always yield the same results.
+     *
+     * @param name The configuration property name
+     * @param noSuchElementExceptionForbidden Indicates whether {@code NoSuchElementException} can be thrown or not.
+     * @return the resolved property value as a {@link org.eclipse.microprofile.config.ConfigValue}
+     * @throws NoSuchElementException in case the property value could not be expanded and
+     *         {@code noSuchElementExceptionForbidden} has been set to {@code false}.
+     */
+    public ConfigValue getConfigValue(String name, boolean noSuchElementExceptionForbidden) {
+        try {
+            final ConfigValue configValue = configSources.getInterceptorChain().proceed(name);
+            return configValue != null ? configValue : ConfigValue.builder().withName(name).build();
+        } catch (NoSuchElementException e) {
+            if (noSuchElementExceptionForbidden) {
+                return ConfigValue.builder().withName(name).build();
+            }
+            throw e;
+        }
     }
 
     /**
@@ -346,7 +373,7 @@ public class SmallRyeConfig implements Config, Serializable {
 
     @Override
     public <T> Optional<T> getOptionalValue(String name, Class<T> aClass) {
-        return getValue(name, getOptionalConverter(aClass));
+        return getValue(name, getConfigValue(name), getOptionalConverter(aClass));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


essobedo:
Hi @radcortez WDYT of this PR and the related ticket?

essobedo:
hum, I've just realized that [you added it into the spec](https://github.com/eclipse/microprofile-config/commit/50bc9d9a81f20d70c32891d8ef3085a3360dc1ed), maybe I misunderstood it 🤔  ?

radcortez:
> hum, I've just realized that [you added it into the spec](https://github.com/eclipse/microprofile-config/commit/50bc9d9a81f20d70c32891d8ef3085a3360dc1ed), maybe I misunderstood it 🤔 ? But if so, how to manage use cases where the value of a property is defined with a variable and this property is optional so there is no need to provide a default value?

I guess we forgot to add a TCK test for this case :)

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -327,8 +332,25 @@ public boolean rawValueEquals(String name, String expected) {
...
+     * Return the {@link org.eclipse.microprofile.config.ConfigValue} for the specified property name from the underlying
+     * {@linkplain ConfigSource
+     * configuration source}. The lookup of the configuration is performed immediately, meaning that calls to
+     * {@link org.eclipse.microprofile.config.ConfigValue} will always yield the same results.
+     *
+     * @param name The configuration property name
+     * @return the resolved property value as a {@link org.eclipse.microprofile.config.ConfigValue}
+     * @throws NoSuchElementException in case the property value could not be expanded.
+     */
+    public ConfigValue getRawConfigValue(String name) {
```
radcortez, line 353: I think it is confusing to have two public API's `getConfigValue` and `getRawConfigValue`. Can we use only one?

radcortez (changes requested):

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
essobedo: You mean, you would like me to overload `getConfigValue` and adds a new method of type `getConfigValue(String name, boolean noSuchElementExceptionForbidden)`, is that what you had in mind?

radcortez:
Hum, I'm wondering about the scenario where you have a composed expression like ${property-one}${property-two} and you have one missing mapped into an `Optional` or `ConfigValue`. In that case we assume that the property does not exist as at all and the `Optional` is empty and `ConfigValue` is `null`?

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -327,10 +332,32 @@ public boolean rawValueEquals(String name, String expected) {
...
+     *
+     * @param name The configuration property name
+     * @param noSuchElementExceptionForbidden Indicates whether {@code NoSuchElementException} can be thrown or not.
+     * @return the resolved property value as a {@link org.eclipse.microprofile.config.ConfigValue}
+     * @throws NoSuchElementException in case the property value could not be expanded and
+     *         {@code noSuchElementExceptionForbidden} has been set to {@code false}.
+     */
+    public ConfigValue getConfigValue(String name, boolean noSuchElementExceptionForbidden) {
+        try {
+            final ConfigValue configValue = configSources.getInterceptorChain().proceed(name);
```
radcortez, line 353: I'm maybe thinking that we only need to thrown the `NoSuchElementException` when we cannot map to an actual type. In that case, it probably makes sense for the chain to not throw the Exception in the interceptor and throw it later when doing the conversion. I think it may be worth investigating. It should probably simplify the code.

essobedo: Yes it would be definitively more simpler but if we do so, we won't be able to throw a `NoSuchElementException` with a specific message indicating that the expression could not be expanded, is that ok with you?

essobedo:
> Hum, I'm wondering about the scenario where you have a composed expression like ${property-one}${property-two} and you have one missing mapped into an `Optional` or `ConfigValue`. In that case we assume that the property does not exist as at all and the `Optional` is empty and `ConfigValue` is `null`?

Good point, in that case, as end user, if `property-one` has no value and `property-two` has a value, with the expression `${property-one}${property-two}`, I would expect to get only the value of `property-two`. WDYT ?

essobedo:
@radcortez Hi, don't forget me please :-) Any answers to my questions to share?

implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
radcortez: We need to check what the spec needs. We can still log an error, but we may loose some of the context to throw the proper exception later, unless we record it somehow in `ConfigValue`.

radcortez:
> > Hum, I'm wondering about the scenario where you have a composed expression like ${property-one}${property-two} and you have one missing mapped into an `Optional` or `ConfigValue`. In that case we assume that the property does not exist as at all and the `Optional` is empty and `ConfigValue` is `null`?
> 
> Good point, in that case, as end user, if `property-one` has no value and `property-two` has a value, with the expression `${property-one}${property-two}`, I would expect to get only the value of `property-two`. WDYT ?

I'm not sure if it is the right behaviour because it means we are silently setting `${property-one}` with an empty String. We have to make sure we provide a consistent behaviour. So for instance if `${property-one}` expansion gives a `NoSuchElement` I would expect to get the same even if used in a nested or composed expression. Otherwise, you have a different behaviour depending on the context.

dmlloyd:
If you done want an exception to be thrown on an unknown property, use the `${propname:}` syntax.

essobedo:
> If you don't want an exception to be thrown on an unknown property, use the `${propname:}` syntax.

Well not really because no value is different from an empty String, If my field is of type `Optional<String>`, I would expect to get an `Optional.empty()`. But indeed it is what I currently do to workaround this problem but it is not an ideal solution, don't you agree?

essobedo:
> If you don't want an exception to be thrown on an unknown property, use the `${propname:}` syntax.

My bad, you are right, it works well

essobedo:
I guess it is useless then we can close it

essobedo:
> If you don't want an exception to be thrown on an unknown property, use the `${propname:}` syntax.

I've just realized that I already proposed this solution before https://github.com/smallrye/smallrye-config/pull/588#issuecomment-853771054 but I found that it was not really logical to throw `NoSuchElementException` for Optionals so I expected a native support instead moreover with this workaround you get an empty String with a `ConfigValue`

PR #595, opened 2021-06-15
Do not merge - jakarta update

Emily-Jiang:


radcortez:
We are going to use the same approach as SR Common like here: https://github.com/smallrye/smallrye-common/pull/109

PR #676, opened 2021-12-06
Close ConfigSource instances when releaseConfig is called

brianwyka:
@radcortez, does this address the `TODO` comments in `SmallRyeConfigSourceProvider.releaseConfig(Config)` ?


radcortez:
Thanks for the PR!
 
Unfortunately, this is very tricky, because as stated in https://github.com/eclipse/microprofile-config/issues/136#issuecomment-535962313 and https://github.com/eclipse/microprofile-config/issues/471 the `Config` may be leaked by the user to some other place, so this may end up being a breaking change (and we may have to do this).

Also, if you manually instantiate the `Config` instance via `SmallRyeConfigBuilder` there is no way for you to close the sources because is not associated with any classloader.

We probably need to think a bit more about this and try out a few things first.

radcortez:
This PR has been stalled for a few months. Closing this for now. Please reopen it if you want to keep working on it. Thank you!

PR #696, opened 2022-01-06
fix race condition in ConfigMappingLoader

Philipp-Borchert-ISH:
Hi,

after enabling parallel unit tests on one of my projects I got test failures caused by a non thread safe implementation in ConfigMappingLoader:

```
java.lang.LinkageError: loader 'app' attempted duplicate class definition for io.smallrye.config.ConfigMappingLoaderTest$Server-1289738797Impl. (io.smallrye.config.ConfigMappingLoaderTest$Server-1289738797Impl is in unnamed module of loader 'app')

	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.System$2.defineClass(System.java:2131)
	at java.base/java.lang.invoke.MethodHandles$Lookup.defineClass(MethodHandles.java:962)
	at io.smallrye.common.classloader.ClassDefiner$1.run(ClassDefiner.java:19)
	at io.smallrye.common.classloader.ClassDefiner$1.run(ClassDefiner.java:14)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at io.smallrye.common.classloader.ClassDefiner.defineClass(ClassDefiner.java:14)
	at io.smallrye.config.ConfigMappingLoader.defineClass(ConfigMappingLoader.java:120)
	at io.smallrye.config.ConfigMappingLoader.loadClass(ConfigMappingLoader.java:98)
	at io.smallrye.config.ConfigMappingLoader.getImplementationClass(ConfigMappingLoader.java:81)
	at io.smallrye.config.ConfigMappingLoaderParallelTest.execute(ConfigMappingLoaderParallelTest.java:33)
``` 

PR contains a bit of a "blunt" fix by adding synchronzied modifier to the according method call - not sure if that's acceptable due to the performance penalty.
Furthermore there is a unit test + junit config to enable parallel test execution for this particular test class. If you remove the synchronized modifier the tests should error out.
Second test had to be adapted due to the additional .properties file.

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/ConfigMappingLoader.java
+++ b/implementation/src/main/java/io/smallrye/config/ConfigMappingLoader.java
@@ -81,7 +81,7 @@ public final class ConfigMappingLoader {
         return (Class<? extends ConfigMappingObject>) loadClass(type, mappingMetadata);
     }
 
-    static Class<?> loadClass(final Class<?> parent, final ConfigMappingMetadata configMappingMetadata) {
+    static synchronized Class<?> loadClass(final Class<?> parent, final ConfigMappingMetadata configMappingMetadata) {
         // Check if the interface implementation was already loaded. If not we will load it.
         try {
             final Class<?> klass = parent.getClassLoader().loadClass(configMappingMetadata.getClassName());
--- a/implementation/src/main/java/io/smallrye/config/ConfigMappingLoader.java
+++ b/implementation/src/main/java/io/smallrye/config/ConfigMappingLoader.java
@@ -6,6 +6,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.UndeclaredThrowableException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.eclipse.microprofile.config.inject.ConfigProperties;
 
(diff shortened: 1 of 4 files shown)


radcortez:
Thanks for the PR. Let me think a little bit about this.

implementation/src/main/java/io/smallrye/config/ConfigMappingLoader.java
```diff
@@ -81,7 +81,7 @@ static ConfigMappingInterface getConfigMappingInterface(final Class<?> type) {
         return (Class<? extends ConfigMappingObject>) loadClass(type, mappingMetadata);
     }
 
-    static Class<?> loadClass(final Class<?> parent, final ConfigMappingMetadata configMappingMetadata) {
+    static synchronized Class<?> loadClass(final Class<?> parent, final ConfigMappingMetadata configMappingMetadata) {
```
radcortez, line 84: While it should be mostly fine, since I believe applications don't ship with a huge number of mapping classes, we could probably improve this with a lock to the class name instead (like the `Classloader`). Check https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/java/lang/ClassLoader.java#L663.

Do you mind updating the PR? Thanks!

radcortez (changes requested):

implementation/src/main/java/io/smallrye/config/ConfigMappingLoader.java
Philipp-Borchert-ISH: Thanks for the feedback, PR has been updated accordingly.

radcortez (approved):

radcortez:
Great! Thanks for the PR. I'll push a release in the next few days.

Issue #693, opened 2021-12-25
feature: Support intermediate config directories in FileSystemConfigSource like /configs.d/*/


vladimirfx:
In process of migrating some Spring Boot projects to Quarkus, we discover a lack of this feature.
Our current Pod config looks like this:

```yaml
          env:
            - name: SPRING_CONFIG_ADDITIONAL_LOCATION
              value: optional:configtree:/config.d/*/
```

At deployment time Ops team projects secrets like that:
```yaml
          volumeMounts:
            - name: activemq-config
              mountPath: /config/activemq
              readOnly: true
            - name: smtp-config
              mountPath: /config.d/smpt
              readOnly: true
            - name: database-config
              mountPath: /config.d/database
              readOnly: true
```

In different environments, the exact number and target of such volume mounts may differ (ex. SMTP config set by main app config in DEV).

It would be nice if this work in Quarkus:
```yaml
          env:
            - name: SMALLRYE_CONFIG_SOURCE_FILE_LOCATIONS
              value: /config.d/*/
```

if you are OK with that feature I can provide a PR.

radcortez:
Hi @vladimirfx 

Sure. Feel free to provide a PR. Thanks :)

radcortez:
I am closing this due to the lack of activity.

dmlloyd:
Perhaps a better approach (for UX at least) to solving the same problem would be (for each config source which deals with the filesystem) to recognize whether the target file name is a directory, and if so, read the directory contents in this manner (sorted order, concatenated). I don't recall if the current code is currently set up to easily do this though.

radcortez:
We added support to read the entire content of a directory for most sources, but it is limited to a single level. I believe the request was to support recursive listing (or at least one more level).

Do you think this is something we should pursue?

dmlloyd:
It's probably not worth going farther then at this time.

PR #705, opened 2022-01-22
Support for wildcard file locations (#693)

vladimirfx:
While testing new functionality I've noticed that the current implementation does not support URIs syntax (actual values not read from the filesystem). It is expected behavior or all variations should read URI format (e.g. `file/conf/etc`)?

### DIFF
--- a/sources/file-system/src/main/java/io/smallrye/config/source/file/FileSystemConfigSourceFactory.java
+++ b/sources/file-system/src/main/java/io/smallrye/config/source/file/FileSystemConfigSourceFactory.java
@@ -2,6 +2,8 @@ package io.smallrye.config.source.file;
 
 import static io.smallrye.config.Converters.newArrayConverter;
 
+import java.io.File;
+import java.net.URI;
 import java.util.Collections;
 import java.util.OptionalInt;
 import java.util.stream.Collectors;
@@ -16,6 +18,7 @@ import io.smallrye.config.Converters;
 
 public class FileSystemConfigSourceFactory implements ConfigSourceFactory {
     public static final String SMALLRYE_CONFIG_SOURCE_FILE_LOCATIONS = "smallrye.config.source.file.locations";
+    private static final String WILDCARD_LOCATION_SUFFIX = "/*/";
 
     @Override
     public Iterable<ConfigSource> getConfigSources(final ConfigSourceContext context) {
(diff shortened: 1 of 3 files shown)


sources/file-system/src/main/java/io/smallrye/config/source/file/FileSystemConfigSourceFactory.java
```diff
@@ -26,8 +29,26 @@
...
+                                     return Stream.of(rootLocation.listFiles(pathname -> pathname.isDirectory()))
+                                                  .map(subLocation -> new FileSystemConfigSource(subLocation));
+                                 } else {
+                                     return Stream.empty();
+                                 }
+                             } else {
+                                 return Stream.of(new FileSystemConfigSource(location));
+                             }
+                         }
+                )
```
radcortez, line 51: I believe the logic should move to `FileSystemConfigSource` because it can be used on its own. If we keep this logic here, then you have different behaviors between `FileSystemConfigSource` and `FileSystemConfigSourceFactory`. On the other hand, `FileSystemConfigSourceFactory` will use the sources produced by `FileSystemConfigSource`, so the behaviors will be aligned.

Also, I would recommend implementing the logic with https://docs.oracle.com/javase/8/docs/api/index.html?java/nio/file/PathMatcher.html.

radcortez (changes requested):

radcortez:
> While testing new functionality I've noticed that the current implementation does not support URIs syntax (actual values not read from the filesystem). It is expected behavior or all variations should read URI format (e.g. `file:/conf/etc`)?
> Closes #693 with an always optional variation.

@vladimirfx, thanks for the PR, I've left a comment to review.

We should support `URI` for sure. You can probably check https://github.com/smallrye/smallrye-config/blob/08918436d1a48b411da530be99e3da97df1e5bfe/implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceLoader.java#L83-L122 to see how to support both styles.

sources/file-system/src/main/java/io/smallrye/config/source/file/FileSystemConfigSourceFactory.java
vladimirfx: > We should support `URI` for sure. You can probably check

We should fix that in this PR or file a separate PR for URI support?

radcortez: You can do everything here. Thanks!

radcortez:
This PR has been stalled for a few months. Closing this for now. Please reopen it if you want to keep working on it. Thank you!

PR #747, opened 2022-05-03
Add `ESCAPES` flag to expression parser rules

### LINKED ISSUE #746: Escaping backslash does not work in front of replacement expression
I can't find a way to evaluate successfully something like `x\${x}` to `x\foo`.

With config:
```
x: foo
a: x/${x}
b: x\${x}
c: x\${x}
d: x\\${x}
e: x\\\${x}
f: x\\\\${x}
```

and code:
```
    @ConfigProperty(name = "x") String x;
    @ConfigProperty(name = "a") String a;
    @ConfigProperty(name = "b") String b;
    @ConfigProperty(name = "c") String c;
    @ConfigProperty(name = "d") String d;
    @ConfigProperty(name = "e") String e;
    @ConfigProperty(name = "f") String f;

    @Test
    public void x() {
        log.info("x=" + x);
        log.info("a=" + a);
        log.info("b=" + b);
        log.info("c=" + c);
        log.info("d=" + d);
        log.info("e=" + e);
        log.info("f=" + f);
    }
```

We get:
```
x=foo
a=x/foo
b=x${x}
c=x${x}
d=x\${x}
e=x\\${x}
f=x\\\${x}
```

The `\` seems to escape the `$`, resulting in `${x}` not being evaluated.
But then I can't escape the escape character.

/cc @radcortez

dmlloyd:
Fixes #746. If it works.

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java
+++ b/implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java
@@ -1,5 +1,6 @@
 package io.smallrye.config;
 
+import static io.smallrye.common.expression.Expression.Flag.ESCAPES;
 import static io.smallrye.common.expression.Expression.Flag.LENIENT_SYNTAX;
 import static io.smallrye.common.expression.Expression.Flag.NO_SMART_BRACES;
 import static io.smallrye.common.expression.Expression.Flag.NO_TRIM;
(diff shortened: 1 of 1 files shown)


dmlloyd:
@radcortez WDYT?

radcortez:
> @radcortez WDYT?

Yes. I know what the issue is. Let me comment on the original one. You are going to love this :)

PR #797, opened 2022-08-03
Create an ASM shaded version of SmallRye Config

jeanouii:
SmallRye Config uses ASM which is a common and widely used library. It leads to some version conflicts when using it.
This PR creates an ASM-shaded version.


jeanouii:
Of course. I did not know about the process sorry.
I'll close this one and create the same one against the main branch

radcortez:
Thanks!

PR #799, opened 2022-08-04
Prototype to secure secret keys

radcortez:
This is a prototype to support secret keys.

- We use property expressions to represent a secret key value in the format `${provider_name:provider_custom_representation}
  - For instance a keystore would use ${file-vault:keystore_name/alias} 
  - The Jasypt provider would use ${jasypt:ENC(secret)}
- Backed by an interceptor to process each provider. Such interceptors run before the default expressions, so it doesn't clash.


sources/file-vault/src/main/java/io/smallrye/config/source/filevault/FileVaultConfigSourceInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    public ConfigValue getValue(final ConfigSourceInterceptorContext context, final String name) {
+        ConfigValue configValue = context.proceed(name);
+
+        if (configValue == null) {
+            return null;
+        }
+
+        Expression expression = Expression.compile(configValue.getValue(), LENIENT_SYNTAX, NO_TRIM, NO_SMART_BRACES);
+        String expanded = expression.evaluate(new BiConsumer<ResolveContext<RuntimeException>, StringBuilder>() {
```
dmlloyd, line 27: I believe that having multiple property expanders in series will cause subtle syntactic problems, particularly with nested expressions but probably some other cases as well (default values maybe?). This is why I proposed using the double-colon feature on the existing property expander, and then adding a separate hook to register handlers for it: it avoids any syntactic problems that can occur with multiple series-style interpolations (which should be regarded as an anti-pattern IMO).

radcortez: Sure, all of this can be changed. The idea was to ensure that it could be implemented with the interceptor approach, and I didn't want to start adding new APIs without some validation first.

Separate handlers will require access to the ConfigContext, and we need to figure out when they should be initialized.

PR #801, opened 2022-08-08
Support OpenWebBeans runner for TCK

jeanouii:
This PR allows to run the TCK using OpenWebBeans instead of Weld. It fails with 3 tests and I'd appreciate community's help to solve those.

To be added in the XML Suite definition for TestNG

      <class name="org.eclipse.microprofile.config.tck.ConfigPropertiesTest"></class>
      <class name="org.eclipse.microprofile.config.tck.CDIPlainInjectionTest"></class>

Thanks

### DIFF
--- a/testsuite/tck/pom.xml
+++ b/testsuite/tck/pom.xml
@@ -27,6 +27,10 @@
 
   <name>SmallRye: MicroProfile Config TCK</name>
 
+  <properties>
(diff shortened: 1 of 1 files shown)


radcortez:
Closing in favor of #803. Thank you for the configuration @jeanouii .

PR #814, opened 2022-09-05
Use the root locale during the validation tests

MaisiKoleni:
The problem messages that get checked in the validation tests are locale-sensitive and therefore, the tests fail on some computers with non-English default locale, e.g., German or French. Setting and restoring the root locale should fix the problem without interfering with other tests / steps in the build process.

Note that I chose to add this directly into the test class such that it works in the IDE without running Maven, too.

Some alternatives would be to extract the logic into a super class or JUnit 5 test extension, but this is the only test in this repository with the problem.

Example errors:
```
// German locale
org.opentest4j.AssertionFailedError: expected: <server.port must be greater than or equal to 8000> but was: <server.port muss größer-gleich 8000 sein>
// French locale
org.opentest4j.AssertionFailedError: expected: <server.port must be greater than or equal to 8000> but was: <server.port doit être supérieur ou égal à 8000>
```

### DIFF
--- a/validator/src/test/java/io/smallrye/config/validator/ValidateConfigTest.java
+++ b/validator/src/test/java/io/smallrye/config/validator/ValidateConfigTest.java
@@ -12,6 +12,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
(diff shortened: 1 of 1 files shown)


radcortez (approved):

radcortez:
Thank you :)

PR #815, opened 2022-09-05
Validate maps and collections completely

MaisiKoleni:
Currently, validation does not work properly with maps and collections containing groups.
This PR aims to enable the usage of direct validation on the maps and collections containing the groups.

The original problem I had was similar to the example in the tests:
```java
@EqualLengthKeys
Map<String, LogLevel> levels();
```
but just with the map values instead and can be found here: https://github.com/MaisiKoleni/javadoc-search/pull/6.

My current approach for that is to always validate the property value, but keep the traversal of groups.
It appears that we only need an explicit traversal for groups because the validator already seems to traverse the properties 
recursively as desired.

However, I am not experienced or knowledgeable enough to know if this change breaks anything. It looks like everything is caught now and nothing is validated twice.

I also refactored the tests a bit and added five more validation problems, three of which failed without the change. All the 15 old tests still work, but I am not sure how much of all validation possibilities they cover.

The reason for the custom assertion is that the old approach was not as informative, as it already failed at the size check. The new approach also reduces the danger that the size assertion and the number of individual contains checks deviate.
This is already implemented in test frameworks such as AssertJ, but those are not available here, and thus a custom solution is required for now, unless we can add a test dependency.

Please let me know what you think about the change, I might have missed the reasoning for the old validation logic and would be grateful for any counterexample (I wouldn't want to break someone else's configuration with the change).

### DIFF
--- a/validator/src/main/java/io/smallrye/config/validator/BeanValidationConfigValidator.java
+++ b/validator/src/main/java/io/smallrye/config/validator/BeanValidationConfigValidator.java
@@ -127,10 +127,8 @@ public interface BeanValidationConfigValidator extends ConfigValidator {
                         throw new UndeclaredThrowableException(t2);
                     }
                 }
-            } else if (collectionProperty.getElement().isLeaf()) {
-                validateProperty(collectionProperty.getElement(), currentPath, namingStrategy, mappingObject, optional,
-                        problems);
             }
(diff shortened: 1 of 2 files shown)


radcortez (approved):

radcortez:
It's great. Thanks!

PR #827, opened 2022-09-21
Support an associated mapping in a ConfigSourceFactory

radcortez:


dmlloyd:
What does this do? It's not clear from an initial reading.

radcortez:
The idea is to be able to load a `ConfigMapping` with a `ConfigSourceFactory`. The motivation for this is to drop the Quarkus Bootstrap phase, use the Factory to configure a source (like Consul, Vault, etc), and still be able to have a mapping object to configure the source.

PR #828, opened 2022-09-22
Option to include quotes in NameIterator segments

radcortez:


dmlloyd:
How does this affect the syntax/grammar of a config key?

radcortez:
> How does this affect the syntax/grammar of a config key?

The issue is that `NameIterator` omits the quotes from the segments, so adding a key like `foo."bar".baz` to `KeyMap` when using a `String` or the `NameIterator`,  the `KeyMap` keys are `foo`, `bar` and `baz`. A lookup to `foo."bar".baz` would not work, and you need to use `foo.bar.baz` instead. 

Also, the adding methods to `KeyMap` are not consistent because of `NameIterator`. For instance, if you use the var args `findOrAdd("foo", "\"bar-bar\"", "baz")`, then the `KeyMap` keys are `foo` , `"bar"` and `baz`. With the `String` from (which delegates to `NameIterator`), the `findOrAdd("foo.\"bar-bar\".baz")`, will add the `KeyMap` keys `foo`, `bar` (no quotes) and `baz`.

radcortez:
This is no longer required.

PR #833, opened 2022-09-30
Secret Keys Handlers

radcortez:
Replaces #799.

- We use property expressions to represent a secret key value in the format `${handler_name::encoded_secret}`.
- The `SecretKeys` API, registers handlers by name via the ServiceLoader.
- The `SecretKeysHandler` is executed in the interceptor that deals with secret keys (if there is a match).
- Any source can contain secrets. It only requires the right handlers to decode/decrypt the value.
- Also includes a `KeyStore` source, to demonstrate how it could work with the `SecretKeysHandler`.

There is still some work to complete:
- Being able to configure `SecretKeysHandler`.
- Certificate support
- Out of the box handlers

The goal of the current PR is to just agree on the design and then move forward.


implementation/src/main/java/io/smallrye/config/SecretKeysConfigSourceInterceptor.java
```diff
@@ -1,27 +1,72 @@
...
+        ConfigValue configValue = context.proceed(name);
+
+        if (configValue == null || handlers.isEmpty()) {
+            return configValue;
+        }
+
+        // This executes before the Expression resolution.
+        // If we find a handler we execute it, if not we proceed
+        Expression expression = Expression.compile(configValue.getValue(), LENIENT_SYNTAX, NO_TRIM, NO_SMART_BRACES,
+                DOUBLE_COLON);
```
dmlloyd, line 50: Really the point of using the expression compiler is that we can combine expression resolution with double-colon expansion in a single step. Once we have two separate resolvers, then things get messy including recursive expression support.

What about adding a `resolveSecret(String name, String value)` to `ConfigSourceInterceptorContext`, and adding secret resolvers to the base implementation along with converters and config sources?

radcortez: > Really the point of using the expression compiler is that we can combine expression resolution with double-colon expansion in a single step. Once we have two separate resolvers, then things get messy including recursive expression support.

Maybe I'm missing something, but when I've tried with a single expression compiler and the `DOUBLE_COLON` flag, the entire expression becomes part of the key when the expression contains a single colon, meaning that it would break the default values expressions. Maybe that is a bug from what it is intended? :) 

> What about adding a `resolveSecret(String name, String value)` to `ConfigSourceInterceptorContext`, and adding secret resolvers to the base implementation along with converters and config sources?

The idea was to add the `SecretKeysHandler` (we can rename it to `SecretKeysResolver`) in the base implementation. Right now, they are just discovered via ServiceLoader. We can add APIs to directly access them, I was trying to be minimalist for now. 

How do you intend to use `resolveSecret(String name, String value)`? Expose it as a separate API in `SmallRyeConfig` to look up secrets?

dmlloyd: > > Really the point of using the expression compiler is that we can combine expression resolution with double-colon expansion in a single step. Once we have two separate resolvers, then things get messy including recursive expression support.
> 
> Maybe I'm missing something, but when I've tried with a single expression compiler and the `DOUBLE_COLON` flag, the entire expression becomes part of the key when the expression contains a single colon, meaning that it would break the default values expressions. Maybe that is a bug from what it is intended? :)

You mean when the expression is e.g. `foo:bar::baz:zap` or for normal expressions as well?

> > What about adding a `resolveSecret(String name, String value)` to `ConfigSourceInterceptorContext`, and adding secret resolvers to the base implementation along with converters and config sources?
> 
> The idea was to add the `SecretKeysHandler` (we can rename it to `SecretKeysResolver`) in the base implementation. Right now, they are just discovered via ServiceLoader. We can add APIs to directly access them, I was trying to be minimalist for now.
> 
> How do you intend to use `resolveSecret(String name, String value)`? Expose it as a separate API in `SmallRyeConfig` to look up secrets?

No, it would be usable by interceptors (specifically the expression interceptor, to implement the `::` lookup) but otherwise not accessible from the main API. The implementation(s) would be loaded via `ServiceLoader` as you already have it.

dmlloyd: The interceptor context is AFAIK the only thing that would be available to the expression interceptor that would allow it to interact with the greater configuration system (outside of hacky things like global variables or thread-locals or things like that). That said, we could opt to make the method non-public.

radcortez: We do initialize the `ExpressionConfigSourceInterceptor` in the builder. What if we just discover and populate the `SecretKeysResolvers` directly into that interceptor? If we need them later for outside of the expression interceptor we can rewrite the implementation without impact.

dmlloyd: Sure, that seems reasonable. :+1:

radcortez:
@dmlloyd let me know what you think of the current proposal. Thanks!

implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java
```diff
@@ -49,25 +46,39 @@ private ConfigValue getValue(final ConfigSourceInterceptorContext context, final
...
-                ConfigValue resolve = getValue(context, resolveContext.getKey(), depth + 1);
+                String key = resolveContext.getKey();
+
+                // Requires a handler lookup
+                String[] split = key.split("::");
+                if (split.length == 2) {
+                    value.withHandler(split[0]);
+                    stringBuilder.append(split[1]);
+                    return;
+                }
```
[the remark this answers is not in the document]
radcortez: We just ignore it and proceed, but we can also implement it in that way.

implementation/src/main/java/io/smallrye/config/SecretKeysHandler.java
```diff
@@ -0,0 +1,10 @@
+package io.smallrye.config;
+
+import io.smallrye.common.annotation.Experimental;
+
+@Experimental("")
+public interface SecretKeysHandler {
+    String handleSecret(String secret);
```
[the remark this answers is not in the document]
radcortez: Let's rename it.

implementation/src/main/java/io/smallrye/config/SecretKeys.java
```diff
@@ -1,13 +1,42 @@
...
-        throw new UnsupportedOperationException();
+    private final Set<String> secrets;
+    private final Map<String, SecretKeysHandler> handlers;
+
+    SecretKeys(final Set<String> secrets, final Map<String, SecretKeysHandler> handlers) {
+        this.secrets = secrets;
+        this.handlers = handlers;
+    }
+
+    public void verifySecret(final String name) {
```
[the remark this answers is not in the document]
radcortez: This is to check if the key is a secret (with the Secrets set that can be configured in the builder) and if we can access the secret. Lets rename it.

implementation/src/main/java/io/smallrye/config/SecretKeysHandlerConfigSourceInterceptor.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    @Override
+    public ConfigValue getValue(final ConfigSourceInterceptorContext context, final String name) {
+        ConfigValue configValue = context.proceed(name);
+        if (configValue != null && configValue.getValue() != null) {
+            String handler = configValue.getHandler();
+            if (handler != null) {
+                if (SecretKeys.isLocked()) {
+                    //throw ConfigMessages.msg.notAllowed(name);
```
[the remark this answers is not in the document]
radcortez: I can't remember :D But I think it is not needed.

sources/keystore/src/main/java/io/smallrye/config/source/keystore/KeyStoreConfigSourceFactory.java
```diff
@@ -0,0 +1,175 @@
...
+import io.smallrye.config.ConfigurableConfigSource;
+import io.smallrye.config.PropertiesConfigSource;
+import io.smallrye.config.SmallRyeConfig;
+import io.smallrye.config.SmallRyeConfigBuilder;
+import io.smallrye.config.source.keystore.KeyStoreConfig.KeyStore.Alias;
+
+public class KeyStoreConfigSourceFactory implements ConfigSourceFactory {
+    @Override
+    public Iterable<ConfigSource> getConfigSources(final ConfigSourceContext context) {
+        SmallRyeConfig config = new SmallRyeConfigBuilder()
```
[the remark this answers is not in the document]
radcortez: Not exactly. This is just a trick to be able to use a mapping to configure a ConfigSource. 

For instance, consider the mapping:
https://github.com/radcortez/smallrye-config/blob/b2bf955328cf8e4382c364f898173477d80d77a3/sources/keystore/src/main/java/io/smallrye/config/source/keystore/KeyStoreConfig.java#L11

How can we create this mapping? Remember, we use `SmallRyeConfig` to call certain API to populate the mapping instance. Ideally, we should decouple that, so we don't need to use the trick.

We create an empty `SmallRyeConfig` instance, just with the mapping class and a `ConfigSource` that delegates to the `ConfigSourceContext`:

https://github.com/radcortez/smallrye-config/blob/b2bf955328cf8e4382c364f898173477d80d77a3/sources/keystore/src/main/java/io/smallrye/config/source/keystore/KeyStoreConfigSourceFactory.java#L38-L41

https://github.com/radcortez/smallrye-config/blob/b2bf955328cf8e4382c364f898173477d80d77a3/sources/keystore/src/main/java/io/smallrye/config/source/keystore/KeyStoreConfigSourceFactory.java#L70-L97

So we can get an instance of the mapping class, by wrapping all of that, but still use the original config context being build.

implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java
```diff
@@ -53,25 +50,39 @@ private ConfigValue getValue(final ConfigSourceInterceptorContext context, final
             return null;
         }
 
-        List<Problem> problems = new ArrayList<>();
+        ConfigValue.ConfigValueBuilder value = configValue.from();
         Expression expression = Expression.compile(escapeDollarIfExists(configValue.getValue()), LENIENT_SYNTAX, NO_TRIM,
-                NO_SMART_BRACES);
+                NO_SMART_BRACES, DOUBLE_COLON);
```
Pepo48, line 55: @radcortez hey Roberto, recently I had a chance to test this PR together with Keycloak. I encountered an issue with `quarkus.locales/quarkus.default-locale` properties. This issue probably affects even some other properties, the locale related ones were just first that I spotted. 
As soon as I added the SNAPSHOT package (with the changes from this PR included) to the Keycloak project, I got an error while trying to build the KC distribution:
```
SRCFG00014: The config property quarkus.locales is required but it could not be found in any config source
````
This was quite weird signal, because afaik by default the locale related properties come from the default config source (if they're not defined somewhere else) set by Quarkus.
Upon further debugging I found out `quarkus.locales` is indeed null, probably because the `${user.language:en}-${user.country:}` expression is not propagated/expanded properly (see the attached screenshot).

I went a little bit further and narrowed the pool of possible culprits to the `DOUBLE_COLON` expression flag. If I remove it and build the package again, everything seems to work properly.

I assume that this expression flag is used primarily for parsing/evaluating expressions like `${aes-gcm-nopadding::encoded::x}` in order to get the encoded part. But somehow it seems to be in collision with the type of expressions mentioned above.

By the way, thanks for all the effort you and your team do around the project! Possibility to secure secrets will help us a lot.

telegram-cloud-photo-size-4-5974175736824707476-y

Comment (radcortez) [reply implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java:55]: Hey @Pepo48,

Thanks for taking the time to try the PR out and finding issues. Highly appreciated :)

Let me have a look.

Comment (Pepo48) [reply implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java:55]: @radcortez you're right, we take our dependencies from quarkus-bom (currently in version 2.13.6.Final), which contains `smallrye-common-expression` version 1.13.1. I overrode only the `smallrye-config` dependencies. 

The bump resolved it. Sorry for bothering, I could've checked the common project's recent issues first.
Thanks for your time!

Comment (radcortez) [reply implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java:55]: No worries, it wasn't that evident :)

Please let me know how the API works and if it solves your issues. We don't have secret handlers implementations yet, but that will come. You can probably try the `jasypt` one or implement your own.

I'm happy to add any missing features or APIs.

**Comment by radcortez** [comment]
> Also, I'm sorry I'm discussing it here but I couldn't find a related issue on the matter.

You don't need to apologize :)

> I'm curious to learn if it shouldn't also be possible to load changes of these secrets as well? For instance, a file containing a password file should probably trigger an event if it is changed so that a consuming application could simply reload. However, as mentioned, it seems some part of this project support it but I'm finding it hard to understand if this PR builds on that or not? To be more specific: will a change of a secretvalue lead to an event being fired

No, we don't build on that. What determines how a configuration is read (and if it is reloadable), is the `ConfigSource`. This PR, only adds the ability to interpret values as secrets and handle them. It is certainly possible with the events extension to provide a `ConfigSource` that triggers the CDI event on a config change or reload.

> To give an example: will a change of a secretvalue lead to an event being fired that can be caught by, let's say, KeyCloak (which uses quarkus) which would be able to restart a connection to a backend resource?

Our stance with Quarkus is that we don't support reloadable configuration (at least the configuration on the `quarkus` namespace). Everyone else is free to do what they want :) The big issue is that is hard to manage user expectations about a configuration reload. Certain configurations will always require a server restart, while others we could get away without, so we prefer to be consistent and always require a restart if you change the Quarkus configuration.

**Comment by UXabre** [comment]
Thanks for your elaborate reply on the matter, really appreciate it and I understand it fully :-) Happy coding!

File: implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java (hunk 7 above)
Comment (radcortez) [reply implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java:55]: @Pepo48 wondering if you have any feedback to share? :)

**Comment by Pepo48** [comment]
Hey @radcortez, as I'm continuing with the spike around this PR, I encountered another issue (hopefully this time it's not something already known 😊) .

It seems that when you work with the Keystore Config Source and `addProfileConfigSource` method tries to load the optional profile resource (for example in this case it can be keystore**-prod**, which doesn't have to exist), a runtime exception thrown here:
https://github.com/smallrye/smallrye-config/blob/dddcc4b900b10a5d891aa427984a2306623e0cf2//sources/keystore/src/main/java/io/smallrye/config/source/keystore/KeyStoreConfigSourceFactory.java#L149-L152
is not handled properly. In my scenario, where single property file is shared among multiple tests, it affected all of them. 

I tried a quick fix, where I caught the exception here:
https://github.com/smallrye/smallrye-config/blob/dddcc4b900b10a5d891aa427984a2306623e0cf2/implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceLoader.java#L235
So now the line looks like this:
```java
catch (FileNotFoundException | NoSuchFileException | RuntimeException e) {
```
and it helped. But you may find a more elegant solution, for sure 😊.
If it will be needed I can share a reproducer as well.

Thanks!

implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java
Pepo48: As I stated above, this one was solved by overriding the dependency as you suggested. Since then haven't encountered any other issues connected to it.

radcortez:
Thanks! Let me have a look!

Pepo48:
> Thanks! Let me have a look!

@radcortez I just wanted to ask whether there is a progress regarding this?

On top of that, what secret handlers are planned to be present in the final implementation? I tried the `jasypt` one and it seems to cover our needs for now.

Is this still aimed for the Quarkus 3 release?

Thanks!

radcortez:
Hey. I'm sorry, I was on PTO the last few days so I couldn't look into it, but it is on my plans to look as soon as possible. I still need to do the final move to Jakarta (including other SR Projects), and then we should be ready to go with new features.

Yes, the `jasypt` one is planned to be included and still aimed at Quarkus 3. 

The idea is to add handlers as required, so if you need a specific one, feel free to add it.

radcortez:
@Pepo48 

I have fixed some of the issues and added some of the missing pieces (namely, supporting a way to configure the secret handlers).

Can you take a final look? Thanks :)

michalvavrik:
Hey @radcortez ,

do you think it is possible for you to release SmallRye Config and bump the version in Quarkus once this gets merged? I'd like to do QE test development for this - https://issues.redhat.com/browse/QUARKUS-2727. I hope release won't eat to much of your time.

Thank you

radcortez:
I'm currently working on the Quarkus integration bits and I believe it should be completed in the next few days. Thanks!

PR #871, opened 2023-01-05
unwrap config instead of just cast

craig-day:
My app uses MicroProfile Config generically, and also uses the SmallRye implementation. My config provider returns a class called `SmallRyeAppSettings` which `implements Config`. I think this is valid according to the MP spec. However, when I try to inject config, I end up with a class cast exception.

```
Caused by: java.lang.ClassCastException: class my.app.SmallRyeAppSettings cannot be cast to class io.smallrye.config.SmallRyeConfig (my.app.SmallRyeAppSettings and io.smallrye.config.SmallRyeConfig are in unnamed module of loader 'app')
```

I traced the error to this specific cast. 

Following the docs from both MP and SmallRye, I think we should use `unwrap` when handling results from `ConfigProvider.getConfig()` rather than casting. I added an instanceof shortcut check here to maintain the previous code path. Probably not necessary. 

Alternatively, should this be using the small rye provider directly? Maybe this?

```
SmallRyeConfigProviderResolver.instance().getConfig(getContextClassLoader());
```
I'm not sure if this will create a circular dependency so I went with the more generic unwrap route first.

I tried to write a test, but to register an override for a `ConfigProvider`, you need to create something `ServiceLoader` can pick up. I couldn't find a nice way to do something for just a single test class that didn't impact all the other test classes.

### DIFF
--- a/cdi/src/main/java/io/smallrye/config/inject/ConfigPropertiesInjectionBean.java
+++ b/cdi/src/main/java/io/smallrye/config/inject/ConfigPropertiesInjectionBean.java
@@ -52,7 +52,7 @@ public class ConfigPropertiesInjectionBean<T> implements Bean<T> {
             }
         }
 
-        SmallRyeConfig config = (SmallRyeConfig) ConfigProvider.getConfig(getContextClassLoader());
+        SmallRyeConfig config = ConfigProvider.getConfig(getContextClassLoader()).unwrap(SmallRyeConfig.class);
         return config.getConfigMapping(getBeanClass(), prefix);
     }
 
--- a/cdi/src/main/java/io/smallrye/config/inject/ConfigInjectionBean.java
+++ b/cdi/src/main/java/io/smallrye/config/inject/ConfigInjectionBean.java
@@ -106,7 +106,7 @@ public class ConfigInjectionBean<T> implements Bean<T>, PassivationCapable {
             } else {
                 Optional<T> optionalValue = (Optional<T>) getConfig().getOptionalValue(key, annotatedTypeClass);
                 return optionalValue.orElseGet(
-                        () -> (T) ((SmallRyeConfig) getConfig()).convert(defaultValue, annotatedTypeClass));
+                        () -> (T) getConfig().unwrap(SmallRyeConfig.class).convert(defaultValue, annotatedTypeClass));
             }
         }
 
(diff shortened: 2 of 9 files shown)


cdi/src/main/java/io/smallrye/config/inject/ConfigPropertiesInjectionBean.java
```diff
@@ -52,8 +53,16 @@ public T create(final CreationalContext<T> creationalContext) {
...
+        Config config = ConfigProvider.getConfig(getContextClassLoader());
+        SmallRyeConfig smallRyeConfig;
+
+        if (config instanceof SmallRyeConfig) {
+            smallRyeConfig = (SmallRyeConfig) config;
+        } else {
+            smallRyeConfig = config.unwrap(SmallRyeConfig.class);
+        }
+
+        return smallRyeConfig.getConfigMapping(getBeanClass(), prefix);
```
dmlloyd, line 65: Couldn't something like this work?

```suggestion
        SmallRyeConfig config = ConfigProvider.getConfig(getContextClassLoader().unwrap(SmallRyeConfig.class);
        return config.getConfigMapping(getBeanClass(), prefix);
```

craig-day: Yep, as I mentioned the instanceof check was just to keep a more similar codepath if it was already a `SmallRyeConfig`, but we should be able to use unwrap directly with no checking.

craig-day: Changed.

radcortez:
Hi, @craig-day Thanks for the PR.

We have other places in the code where we are casting to `SmallRyeConfig`. Do you mind also fixing these cases?

craig-day:
@radcortez I've updated all the casts I could find that were not test. Do you think we should unwrap in test too?

radcortez:
> @radcortez I've updated all the casts I could find that were not test. Do you think we should unwrap in test too?

Yes, please, go ahead. Thank you!

craig-day:
@radcortez I replaced all the casts with `unwrap` in test too. Also updated to latest main and squashed my commits. Thanks!

radcortez:
Can you run the build locally to fix the formatting issues and push again? :) Thanks!

craig-day:
@radcortez updated and pushed :D

radcortez:
Great! Thank you!

radcortez (approved):

PR #877, opened 2023-01-18
provide a way to release config via classloader

craig-day:
I ran into problems with config already being registered, presumably due to races related to configs and class loaders. 

Basically, my app has a function like this

(Note: these are just simplified snippets of the code that I think are relevant here)

```java
  private static AppSettings register(SmallRyeAppSettings settings) {
    ConfigProviderResolver.instance().releaseConfig(ConfigProvider.getConfig());
    ConfigProviderResolver.instance().registerConfig(settings, Thread.currentThread()
        .getContextClassLoader());
    return settings;
  }
```

Sometimes I'll get an `IllegalStateException` saying the config is already registered. Even if I update the function to fetch config with the class loader, it doesn't always fix it. 

After doing some looking I noticed a few TODO comments that pointed to seemingly related issues. Most specifically, I think this is the root of the problem: https://github.com/eclipse/microprofile-config/issues/471

It should be reasonable to release some config and then be able to re-register it in a reliable fashion. As pointed out by that issue, you cannot. Registration is done with a given class loader, but releasing is not. The class loader is the key in the underlying map, but release does not use it. Instead, release iterates all the configs and removes any that are the same instance (not even just equivalent to), regardless of class loader. I know that is basically what the spec says, which is what brought me here. 

Does it seem reasonable to add a new public function within smallrye that allows a consumer to release config for a given class loader? To reuse my previous helper function, I think it would be useful to update it to something like this: 

```java
  private static AppSettings register(SmallRyeAppSettings settings) {
    ClassLoader cl = Thread.currentThread().getContextClassLoader();
    ConfigProviderResolver cpr = ConfigProviderResolver.instance();

    if (cpr instanceof SmallRyeConfigProviderResolver) {
      ((SmallRyeConfigProviderResolver) cpr).releaseConfig(settings, cl);
    }
    cpr.releaseConfig(ConfigProvider.getConfig());
    cpr.registerConfig(settings, cl);

    return settings;
  }
```

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/SmallRyeConfigProviderResolver.java
+++ b/implementation/src/main/java/io/smallrye/config/SmallRyeConfigProviderResolver.java
@@ -130,6 +130,14 @@ public class SmallRyeConfigProviderResolver extends ConfigProviderResolver {
         }
     }
 
+    public void releaseConfig(ClassLoader classLoader) {
+        final ClassLoader realClassLoader = getRealClassLoader(classLoader);
+        final Map<ClassLoader, Config> configsForClassLoader = this.configsForClassLoader;
+        synchronized (configsForClassLoader) {
+            configsForClassLoader.remove(realClassLoader);
+        }
+    }
+
     static ClassLoader getRealClassLoader(ClassLoader classLoader) {
         if (classLoader == null) {
             classLoader = getContextClassLoader();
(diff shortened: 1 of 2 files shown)


craig-day:
@dmlloyd and @radcortez do you have any thoughts on this? I see you both on the issues and am curious if this is a non-starter, or if an approach along these lines is worth working on further.

radcortez:
Hi @craig-day,

I think it is reasonable for us to add a new method that releases the Config by Classloader. Please add some tests to this; we will be happy to take it.

craig-day:
@radcortez got a little distracted from this one, but I put up a simple test.

radcortez (approved):

radcortez:
Thanks!

PR #910, opened 2023-03-18
Add converter for java.nio.file.Path

### LINKED ISSUE #909: Add converter for java.nio.Path
I initially thought `java.nio.Path` would be supported out-of-the-box, given the automatic fallback to a static .of() method that takes a String argument.

But it does not work:
`SRCFG02007: No Converter registered for interface java.nio.file.Path`

It's probably because https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Path.html#of(java.lang.String,java.lang.String...) takes the additional String... vararg argument.

I think the usecase of injecting a java.nio.Path value may be common enough to warrant an out-of-the-box converter?

jpraet:
Fixes #909

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/Converters.java
+++ b/implementation/src/main/java/io/smallrye/config/Converters.java
@@ -175,7 +175,7 @@ public final class Converters {
             newTrimmingConverter(newEmptyValueConverter(Pattern::compile)));
 
     static final Converter<Path> PATH_CONVERTER = BuiltInConverter.of(18,
-            newTrimmingConverter(newEmptyValueConverter(Path::of)));
+            newEmptyValueConverter(Path::of));
 
     static final Map<Class<?>, Class<?>> PRIMITIVE_TYPES;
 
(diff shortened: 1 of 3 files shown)


implementation/src/main/java/io/smallrye/config/Converters.java
```diff
@@ -173,6 +174,9 @@ private Converters() {
     static final Converter<Pattern> PATTERN_CONVERTER = BuiltInConverter.of(17,
             newTrimmingConverter(newEmptyValueConverter(Pattern::compile)));
 
+    static final Converter<Path> PATH_CONVERTER = BuiltInConverter.of(18,
+            newTrimmingConverter(newEmptyValueConverter(Path::of)));
```
dmlloyd, line 178: This one shouldn't be a trimming converter IMO. Paths may contain, start with, or end with spaces. Not that that's a good idea to do 🙂

dmlloyd (approved):

radcortez (approved):

PR #925, opened 2023-04-18
Minor documentation fixes

manofthepeace:
I'd like to understand what the doLocked actually does in the interceptor :) I created an interceptor and I am not sure if I actually need it or not. But surely the example in this PR would not compile without the static import.


dmlloyd (approved):

radcortez (approved):

radcortez:
Thanks :)

PR #927, opened 2023-04-21
Support unnamed Map keys

radcortez:


implementation/src/main/java/io/smallrye/config/WithUnnamedKey.java
```diff
@@ -0,0 +1,23 @@
...
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.METHOD, ElementType.TYPE_USE })
+public @interface WithUnnamedKey {
+    /**
+     * The key name to use to populate configuration values from a configuration path without a key.
+     *
+     * @return the Map key
+     */
+    String value() default "unnamed";
```
yrodiere, line 22: Is there any way this could default to `null` instead? I mean an actual `null`, not the string `"null"` of course.

Asking because in Hibernate Search for example, we have a map of backend config, and the "default"/"unnamed" backend typically has its name set to `null` throughout all SPIs. So forcing a non-null key here leads to all sorts of conversions that introduce some fragility (if I forget to convert at some point, it will likely not work and might not be spotted by tests).

radcortez: I guess we could. This was to provide a more explicit lookup and avoid having to look for `null` key, but it should be fine.

documentation/src/main/docs/config/mappings.md
```diff
@@ -389,6 +389,36 @@ maps to the `Map<String, List<Alias>> aliases()` member, where `localhost` is th
...
+
+The `sever.aliases.name` is an unnamed `Map` property, because it does not contain the `Map` key to populate the `Map` 
+entry. Due to `@WithUnnamedKey("localhost")` the `Map` key is not required in the configuration path. The key used to 
+look up the Map entry is given by `io.smallrye.config.WithUnnamedKey#value`:
+
+```java
+Server server = config.getConfigMapping(Server.class);
+Map<String, Alias> localhost = server.aliases.get("localhost");
+```
+
```
dmlloyd, line 421: It should also be explained what happens if the unnamed key (in this case `localhost`) is explicitly mentioned in the map (what is the precedence order, or is there an error, etc.). Also it could be good to show what a YAML example of this would look like.

radcortez: Correct. I forgot to clarify that.

At the moment, the unnamed key always has precedence over the explicit key with the same name, meaning that properties that use an explicit key with the same name as the unnamed key will get effectively ignored. I've consider  reading both and merge them, but I thought it would be a bit confusing to mix both, and don't have clarity when looking into an `application.properties` file.

I can probably add a warning message for such cases, instead of silently ignoring them.

yrodiere: I guess it's one more argument for defaulting to a `null` map key, since that's the only way we could use the feature without preventing users from using a particular key?

(Though don't take this wrong, the limitation is perfectly fine in most cases)

implementation/src/main/java/io/smallrye/config/WithUnnamedKey.java
radcortez: I've changed the behaviour to use the null key. Can you try it out?

yrodiere: I updated my PR (https://github.com/quarkusio/quarkus/pull/32854) and I confirm null keys work fine. Thank you!

radcortez: Great!

Issue #935, opened 2023-05-10
Class level validations on configuration don't report the path of the failing property element


BenjaminTanguay:
I have a quarkus project that uses smallrye-config and hibernate-validator to handle the project configuration and its validation. Because the project is configuration driven, I want to support the developpers using it as much as possible by providing useful error messages when a configuration is invalid. 

I have an interface that represents a portion of the configuration with some validation logic in it:
```
// @CrossValidation(crossValidationType = CrossValidationType.THEN_RESPOND_MESSAGE_FIELD_COHERENT_WITH_RESPONSE_STRATEGY,
//        message = "Message field must be defined when response strategy is set to MESSAGE")
public interface ThenRespond {
    @NotNull
    @WithConverter(ResponseStrategy.ResponseStrategyConverter.class)
    ResponseStrategy responseStrategy();
    Optional<@Min(100) @Max(599) String> code();

    Optional<String> message();

    @AssertTrue(message = "Message field must be defined when response strategy is set to MESSAGE")
    private boolean isResponseStrategyAndMessageFieldCoherent() {
        return responseStrategy() != ResponseStrategy.MESSAGE ||
                message().isPresent();
    }
}
```

Here, `@Min` and `@Max` refer to custom function level annotations that validate the bounds of the property found under code(). Those annotations yield the expected result. When my configuration contains a code property that is out of bound, I get the following error message:
io.smallrye.config.ConfigValidationException: Configuration validation failed:
	**_universal-stub.stubs[2].defaultRule.thenRespond.code_** @Max validation failed. Field must have a lower or equal value to 599. Value found: 700

Notice here I get the full path of the failing configuration element.

This interface also contains a class level cross-validation between two fields. As a test, I've implemented this cross validation in two different ways: through a custom @CrossValidation annotation at the top of the class and through a private method wired with the @AssertTrue annotation. 

Both mecanisms produce a similar error message:

io.smallrye.config.ConfigValidationException: Configuration validation failed:
	responseStrategyAndMessageFieldCoherent Message field must be defined when response strategy is set to MESSAGE

Notice here how there is no path pointing towards the failing property in the configuration. I would instead have expected an error message having a shape similar to this:

io.smallrye.config.ConfigValidationException: Configuration validation failed:
	universal-stub.stubs[2].defaultRule.thenRespond.responseStrategyAndMessageFieldCoherent Message field must be defined when response strategy is set to MESSAGE

[...]

PR #938, opened 2023-05-18
Added path information to failing class level validations

BenjaminTanguay:
Resolves issue #935

### DIFF
--- a/validator/src/main/java/io/smallrye/config/validator/BeanValidationConfigValidator.java
+++ b/validator/src/main/java/io/smallrye/config/validator/BeanValidationConfigValidator.java
@@ -38,7 +38,7 @@ public interface BeanValidationConfigValidator extends ConfigValidator {
         if (mappingInterface != null) {
             validateMappingInterface(mappingInterface, prefix, mappingInterface.getNamingStrategy(), mappingObject, problems);
         } else {
-            validateMappingClass(mappingObject, problems);
+            validateMappingClass(mappingObject, problems, prefix);
         }
 
         if (!problems.isEmpty()) {
@@ -63,7 +63,7 @@ public interface BeanValidationConfigValidator extends ConfigValidator {
             validateProperty(property, currentPath, namingStrategy, mappingObject, false, problems);
         }
 
-        validateMappingClass(mappingObject, problems);
+        validateMappingClass(mappingObject, problems, currentPath);
     }
 
     default void validateProperty(
@@ -225,11 +225,15 @@ public interface BeanValidationConfigValidator extends ConfigValidator {
         }
     }
 
-    default void validateMappingClass(final Object mappingObject, final List<Problem> problems) {
+    default void validateMappingClass(
+            final Object mappingObject,
+            final List<Problem> problems,
+            final String currentPath) {
         final Set<ConstraintViolation<Object>> violations = getValidator().validate(mappingObject);
         for (ConstraintViolation<Object> violation : violations) {
-            problems.add(violation.getPropertyPath().toString().isEmpty() ? new Problem(violation.getMessage())
-                    : new Problem(violation.getPropertyPath() + " " + violation.getMessage()));
+            problems.add(violation.getPropertyPath().toString().isEmpty() ?
+                    new Problem(currentPath + " " + violation.getMessage())
+                    : new Problem(currentPath + " " + violation.getPropertyPath() + " " + violation.getMessage()));
         }
     }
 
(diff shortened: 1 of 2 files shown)


validator/src/test/java/io/smallrye/config/validator/ValidateConfigTest.java
```diff
@@ -26,11 +26,7 @@
 import jakarta.validation.ConstraintValidator;
 import jakarta.validation.ConstraintValidatorContext;
 import jakarta.validation.Payload;
-import jakarta.validation.constraints.Max;
-import jakarta.validation.constraints.Min;
-import jakarta.validation.constraints.NotBlank;
-import jakarta.validation.constraints.PositiveOrZero;
-import jakarta.validation.constraints.Size;
+import jakarta.validation.constraints.*;
```
manofthepeace, line 29: I think there should be no start imports.

validator/src/main/java/io/smallrye/config/validator/BeanValidationConfigValidator.java
```diff
@@ -225,11 +225,15 @@ default void validatePropertyValue(
...
-    default void validateMappingClass(final Object mappingObject, final List<Problem> problems) {
+    default void validateMappingClass(
+            final Object mappingObject,
+            final List<Problem> problems,
+            final String currentPath) {
         final Set<ConstraintViolation<Object>> violations = getValidator().validate(mappingObject);
         for (ConstraintViolation<Object> violation : violations) {
-            problems.add(violation.getPropertyPath().toString().isEmpty() ? new Problem(violation.getMessage())
-                    : new Problem(violation.getPropertyPath() + " " + violation.getMessage()));
+            problems.add(violation.getPropertyPath().toString().isEmpty() ?
```
radcortez, line 234: The path should be built with `dot` as a separator.

validator/src/test/java/io/smallrye/config/validator/ValidateConfigTest.java
```diff
@@ -124,7 +123,7 @@ void validateConfigMapping() {
                 "server.info.admins.root[1].username size must be between 0 and 4",
                 "server.info.admins.root size must be between 0 and 1",
                 "server.info.firewall.accepted[1] size must be between 8 and 15",
-                "server is not prod");
+                "server server is not prod");
```
radcortez, line 126: Add more levels of nesting to ensure the path is being built correctly.

radcortez (changes requested):

radcortez:
Thank you for the PR. I've left some comments.

Also, CI is not happy.

validator/src/test/java/io/smallrye/config/validator/ValidateConfigTest.java
BenjaminTanguay: I added a class level validation that has more nesting elsewhere in the test. You can find it a few lines above this one. The test related to this validation is:
```"server.cors.origins[2] someClassLevelCrossValidation If host is server3, then port value must be 3",```

validator/src/main/java/io/smallrye/config/validator/BeanValidationConfigValidator.java
BenjaminTanguay: Sure, I'll make the modifications. Just to be clear, in the example provided in the test, the error message will go from `"server.cors.origins[2] someClassLevelCrossValidation If host is server3, then port value must be 3"` to `"server.cors.origins[2].someClassLevelCrossValidation If host is server3, then port value must be 3"`.

violation.getPropertyPath() returns a path object that starts at the level of the class that contains the validation which means that there is generaly not much information available.

radcortez: Ok, I guess I was confused a bit about how it was being built. I think that the first part of the message should always map to configuration property name, so it is clear for the user what to change.

In this case, the `someClassLevelCrossValidation` is not part of the configuration name, so we shouldn't append it with a `dot`, but it can be included in the message.

validator/src/test/java/io/smallrye/config/validator/ValidateConfigTest.java
radcortez: It is enough.

What I don't understand is the double `server` in the message. Shouldn't we just use the message as is from the custom annotation?

BenjaminTanguay: In the test, the first "server" is the full path to the failing object. Then the following string "server is not prod" is the error message defined for this validation.

The result is weird but the algorithm works as intended.

validator/src/main/java/io/smallrye/config/validator/BeanValidationConfigValidator.java
BenjaminTanguay: Sure, I'll revert this portion to a space.

radcortez (approved):

radcortez:
Thanks for the PR. Appreciated!

PR #983, opened 2023-08-24
Introduce SmallRyeConfig.subset

### LINKED ISSUE #981: Support subsets from a Config
Because the `Config` interface has a nice way to convert values and handle optional values, it would be nice if the API allowed to extract a subset of a `Config` to pass it as a parameter to methods that don't care about the external structure: 

Eg.

```properties
app.foo.username=guest
app.foo.password=guest
```

```java
SmallRyeConfig config = ...
Config subset = config.subset("app.foo");
// Now I can call methods taking the config as a parameter that only cares about the leaf nodes:
configure(subset);

public void configure(Config subset) {
   String username = subset.getValue("username",String.class); // This would return "guest"
}
```

gastaldi:
- Fixes #981

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/SmallRyeSubsetConfig.java
+++ b/implementation/src/main/java/io/smallrye/config/SmallRyeSubsetConfig.java
@@ -1,0 +1,93 @@
+package io.smallrye.config;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.ConfigValue;
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.eclipse.microprofile.config.spi.Converter;
+
+/**
+ * @author George Gastaldi
+ */
+class SmallRyeSubsetConfig implements Config {
+
+    private final String prefix;
+
+    private final Config delegate;
+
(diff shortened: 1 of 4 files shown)


implementation/src/main/java/io/smallrye/config/SmallRyeSubsetConfig.java
```diff
@@ -0,0 +1,85 @@
...
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.ConfigValue;
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.eclipse.microprofile.config.spi.Converter;
+
+/**
+ * @author George Gastaldi
+ */
+public class SmallRyeSubsetConfig implements Config {
```
dmlloyd, line 16: I wouldn't make this public.

dmlloyd (changes requested):
I'd only make one change for now (make the implementation class non-public). There are a few other places which could be more optimal given some internal architecture changes (specifically, iterating the nested keys).

implementation/src/main/java/io/smallrye/config/SmallRyeSubsetConfig.java
gastaldi: Done

dmlloyd (dismissed):
LGTM if @radcortez is OK with it.

gastaldi:
Pushed a small fix that filters out properties that don't belong to the subset

PR #984, opened 2023-08-27
Reduce String/StringBuilder allocations

franz1981:
This is addressing https://github.com/quarkusio/quarkus/issues/35406 is the most brutal way: reducing immutability and ensuring to reduce hot-paths useless allocations as much as possible

### DIFF
--- a/common/src/main/java/io/smallrye/config/common/utils/StringUtil.java
+++ b/common/src/main/java/io/smallrye/config/common/utils/StringUtil.java
@@ -165,51 +165,48 @@ public class StringUtil {
             if ('_' == c) {
                 if (i == 0) {
                     // leading _ can only mean a profile
-                    sb.append("%");
+                    sb.append('%');
                     continue;
                 }
 
                 // Do not convert to index if the first segment is a number
                 if (beginSegment > 0) {
-                    try {
-                        String segment = sb.substring(beginSegment, i);
-                        Integer.parseInt(segment);
-                        sb.replace(beginSegment - 1, beginSegment, "[").append("]");
+                    if (isNumeric(sb, beginSegment, i)) {
+                        sb.setCharAt(beginSegment - 1, '[');
+                        sb.append(']');
 
                         int j = i + 1;
                         if (j < length) {
                             if ('_' == name.charAt(j)) {
-                                sb.append(".");
+                                sb.append('.');
                                 i = j;
                             }
                         }
 
                         continue;
-                    } catch (NumberFormatException e) {
-                        // Ignore, it is not an indexed number
                     }
                 }
 
                 int j = i + 1;
                 if (j < length) {
                     if ('_' == name.charAt(j) && !quotesOpen) {
-                        sb.append(".");
-                        sb.append("\"");
+                        sb.append('.');
+                        sb.append('\"');
                         i = j;
                         quotesOpen = true;
                     } else if ('_' == name.charAt(j) && quotesOpen) {
-                        sb.append("\"");
+                        sb.append('\"');
                         // Ending
                         if (j + 1 < length) {
-                            sb.append(".");
+                            sb.append('.');
                         }
                         i = j;
                         quotesOpen = false;
                     } else {
-                        sb.append(".");
+                        sb.append('.');
                     }
                 } else {
-                    sb.append(".");
+                    sb.append('.');
                 }
                 beginSegment = j;
             } else {
@@ -217,6 +214,23 @@ public class StringUtil {
             }
         }
         return sb.toString();
+    }
+
+    public static boolean isNumeric(CharSequence digits) {
+        return isNumeric(digits, 0, digits.length());
+    }
+
+    public static boolean isNumeric(CharSequence digits, int start, int end) {
+        if (digits.length() == 0) {
+            return false;
+        }
+
+        for (int i = start; i < end; i++) {
+            if (!Character.isDigit(digits.charAt(i))) {
+                return false;
+            }
+        }
+        return true;
     }
 
     public static String skewer(String camelHumps) {
--- a/common/src/main/java/io/smallrye/config/common/utils/StringUtil.java
+++ b/common/src/main/java/io/smallrye/config/common/utils/StringUtil.java
@@ -88,14 +88,62 @@ public class StringUtil {
         return list.toArray(NO_STRINGS);
     }
 
+    private static boolean isAsciiLetterOrDigit(char c) {
+        return 'a' <= c && c <= 'z' ||
+                'A' <= c && c <= 'Z' ||
+                '0' <= c && c <= '9';
+    }
+
+    private static boolean isAsciiUpperCase(char c) {
+        return c >= 'A' && c <= 'Z';
+    }
+
+    private static char toAsciiLowerCase(char c) {
+        return isAsciiUpperCase(c) ? (char) (c + 32) : c;
+    }
+
+    public static boolean equalsIgnoreCaseReplacingNonAlphanumericByUnderscores(final String property,
+            CharSequence mappedProperty) {
+        int length = mappedProperty.length();
+        if (property.length() != mappedProperty.length()) {
+            // special-case/slow-path
+            if (property.length() != mappedProperty.length() + 1) {
+                return false;
+            }
+            if (mappedProperty.charAt(length - 1) == '"' &&
+                    property.charAt(length - 1) == '_' && property.charAt(length) == '_') {
+                length = mappedProperty.length() - 1;
+            } else {
+                return false;
+            }
+        }
+        for (int i = 0; i < length; i++) {
+            char ch = mappedProperty.charAt(i);
+            if (!isAsciiLetterOrDigit(ch)) {
+                if (property.charAt(i) != '_') {
+                    return false;
+                }
+                continue;
+            }
+            final char pCh = property.charAt(i);
+            // in theory property should be ascii too, but better play safe
+            if (pCh < 128) {
+                if (toAsciiLowerCase(pCh) != toAsciiLowerCase(ch)) {
+                    return false;
+                }
+            } else if (Character.toLowerCase(property.charAt(i)) != Character.toLowerCase(ch)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
     public static String replaceNonAlphanumericByUnderscores(final String name) {
         int length = name.length();
         StringBuilder sb = new StringBuilder(length);
         for (int i = 0; i < length; i++) {
             char c = name.charAt(i);
-            if ('a' <= c && c <= 'z' ||
-                    'A' <= c && c <= 'Z' ||
-                    '0' <= c && c <= '9') {
+            if (isAsciiLetterOrDigit(c)) {
                 sb.append(c);
             } else {
                 sb.append('_');
--- a/common/src/main/java/io/smallrye/config/common/utils/StringUtil.java
+++ b/common/src/main/java/io/smallrye/config/common/utils/StringUtil.java
@@ -107,7 +107,7 @@ public class StringUtil {
         int length = mappedProperty.length();
         if (property.length() != mappedProperty.length()) {
             // special-case/slow-path
-            if (property.length() != mappedProperty.length() + 1) {
+            if (length == 0 || property.length() != mappedProperty.length() + 1) {
                 return false;
             }
             if (mappedProperty.charAt(length - 1) == '"' &&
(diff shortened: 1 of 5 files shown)


franz1981:
Speedwise for checking contains of char sequence in batch a nice/better approach can be https://github.com/facebook/folly/blob/main/folly/String.cpp#L523-L613 which I've used on Netty as well, and works pretty well.

franz1981:
@radcortez fingers crossed it didn't break anything

franz1981:
And indeed It seems I have to fix something I have done to work with https://github.com/smallrye/smallrye-config/blob/6bd3669d79732288e5786d7659809c737f0dc316/common/src/main/java/io/smallrye/config/common/utils/StringUtil.java#L102

franz1981:
ok, I've fixed the case of https://github.com/smallrye/smallrye-config/pull/984#issuecomment-1694737105, but now the concern is related the speed of qualsIgnoreCaseReplacingNonAlphanumericByUnderscore` which is performing 
many comparisons to check if a letter is an Ascii letter or a digit.

This could harm performance in this case 

image

where, in order to avoid creating a new `String`, we skip performing `replaceNonAlphanumericByUnderscores` just once:  it means that we would performing several times the checks to replace non alphanumeric character.

It would be ideal to transform them in something more resilient to branch misprediction.
Running the case of https://github.com/quarkusio/quarkus/issues/35406 will help reveal if startup performance is impacted, and worth writing a microbenchmark to improve it.

implementation/src/main/java/io/smallrye/config/ConfigMappingProvider.java
```diff
@@ -67,14 +68,16 @@ final class ConfigMappingProvider implements Serializable {
         this.validateUnknown = builder.validateUnknown;
 
         final ArrayDeque<String> currentPath = new ArrayDeque<>();
+        final NameIterator nameIterator = NameIterator.empty();
         for (Map.Entry<String, List<Class<?>>> entry : roots.entrySet()) {
```
geoand, line 72: @franz1981 IIRC, this kind of iteration is not great. Would using `Maps.forEach` be better?

franz1981: I didn't focus on the collections here, tbh, but will take a look! Usually forEach is the best but when the code is hot, and not cold (at interpreted/C1) compilation level... lambda metafactory is terribly slow if the compilation Is not C2!
Indeed for interpreted/C1 compilation usually worth avoid lambdas and operations, in general: the less is better (which goes a bit against what I have said at https://github.com/smallrye/smallrye-config/pull/984#issuecomment-1694765001 sigh!)

geoand: How about if MethodHandle was being used?

common/src/main/java/io/smallrye/config/common/utils/StringUtil.java
```diff
@@ -88,14 +88,62 @@ public static String[] split(String text) {
         return list.toArray(NO_STRINGS);
     }
 
+    private static boolean isAsciiLetterOrDigit(char c) {
+        return 'a' <= c && c <= 'z' ||
+                'A' <= c && c <= 'Z' ||
```
metacosm, line 93:
```suggestion
               isAsciiUpperCase(c) ||
```
Or would a method call here make things too slow? Though, it might get inlined, no?

franz1981: I am in favour of reusing what we got; in theory the size (in term of bytecode size) of both methods is so small that will likely being inlined, if hot, unless we are unlucky and the depth of the call fall on the max inlining depth (which is 15 from JDK>=11 IIRC).
In theory these methods will be called for each character of each property names, which means they will likely (to be verified, really!) C2 compiled, and inlining, as intrinsics, won't kick in with previous compilation levels AFAK, so, If lucky, everything will work as expected.
To play safe, we should manually inline everything if this is supposed to happen at startup and we won't expect C2 compilation to kick-in.

TBH in a later iteration of this code, due to https://github.com/smallrye/smallrye-config/pull/984#issuecomment-1694765001 I will likely turn this code into some branchless variant, which can work on 4 chars at time, which will give 3 important benefits:
- most comparisons are on common prefixes, hence batching will grant to speedup moving to the different part of strings to compare
- most names have a mix of alphabetical and not chars, very unpredictable; a branchless approach will pay less in such cases 
- the batchy and branchless variant will amortize the cost of unlucky inlining decisions or lower compilation level than C2 by performing more ops in a row

implementation/src/main/java/io/smallrye/config/ConfigMappingProvider.java
franz1981: Mmmm same thing likely, these methods usually have some not negligible bootstrap cost...which is the reason why in cold cases, I prefer Atomic updaters vs VarHandle (which requires method handles)

franz1981: It's not easy but if you are curious (I am), I can distill some microbench on it, wdyt?

geoand: Not if you have better things to do :)

franz1981: We always have, but this one could be a good direction to improve out start-up times; it was sometime I wanted to write some "do/don't" guide to help Devs to write good optimized code for startup, which sadly need to have slightly different style in the "core" parts we care

geoand: 👌🏼

radcortez:
Thanks for the PR. Let me do a few runs to check the baseline, and I think I have some other ideas to improve.

radcortez:
Aside from the allocation issues, we are making unnecessary comparisons. I've just added a filter to the properties we need to match (only look at properties in the namespaces being mapped and not every env property), and this drastically cuts all the iterations and allocations.

Of course, your code will help when we need to do the matching, so I think the combination of both should be good.

I have some other ideas on how to improve this, but it will take a considerable amount of time.

implementation/src/main/java/io/smallrye/config/NameIterator.java
```diff
@@ -7,7 +7,7 @@
 /**
  * An iterator for property name strings.
  */
-public final class NameIterator {
+public final class NameIterator implements AutoCloseable {
```
dmlloyd, line 10: I don't really like the stylistic change here of making the NI be closeable; does the lifespan of the NI really leak out enough that the string needs to be cleared?

franz1981: It's not to clear the String, but to avoid re allocating the buffer underlying the StringBuilder used to extract the segment. Sadly I have performed a similar optimization on Jandex, for a similar reason..
Escape analysis kick in on C2 and still, even if it does it, it won't make the builder to be scalar replaced (because its underlying buffer can be enlarged, so...cannot be constant folded :/)
The reason why I made it closeable is to leverage the try with resource and reset the underlying state, to not mess up with subsequent reuse

franz1981: Anyway, the change I did to make getNextSegment able to use an external builder, could be enough to make it immutable again...but I have the strong feeling that other core path cannot leverage the same new API and would benefit from reusingnthe underlying builder...I am open to any other idea to improve it/make it right.

Pooling usually suck, API wise, due to introduced mutability

implementation/src/main/java/io/smallrye/config/ConfigMappingProvider.java
```diff
@@ -1065,6 +1069,32 @@ private static Set<String> additionalMappedProperties(final Set<String> mappedPr
             mappedProperties.remove(propertyName);
         }
 
+        Set<String> envRoots = new HashSet<>();
```
franz1981, line 1072: Roots are already contained in a set, do we really need another set here? Or we can just use a simple array, pre-sized? There are good reasons we could get any duplicate, after replacing non alphanumeric with underscores?

Saying that because both iteration and set constructions will be way cheaper with a simple array

implementation/src/main/java/io/smallrye/config/ConfigMappingProvider.java
```diff
@@ -1065,6 +1069,32 @@ private static Set<String> additionalMappedProperties(final Set<String> mappedPr
...
+        for (String root : roots) {
+            envRoots.add(replaceNonAlphanumericByUnderscores(root));
```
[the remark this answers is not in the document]
radcortez: Not, these are short. These represent the namespace of each mapping like `quarkus.config`, `quarkus. opentelemetry`, etc.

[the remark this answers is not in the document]
franz1981: I can always provide a version which reuse the SrringBuilder which will be allocated later anyway...and we will save creating each time for each invocation of that method, wdyt? 
Unless are so few we really don't care to allocate twice (builders + strings)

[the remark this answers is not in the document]
radcortez: I think it is worth converting them all and doing the check later. I was using that method first and it seems to me it got a bit worse because you need to match each root to each env, but I didn't microbenchmark it.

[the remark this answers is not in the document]
radcortez: > I can always provide a version which reuse the SrringBuilder which will be allocated later anyway...and we will save creating each time for each invocation of that method, wdyt?

Probably it makes sense to have a version that takes a builder so we can reuse it.

implementation/src/main/java/io/smallrye/config/ConfigMappingProvider.java
franz1981: It maybe still worth to pre size it at least (in the worst case will be underutilized, but won't suffer been enlarged?)

implementation/src/main/java/io/smallrye/config/ConfigMappingProvider.java
```diff
@@ -1065,6 +1069,32 @@ private static Set<String> additionalMappedProperties(final Set<String> mappedPr
...
+        }
+
+        // Ignore Env properties that don't belong to a root
+        Set<String> envPropertiesUnmapped = new HashSet<>();
```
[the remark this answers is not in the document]
radcortez: Ah, its because we remove the unmapped ones from the main set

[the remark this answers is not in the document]
franz1981: Yeah, but https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Set.html#removeAll(java.util.Collection) requires just a collection

radcortez:
I've added some additional improvements.

implementation/src/main/java/io/smallrye/config/NameIterator.java
radcortez: @dmlloyd are you ok with this?

common/src/main/java/io/smallrye/config/common/utils/StringUtil.java
```diff
@@ -171,9 +171,8 @@ public static String toLowerCaseAndDotted(final String name) {
 
                 // Do not convert to index if the first segment is a number
                 if (beginSegment > 0) {
-                    try {
-                        String segment = sb.substring(beginSegment, i);
-                        Integer.parseInt(segment);
+                    String segment = sb.substring(beginSegment, i);
+                    if (isNumeric(segment)) {
```
franz1981, line 175: No need to substring here, just expose a ranged check...and performing the parsing on it @radcortez

common/src/main/java/io/smallrye/config/common/utils/StringUtil.java
```diff
@@ -171,9 +171,8 @@ public static String toLowerCaseAndDotted(final String name) {
...
                         sb.replace(beginSegment - 1, beginSegment, "[").append("]");
```
franz1981, line 176: Lovely, not crucial, but you can spend a char, not need the string. For the replacement is the same? You can use maybe https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/StringBuilder.html#insert(int,char)

common/src/main/java/io/smallrye/config/common/utils/StringUtil.java
radcortez: Sure.

common/src/main/java/io/smallrye/config/common/utils/StringUtil.java
radcortez: There is also a `setChar`, https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/StringBuilder.html#setCharAt(int,char). I've used that instead.

radcortez (approved):

PR #989, opened 2023-09-01
Introduce `smallrye-config-api` module 

gastaldi:
I went ahead with @dmlloyd's suggestion in https://github.com/smallrye/smallrye-config/issues/981#issuecomment-1702760542 and created the `Config` interface. 

This avoids breaking existing code.

/cc @radcortez @dmlloyd

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.smallrye.config</groupId>
+        <artifactId>smallrye-config-parent</artifactId>
+        <version>3.3.4-SNAPSHOT</version>
+    </parent>
+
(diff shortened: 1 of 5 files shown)


implementation/src/main/java/io/smallrye/config/EnhancedConfig.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.Optional;
+import java.util.function.IntFunction;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.eclipse.microprofile.config.spi.Converter;
+
+import io.smallrye.common.annotation.Experimental;
+
+public interface EnhancedConfig extends Config, Serializable {
```
dmlloyd, line 17: I think maybe something like `ConfigView` is more appropriate. We don't want to present this as "an enhanced version of MP Config", but rather, "the basic SmallRye configuration view".

dmlloyd (dismissed):
Looks OK to me, as long as we're sticking with methods that were already public and not adding more API. I don't like adding more API - the more API you have, the more "locked in" you are to a design, compatibility-wise.

radcortez:
> Looks OK to me, as long as we're sticking with methods that were already public and not adding more API. I don't like adding more API - the more API you have, the more "locked in" you are to a design, compatibility-wise.

Ok, how about if we keep it empty, and we add the methods we need as we go?

Also, some time ago we discussed about having an `api` module for this sort of thing. Should we add it now? Other things should definitely go there too, but they need to live on a different namespace like `io.smallrye.config.api`. Unfortunately, the implementation root is becoming too cramped :(

PR #991, opened 2023-09-03
Support dynamic keys in a map's key (#986)

### LINKED ISSUE #986: Support for properties with dynamic names (e.g. Quarkus' "%dev,test") in Map-based configuration
I was told to report an issue here as the issue stems from a lack of feature support in smallrye-config:

Given this Quarkus application.yaml file:
```yaml
"%dev,test":
  quarkus:
    keycloak:
      devservices:
        roles:
          alice: sample_role
```
...I was expecting Quarkus (and underneath, it sounds like smallrye-config) to understand that the "quarkus" dictionary to belong to both the "dev" and "test" Quarkus profiles, but Quarkus failed to apply this configuration.

Instead, I was forced to break out the config into two duplicate copies:
```yaml
---
"%dev":
  quarkus:
    keycloak:
      devservices:
        roles:
          alice: sample_role
---
"%test":
  quarkus:
    keycloak:
      devservices:
        roles:
          alice: sample_role
```

~~This works as expected if I used a `properties` formatted file instead, so this is specific to how YAML is interpreted.~~ Update: this is not true. This does not work whenever the configuration terminates in a Map.

ejba:
Fixes #986


radcortez (changes requested):
Hi @ejba! Thank you very much for the PR. Highly appreciated.

While the original issue is reported against the YAML source, the problem is not in the YAML source implementation but in how the overall mechanism of multiple profiles is implemented.

The current approach works fine for non-dynamic lookups. The problem is when mapping a `Map`, the config system has to rely on the list of property names available to discover the dynamic key to place in the `Map`. If the list of property names does not contain the desired property in the active profile, the config system cannot discover it and the mapping will either fail or not map at all.

I believe a better fix may be provided in `ProfileConfigSourceInteceptor` by rewriting the multiple profile keys to be part of the active profile, which would solve the issue for any configuration source and not just YAML. There is a relocate rule to rewrite the multiple profile keys in `SmallRyeConfigBuilder`, which may also require some adjustments: https://github.com/smallrye/smallrye-config/blob/6bd3669d79732288e5786d7659809c737f0dc316/implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java#L235-L264

ejba:
Hi @radcortez! Thank you with your kind words. I should discussed the solution before opening a PR, apologies.

implementation/src/test/java/io/smallrye/config/ProfileConfigSourceInterceptorTest.java
```diff
@@ -220,6 +220,29 @@ void multipleProfiles() {
         assertEquals("5678", config.getRawValue("my.prop"));
     }
 
+    @Test
+    void multipleProfilesWithDynamicKey() {
+        SmallRyeConfig config = new SmallRyeConfigBuilder()
+                .withSources(config(SMALLRYE_CONFIG_PROFILE, "\"%common,prof\"", "config_ordinal", "1000"))
```
radcortez, line 226: We don't expect the profiles to be quoted or include the `%`. Profiles are usually set as:

```properties
smallrye.config.profile=prod
smallrye.config.profile=common,prod
smallrye.config.profile=one,two,three
```

The `%` is just to mark that the next name in the configuration property is the profile to be used, but it shouldn't be required when setting the profile. 

Did you find a situation where you need the profiles to be quoted?

ejba: No, I didn't find any practical example that requires quoting profile names. I will remove this part.

ejba: Removed :)

ejba: > I believe a better fix may be provided in ProfileConfigSourceInteceptor by rewriting the multiple profile keys to be part of the active profile ... 

Looking at your comment, maybe I missed something. Why does `ProfileConfigSourceInteceptor` need to be changed?

radcortez: Consider:

```java
@ConfigMapping
interface MyConfig {
 Map<String, String> foo();
}
```

```properties
foo.bar=baz
%dev.foo.bar=baz
%common,dev.foo.bar=baz
```

In this case, because the second segment of `foo.bar` is dynamic (the `Map` key), we cannot query it directly. We need to iterate the list of property names and find the ones in the `Map` namespace of `foo`.

With profiles, I guess we could search the profile namespace too (or multiple ones like in this case). On the other hand, when we added profiles, to not explode the list of property names, or even confuse users with the added prefix, we rewrite profile configuration names in their non-profile version (active):

https://github.com/smallrye/smallrye-config/blob/554783076711a2ccc89897075b237abfd19ef61c/implementation/src/main/java/io/smallrye/config/ProfileConfigSourceInterceptor.java#L60-L68

But in the case of multiple profile configuration names, this is not working. I believe that rewriting the multiple property names in their non-profiled format should do the trick (as long as the name found has any of the profiles active).

radcortez: Hey @ejba, how are you? Are you still planning to work on this? Thanks!

PR #992, opened 2023-09-07
Half-recursive getProperties

franz1981:
Startup measurement shows that the recursive nature of this method won't be handled nicely by the JIT interpreter: this change save recursion to happen unless more than a single property is found


radcortez:
Thanks! Apparently, CI is not happy :(

franz1981:
@radcortez some silly logic mistake of mine...which I cannot reproduce locally

franz1981:
I've left a comment on zulip which doesn't make this change so appealing :"/

dmlloyd (approved):
Seems OK to me.

radcortez (approved):

PR #993, opened 2023-09-07
Improve DefaultValuesConfigSource

radcortez:


dmlloyd:
What's the improvement?

radcortez:
We noticed some allocation activity on each lookup due to the search in the `KeyMap`, because of `getNextSegment` (which I'm algo looking to improve). 

In this case, we don't really need to pay the cost of storing all the defaults in the `KeyMap`, so the idea is that a plain name is stored in a simple `Map` and wildcard defaults are stored in the `KeyMap`.

PR #996, opened 2023-09-12
Improve EnvConfigSource

radcortez:
- Reduce allocations     
- Reduce lookups

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
+++ b/implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
@@ -18,22 +18,67 @@ package io.smallrye.config;
 import static io.smallrye.config.common.utils.ConfigSourceUtil.CONFIG_ORDINAL_KEY;
 import static io.smallrye.config.common.utils.StringUtil.replaceNonAlphanumericByUnderscores;
 import static java.security.AccessController.doPrivileged;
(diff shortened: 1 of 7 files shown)


implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
```diff
@@ -80,17 +142,30 @@ private static String getValue(final String name, final Map<String, String> prop
...
         if (value != null) {
             return Converters.INTEGER_CONVERTER.convert(value);
         }
         return ordinal;
     }
 
+    private static boolean isDottedFormat(final String propertyName) {
+        for (int i = 0; i < propertyName.length(); i++) {
+            char c = propertyName.charAt(i);
+            if (!('a' <= c && c <= 'z') && !('0' <= c && c <= '9') && c != '.') {
```
franz1981, line 162: How many times it happen to be false and how? In such case you can think to move the conditions or add new ones which can easily peel-off the common case eg  if are mostly uppercase, check if belong to that range first or anything less than `.`...

radcortez: The most usual case is looking for a dotted format property as `x.y.z`. This will let us know if we need to perform the env var conversion since we keep a cache of candidate dotted properties converted from env.

PR #1000, opened 2023-09-19
Reimplement handling of environment variables names to improve performance and allocation

radcortez:
- With a dedicated representation that can use equals and hashcode    
- With filtering and matching only what is required


implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
```diff
@@ -177,4 +142,141 @@ Object readResolve() {
...
+            for (int i = 0; i < length; i++) {
+                char c = name.charAt(i);
+                if (i == 0) {
+                    if (c == '%' || c == '_') {
+                        h = 31 * h + 31;
+                        continue;
+                    }
+                }
+
+                if (c == '.' || c == '_' || c == '-' || c == '"' || c == '*' || c == '[' || c == ']' || c == '/') {
```
franz1981, line 188: use a `switch` here, cause the JVM can optimize it very effectively

implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
```diff
@@ -177,4 +142,141 @@ Object readResolve() {
...
+                    continue;
+                }
+                h = 31 * h + Character.toLowerCase(c);
+            }
+            return h;
+        }
+
+        static boolean equals(final String name, final String other) {
+            if (Objects.equals(name, other)) {
```
franz1981, line 197: inline what `equals` will do here: the "type profile" (ie the concrete type seen in the `Object::equals` method) of this method risk to have seen different other concrete types already, and adding `String` too, will made the performed `String::equals` within to become megamorphic.
This is called type profile pollution: see https://wiki.openjdk.org/display/HotSpot/MethodData

Given that `String` override `equals` and `Objects::equals`  is calling `Object:equals` on the not-null parameter, there's a chance to slow down due to megamoprhic virtual calls there

implementation/src/main/java/io/smallrye/config/NameIterator.java
```diff
@@ -35,20 +35,6 @@ public NameIterator(final String name, final int pos) {
         with(name, pos);
     }
 
-    private StringBuilder acquireTmpBuilder() {
```
franz1981, line 38: Where's `tmp`? It doesn't seem to be deleted from `NameIterator`.

implementation/src/main/java/io/smallrye/config/NameIterator.java
```diff
@@ -292,7 +278,7 @@ public boolean nextSegmentEquals(String other, int offs, int len) {
     }
 
     public String getNextSegment() {
-        return getNextSegment(acquireTmpBuilder()).toString();
+        return getNextSegment(new StringBuilder()).toString();
```
franz1981, line 281: Are we fine that we allocate a new one each time?

implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
radcortez: Started with only and then kept adding :)

implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
radcortez: Actually we only need the `==` and we can skip `String.equals`, because the rest of the method can handle that and we avoid 2 passes.

implementation/src/main/java/io/smallrye/config/NameIterator.java
radcortez: Correct. Deleted.

implementation/src/main/java/io/smallrye/config/NameIterator.java
radcortez: Yes, we considerably reduced the use of the method with all of the other changes. The remaining uses are not in the hot path. I think we can get rid of it completely, but not a priority.

We still keep a `getNextSegment(StringBuilder b)` to use when required.

PR #1003, opened 2023-09-21
Improve normalization of profiled property

radcortez:


implementation/src/main/java/io/smallrye/config/ProfileConfigSourceInterceptor.java
```diff
@@ -82,13 +82,14 @@ public String[] getProfiles() {
...
     }
 
-    private String normalizeName(final String name) {
-        for (String profile : profiles) {
-            if (name.startsWith("%" + profile + ".")) {
-                return name.substring(profile.length() + 2);
+    public String normalizeName(final String name) {
+        if (name.length() > 0 && name.charAt(0) == '%') {
+            for (String profile : profiles) {
+                if (name.startsWith(profile + ".", 1)) {
```
franz1981, line 88: '.' :P

PR #1021, opened 2023-10-03
Config instance builder

### LINKED ISSUE #1001: API for using mapped configurations in a standalone way
Mapped configurations are very good in terms of brevity, clean API, and usability. However, they are not usable unless you're using SmallRye Config to read a configuration and create them indirectly in that way.

As a configuration user, I would like to have the ability to use mapped configurations to configure my application using my application's API, but have the usage of the actual configuration file be optional. This way I can configure my application using a file, *or* programmatically, all using the same API.

To allow this, we should have an API which allows configuration interfaces to be instantiated directly using a builder API. The API could look something like this:

```java
@ConfigMapping(prefix = "hello-world")
public interface HelloWorldConfig {
    @WithDefault("hello")
    String helloMessage();

    @WithDefault("5")
    int numberOfTimes();

    Optional<Path> outputPath();
}
```

```java
    public static void main(String[] args) {
        ConfigMappingBuilderFactory factory = ...;
        // manually create a config
        ConfigMappingBuilder<HelloWorldConfig> configBuilder = factory.builder(HelloWorldConfig.class);
        configBuilder.set(HelloWorldConfig::helloMessage, "good morning!");
        // convenience for optional properties
        configBuilder.setOptional(HelloWorldConfig::outputPath, Path.of("/tmp/good-morning.txt"));
        // primitives without boxing
        configBuilder.set(HelloWorldConfig::numberOfTimes, 10);
        // oops, I made a mistake, let's revert to default
        configBuilder.unset(HelloWorldConfig::numberOfTimes);
        HelloWorldConfig config = configBuilder.build();
        // now do something with config
    }
```

An API with this style is easier to use than source code generation based approaches. The builder factory would track generated classes for each configuration interface, which ideally could be reused by the existing config mapping implementation (though that implementation would not use the builder API itself per se). When running on Java 16 and later, these implementation classes should likely be implemented as records.

The generated record would be able to enforce constraints like non-nullity and validation in its canonical constructor (ideally using `smallrye-common-constraint` to validate nullity and `smallrye-config-validator` for the user validation rules as expected), ensuring that the consuming code will always receive a correct object. Optional properties would default to `Optional.empty()` and properties with explicit defaults would have those defaults set if no value is given.

Note that by waiting for the baseline JDK to move from 11 to 17, in addition to using records on the back end, we could possibly use [my backport of the JDK classfile API](https://github.com/dmlloyd/jdk-classfile-preview) to make class generation easiest.

dmlloyd:
Update: there is now a basic (but incomplete) implementation.

This is a first draft of the builder API for constructing configuration instances. No implementation is included yet; this is just for API review.

Fixes #1001.

### DIFF
--- a/documentation/mkdocs.yaml
+++ b/documentation/mkdocs.yaml
@@ -16,6 +16,7 @@ nav:
       - 'Indexed Properties': config/indexed-properties.md
       - 'ConfigValue': config/config-value.md
       - 'Mappings': config/mappings.md
+      - 'Instance Builder': config/instance-builder.md
       - 'Map Support': config/map-support.md
       - 'Secret Keys': config/secret-keys.md
       - 'Customizer': config/customizer.md
--- a/documentation/src/main/docs/config/instance-builder.md
+++ b/documentation/src/main/docs/config/instance-builder.md
@@ -1,0 +1,215 @@
+# Instance Builder
+
+With the `io.smallrye.config.ConfigInstanceBuilder` API, it is possible to create instances of a configuration interface 
+programmatically, without requiring a `SmallRyeConfig` or any configuration source. This is useful for testing, default 
+configurations, or any scenario where configuration values are known at build time.
+
+## Usage
+
+A configuration interface instance is created with `ConfigInstanceBuilder.forInterface()`:
+
+```java
+interface Server {
+    String host();
+
+    int port();
+}
+
+Server server = ConfigInstanceBuilder.forInterface(Server.class)
+        .with(Server::host, "localhost")
+        .with(Server::port, 8080)
+        .build();
+```
+
+The builder uses method references to identify which property to set, providing compile-time type safety without 
+string-based property names. The `with` method accepts a method reference to the configuration interface member and the 
+value to set.
+
+!!! info
+
+    The configuration interface does not require the `@ConfigMapping` annotation to work with the builder. Any valid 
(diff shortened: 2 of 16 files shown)


radcortez (dismissed):
Looks good to me.

radcortez:
@dmlloyd 

I've now started to look into this and pushed some initial implementation that includes:
-  Builder class generation
- Simple leaf types and primitives
- Simple optionals
- @WithDefault

Of course, there is still a lot missing, and even the pieces I did are only for the most basic cases, so I need to start looking into all the edge cases, but it is a start.

In terms of the API, I think we could start with something smaller and drop `withString` and `withDefaultFor` variants. We could add them later if required.

On another note, `with(F getter, int value)` and `with(F getter, long value)` (as well as their `Optional` counterparts), are ambiguous when we try to call them:

```java
interface Server {
    String host();
    int port();
}

Server server = forInterface(Server.class)
        .with(Server::host, "localhost")
        .with(Server::port, 8080)
        .build();
```

So, `.with(Server::port, 8080)` is reported as ambiguous by the compiler. Casting the value type doesn't seem to help, and it seems to be related to the Function type and method reference inferred type. Would you happen to have any idea?

dmlloyd:
Seems OK. Don't forget to JavaDoc the rest of the API.

PR #1022, opened 2023-10-04
Move logging & messages to private package

dmlloyd:
We can follow up with moving other non-API classes here as well.


radcortez (approved):

radcortez:
@dmlloyd we never moved other internal / private APIs to this package, mostly because we never had the time to do it.

Recently, I've started to work on a split of `@ConfigMapping` and MP `@ConfigProperties` implementations (they are mostly mashed up together), and thought it was a good time to move some stuff.

Now with the modularity work, we export the package because the only classes there are the logging ones, and we use them across other modules. The code I intended to move shouldn't be exported at all.

What do you recommend? Should we add an `internal` package to put these in? Or maybe keep `_private` for pieces that are only for the own module, and have `internal` as an export to other config modules?

PR #1026, opened 2023-10-11
Match any file extension from the acceptable list when loading profile locations

radcortez:


vsevel:
just to be sure
so this is solving `application.yml` with `application-base.yaml`
and not  `application.properties` with `application-base.yaml`

radcortez:
Correct.

Issue #1056, opened 2023-11-29
Handling of $$ in value


shawkins:
The current handling will replace $$ in a value with a single $ - this is due to the internal escaping behavior, which escapes escapes $ as $$.  However this behavior isn't documented and can be surprising to users.  Should https://github.com/smallrye/smallrye-config/blob/5a96de2849c87d144586df07dd4529e28d7e3b85/implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java#L101 also account for $$, or should this behavior be documented?

I don't mind submitting a patch if it's the former.

rtmolnar:
We are currently facing this issue, we receive our DB password through an AWS secret and there's a password rotation and every now and then we have a password that contains $$ which is replaced by $ which makes our connection to DB fails.

Everything is automated so we can't escape the $$.

There's no available configuration to change this behaviour since the $$ is replaced even before it's compiled in the `ExpressionConfigSourceInterceptor.java`:

`Expression expression = Expression.compile(escapeDollarIfExists(configValue.getValue()), LENIENT_SYNTAX, NO_TRIM,
                NO_SMART_BRACES, DOUBLE_COLON);`

Is there any hope this behaviour will change in the future?

I see this issue is quite old, should I open a new one?

PR #1059, opened 2023-12-04
fix: escapes $$ prior to escaping \$

shawkins:
This seems to be the most consistent handling for $ escaping - escape existing instances of $$ before replacing \$ with $$.  This is not of course completely compatible with the existing handling.

closes: #1056

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java
+++ b/implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java
@@ -98,14 +98,18 @@ public class ExpressionConfigSourceInterceptor implements ConfigSourceIntercepto
      * dollar.
      */
     private String escapeDollarIfExists(final String value) {
-        int index = value.indexOf("\\$");
+        return escapeIfExists(escapeIfExists(value, "$$", "$$$$"), "\\$", "$$");
+    }
+
+    private String escapeIfExists(final String value, final String toEscape, final String escaped) {
+        int index = value.indexOf(toEscape);
         if (index != -1) {
             int start = 0;
             StringBuilder builder = new StringBuilder();
             while (index != -1) {
-                builder.append(value, start, index).append("$$");
-                start = index + 2;
-                index = value.indexOf("\\$", start);
+                builder.append(value, start, index).append(escaped);
+                start = index + toEscape.length();
+                index = value.indexOf(toEscape, start);
             }
             builder.append(value.substring(start));
             return builder.toString();
--- a/implementation/src/test/java/io/smallrye/config/ExpressionConfigSourceInterceptorTest.java
+++ b/implementation/src/test/java/io/smallrye/config/ExpressionConfigSourceInterceptorTest.java
@@ -119,10 +119,10 @@ class ExpressionConfigSourceInterceptorTest {
 
     @Test
     void escape() {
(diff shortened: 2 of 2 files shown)


dmlloyd:
I think this trades one problem for another. If there's a specific parsing mode that we need for expression strings then it should be implemented directly in `smallrye-common-expression`. The optimization that is attempted by this existing code is problematic at best IMO.

shawkins:
@dmlloyd it seems like there isn't a great way to do this.  The comment https://github.com/smallrye/smallrye-config/blob/889fa9ffb91cd44d3d4b77e82b39c9cb81fc7750/implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java#L94 seems to make it clear that the more general handling of escape will not happen, so this workaround will remain in place for a while.  Any new feature flag for just the handling of $$ can't be added without affecting this workaround.

Are you thinking that something new like Flag.ESCAPES_COMPAT is needed to handle both \\$ and treat not automatically treat $$ as an escaped $?

Or @radcortez based upon your issue comment, are you wanting to further limit where the case and not do the $$ to $ substitution as long as Flag.MINI_EXPRS is not enabled and/or the next character is not {?

dmlloyd:
> @dmlloyd it seems like there isn't a great way to do this. The comment
> 
> https://github.com/smallrye/smallrye-config/blob/889fa9ffb91cd44d3d4b77e82b39c9cb81fc7750/implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java#L94
> 
> seems to make it clear that the more general handling of escape will not happen, so this workaround will remain in place for a while. Any new feature flag for just the handling of $$ can't be added without affecting this workaround.

The existing workaround is already concerning IMO. What if it hits `\$$`? or `\\$` or `\\\$` or `\\\$$`... etc. I don't think all of these cases are well-defined in the existing code.

> Are you thinking that something new like Flag.ESCAPES_COMPAT is needed to handle both \$ and treat not automatically treat $$ as an escaped $?

Exactly. But MP Config itself also does a very poor job of defining the behavior of `\\` in a general sense. Once you involve lists or any converter which uses `\\` in any other way, you end up with a big mess of confusing semantics.

dmlloyd:
Adding on to that, I think MP Config's treatment of `\$` is an error as well, precisely because of how it interferes with traditional expression syntax, converters, *and* the properties file spec. How many `\\` do you need? It's difficult to know for any given situation.

shawkins:
> The existing workaround is already concerning IMO. What if it hits \$$? or \\$ or \\\$ or \\\$$... etc. I don't think all of these cases are well-defined in the existing code.

I agree it's not well defined.

My interest here was primarily the $$ case to resolve the downstream issue https://github.com/keycloak/keycloak/issues/19831 - but obviously someone will hit the issue of additional or combined escapes eventually. 

Downstream I'm proposing that we simply document for now that $$ needs to be escaped as \\$\\$ https://github.com/keycloak/keycloak/pull/25218 - seems like it should work regardless of the resolution here.  If that's all that's worth doing at this time, I'm also fine just with closing this pr.

radcortez:
@shawkins, thank you for the PR.

We are now proposing a new alternative to fix the issue:  https://github.com/smallrye/smallrye-common/pull/344

shawkins:
> We are now proposing a new alternative to fix the issue: https://github.com/smallrye/smallrye-common/pull/344

Great, hopefully we'll be able to adopt that in Keycloak.

PR #1061, opened 2023-12-11
Generate all mappings binding code

radcortez:
Rewrote @ConfigMapping internal implementation to remove the runtime introspection and runtime functions. Required information can be extracted by build time runtimes and fed directly into the builder to reduce runtime operations.


radcortez:
@dmlloyd this is quite a big PR, but I think it would be hard to make it incrementally. The main goal was to remove all the runtime introspection code and mapping functions and make everything generated instead. I plan to provide some performance measurements, but early testing seems to indicate an improvement in allocation and CPU time.

dmlloyd (dismissed):

Issue #1068, opened 2023-12-19
Error message isn't helpful when config keystore password is missing


shawkins:
The error mentions the keystore name, but not the property.

PR #1069, opened 2023-12-19
fix: passing the whole property name to the error message

shawkins:
closes: #1068


dmlloyd (approved):
Makes sense to me...

radcortez (approved):

radcortez:
Thanks for the PR :)

PR #1074, opened 2023-12-26
Quality Gate back to Green

YassinHajaj:
I fixed the quality gate by adding the missing `@SuppressWarnings("squid:S4973")`.

Although I'm not sure why `Objects#equals` is not used here instead of `==`.

Two lines below it, I'm also not sure why `String#length == 0` is used instead of `String#isEmpty`

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/PropertyName.java
+++ b/implementation/src/main/java/io/smallrye/config/PropertyName.java
@@ -21,6 +21,7 @@ public class PropertyName {
         return equals(this.name, that.name) || equals(that.name, this.name);
     }
 
+    @SuppressWarnings("squid:S4973")
     static boolean equals(final String name, final String other) {
         //noinspection StringEquality
         if (name == other) {


radcortez:
Thank you for the PR :)

> Although I'm not sure why `Objects#equals` is not used here instead of `==`.

We want to avoid using `Objects#equals`, because it will use `String#equals` method, which does not really suit how we want to match the property names here, and it is just a waste of operations. Also, because property names tend to have a prefixed name, it is more efficient to match the name's backwords.
 
> In `EnvConfigSource`, I'm also not sure why `String#length == 0` is used instead of `String#isEmpty`

Yes, it could use `isEmpty`. Feel free to send a PR for that. Thanks.

radcortez (approved):

YassinHajaj:
Hi @radcortez ,

Thanks for the feedback. I understand a bit better the usage of `==`. Makes total sense.

I have written a PR for the `isEmpty` situation : https://github.com/smallrye/smallrye-config/pull/1079

PR #1079, opened 2024-01-08
Replaced String#length == 0 by String#isEmpty

YassinHajaj:


radcortez:
Thanks! Can you please fix the conflict?

YassinHajaj:
Hi @radcortez ,

Sure, it's fixed, and I also squashed the commits so to keep a single one and flatten the history (avoid a merge commit).

radcortez:
Great! Thanks!

radcortez (approved):

PR #1086, opened 2024-01-11
Adding the protected modifier back to getDefaultSources() in SmallRyeConfigBuilder

### LINKED ISSUE #1085: Add the protected modified back to getDefaultSources()
In 3.5.0 the `protected` modifier was removed from `SmallRyeConfigBuilder.getDefaultSources()` [(here](https://github.com/smallrye/smallrye-config/pull/1061/files#diff-daa60850df5b6b17463945eea3ca82d70cfb4ea741843899019cade8b32be1a3L184))

I am currently trying to update the `SmallRye` version that we are using to `3.5.1` but because `getDefaultSources` is no longer protected; our implementation is now causing failures as we are no longer able to Override and extend this method.

jakub-pomykala:
Fixes #1085 

Adding back the `protected` modified to `SmallRyeConfigBuilder.getDefaultSources()` to be able to override and extend the list

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
+++ b/implementation/src/main/java/io/smallrye/config/SmallRyeConfigBuilder.java
@@ -186,7 +186,7 @@ public class SmallRyeConfigBuilder implements ConfigBuilder {
         return this;
     }
 
-    List<ConfigSource> getDefaultSources() {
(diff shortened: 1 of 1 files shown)


radcortez:
Thanks! Sorry for the inconvenience.

radcortez (approved):

PR #1090, opened 2024-01-12
Do not skip a character if the env var character after a number is

Shortleft:
alphabetic

If the env var had the format FOO_BAR_1BAZ the second B was missed.

There was an error that occurred when we had an env var that combined both digits and alphabetical characters.
It turns out that if a digit is found, but its not an index in a map the character after the digit was missed.
This leads to 

`SRCFG00050: foo.bar.1 az in null does not map to any root`

when trying to start an application that relies on Smallrye (example is Quarkus 3.5.0 and later)


radcortez:
Thanks for the PR :)

PR #1097, opened 2024-01-24
Use map factory in getValues

### LINKED ISSUE #1096: Bug: SmallRyeConfig#getValues: mapFactory is not used
`io.smallrye.config.SmallRyeConfig#getValues(java.lang.String, org.eclipse.microprofile.config.spi.Converter<K>, org.eclipse.microprofile.config.spi.Converter<V>, java.util.function.IntFunction<java.util.Map<K,V>>)`

```java
public <K, V> Map<K, V> getValues(
    String name,
    Converter<K> keyConverter,
    Converter<V> valueConverter,
    IntFunction<Map<K, V>> mapFactory // 1️⃣ any Map could be created
){
  try {
    // 2️⃣ but in most cases caller will receive HashMap (with optional ClassCastException)
    return getValue(name, newMapConverter(keyConverter, valueConverter));
  } catch (NoSuchElementException e) {
    Map<String, String> mapKeys = getMapKeys(name);
    if (mapKeys.isEmpty()) {
        throw new NoSuchElementException(ConfigMessages.msg.propertyNotFound(name));
    }
    // 3️⃣ only this branch creates required type of Map
    Map<K, V> map = mapFactory.apply(mapKeys.size());
    for (Map.Entry<String, String> entry : mapKeys.entrySet()){
       map.put(convertValue(ConfigValue.builder().withName(entry.getKey()).withValue(entry.getKey()).build(),
                keyConverter), getValue(entry.getValue(), valueConverter));
    }
```

radcortez:
- Fixes #1096

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/Converters.java
+++ b/implementation/src/main/java/io/smallrye/config/Converters.java
@@ -322,9 +322,15 @@ public final class Converters {
      * @param <V> the type of the values
      * @return the new converter (not {@code null})
      */
+    @Deprecated
     public static <K, V> Converter<Map<K, V>> newMapConverter(Converter<? extends K> keyConverter,
             Converter<? extends V> valueConverter) {
-        return new MapConverter<>(keyConverter, valueConverter);
+        return newMapConverter(keyConverter, valueConverter, HashMap::new);
+    }
+
+    public static <K, V> Converter<Map<K, V>> newMapConverter(Converter<? extends K> keyConverter,
+            Converter<? extends V> valueConverter, IntFunction<Map<K, V>> mapFactory) {
+        return new MapConverter<>(keyConverter, valueConverter, mapFactory);
     }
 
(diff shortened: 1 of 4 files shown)


implementation/src/main/java/io/smallrye/config/Converters.java
```diff
@@ -322,9 +322,15 @@ public static <A, T> Converter<A> newArrayConverter(Converter<? extends T> itemC
      * @param <V> the type of the values
      * @return the new converter (not {@code null})
      */
+    @Deprecated
```
dmlloyd, line 325: Why deprecate this one?

radcortez: For the collection converter, we don't offer an API method without the factory, so I wanted to make the API consistent. I guess we could also add a method API that defaults the collection type like the map converter.

PR #1123, opened 2024-03-01
Include mappings nested super types metadata

radcortez:


implementation/src/main/java/io/smallrye/config/ConfigMappingInterface.java
```diff
@@ -138,9 +140,9 @@ String getClassInternalName() {
     }
 
     List<ConfigMappingInterface> getNested() {
-        ArrayList<ConfigMappingInterface> nested = new ArrayList<>();
+        Set<ConfigMappingInterface> nested = new HashSet<>();
```
dmlloyd, line 143: Should this be a `LinkedHashSet` to avoid indeterminate ordering?

implementation/src/main/java/io/smallrye/config/ConfigMappingLoader.java
```diff
@@ -35,12 +35,12 @@ public static List<ConfigMappingMetadata> getConfigMappingsMetadata(Class<?> typ
...
             }
         }
-        final ConfigMappingClass configMappingClass = ConfigMappingClass.getConfigurationClass(type);
+        ConfigMappingClass configMappingClass = ConfigMappingClass.getConfigurationClass(type);
         if (configMappingClass != null) {
             mappings.add(configMappingClass);
             mappings.addAll(getConfigMappingsMetadata(getConfigMapping(type).getInterfaceType()));
         }
-        return mappings;
+        return new ArrayList<>(mappings);
```
dmlloyd, line 43: Maybe this should be `List.copyOf(mappings)` instead so we don't return a mutable list.

PR #1125, opened 2024-03-05
Dramatically increase recursion cap

dmlloyd:
When using compatibility interceptors (`restart`) it is very easy to hit the very low cap of 20 interceptor actions. Raise it substantially.


dmlloyd:
No, this was actually the check doing its job. Closing for now.

PR #1127, opened 2024-03-06
Re-introduce a no-args constructor for LoggingConfigSourceInterceptor...

turing85:
… so that loading the service as described on https://smallrye.io/smallrye-config/Main/extensions/logging/ is still/again possible


radcortez (approved):

radcortez:
Thanks :)

PR #1128, opened 2024-03-07
Allow interceptors to interpose between "regular" sources and default sources

dmlloyd:
We split the interceptors and sources into two groups: those with negative priority and those with positive priority. The positive interceptors come before positive sources, which come before negative interceptors, which come before negative sources.


dmlloyd:
In draft until I can test it with Quarkus.

radcortez (approved):

dmlloyd:
Thanks!

PR #1133, opened 2024-03-13
Reduce allocations and batch appends on StringUtil::replaceNonAlphanumericByUnderscores

franz1981:
@radcortez  this is still a draft, but it:
- avoid allocating at all any `StringBuilder` if not necessary (usually it happens)
- it avoid advancing the `StringBuilder` 's length for each appended ascii/numeric char (including performing bound checks over and over), but try hard performing an optimized batch append

I have to write a JMH bench to verify if it is *much* better (as I expect, given that while still C1 compiled, usually, having less work overall deliver better performance) or instead is better to split the state machine into different loops (not on this PR yet).

I'll provide a JMH benchmark to show if it brings the benefits I expect

### DIFF
--- a/common/src/main/java/io/smallrye/config/common/utils/StringUtil.java
+++ b/common/src/main/java/io/smallrye/config/common/utils/StringUtil.java
@@ -94,21 +94,37 @@ public class StringUtil {
     }
 
     public static String replaceNonAlphanumericByUnderscores(final String name) {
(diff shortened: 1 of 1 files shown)


franz1981:
As usual, humans sucks at estimation :P
JMH reveal something interesting, when the code is interpreted; will update soon this draft

PR #1181, opened 2024-06-21
Generate equals and hashcode for @ConfigMapping implementations

### LINKED ISSUE #1067: Add the ability to generate equals and hashcode for @ConfigMapping
Currently, it's possible to generate toString as explained here: 
https://smallrye.io/smallrye-config/Main/config/mappings/#tostring

But there is nothing for equals and hashcode. 

The use case is for Quarkus extension config, without this, it's very ugly to deal with equals as shown in this PR:
https://github.com/quarkiverse/quarkus-quinoa/pull/594

When this is done we should update the Quarkus writing extension guide to mention this.

radcortez:
- Fixes #1067

Consider whether we should generate `equals` and `hashCode` by default. Right now, it always generates. Alternatively, we could mandate the `equals` and `hashcode` methods in the mapping interface to trigger the generation.

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/ConfigMappingGenerator.java
+++ b/implementation/src/main/java/io/smallrye/config/ConfigMappingGenerator.java
@@ -1,5 +1,6 @@
 package io.smallrye.config;
 
+import static org.objectweb.asm.Opcodes.AASTORE;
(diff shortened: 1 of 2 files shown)


ia3andy (approved):
Thanks @radcortez 😍

PR #1188, opened 2024-07-04
Update SmallRyeConfig.java

zhouzhou19950825:
Why not construct it directly, but construct an empty one first?


radcortez:
The `EMPTY` interceptor is the last in the chain to break the cycle. If you start with the current one, it will try to call the next in the chain which is `null` and fail.

PR #1194, opened 2024-07-04
Remove SecurityManager

radcortez:
Removed `SecuritySupport` and all `SecurityManager` API usage (`AccessController`, `PrivilegedAction`, `doPrivileged`). `SecurityManager` has been deprecated for removal since Java 17 (JEP 411).


dmlloyd:
Do we know that all downstream consumers (specifically WildFly) are OK with not having security manager support?

darranl (changes requested):
WildFly is going to need to retain security manager support as long as we are supporting Jakarta EE 10 - this also looks to be submitted to a branch which is producing micro releases, should a removal like this trigger a larger version bump?

radcortez:
> WildFly is going to need to retain security manager support as long as we are supporting Jakarta EE 10 - this also looks to be submitted to a branch which is producing micro releases, should a removal like this trigger a larger version bump?

We only set the release version on the actual release. Right now, it is just the version generated by Maven. We may consider moving to something like `999-SNAPSHOT` for development like Quarkus.

Ok, let's keep the SecurityManager for a little longer.

radcortez:
Moving forward with this, since Wildfly 42 is already out:
https://github.com/smallrye/smallrye-common/pull/397#issuecomment-3907695094

PR #1205, opened 2024-07-31
Keep a reference to the full interceptor chain to call restart

radcortez:


snazy (approved):
Pulled this change into the Quarkus 3.13.0 tag and it fixes the issue reported in https://github.com/quarkusio/quarkus/issues/42240

PR #1232, opened 2024-10-02
Secret type to represent secrets and integrate with the secrets handling

### LINKED ISSUE #1199: Provide Types to represent Secrets

radcortez:
- Fixes #1199

Adds a new wrapper type, `Secret<T>`, to represent a secret. When a `Secret` is part of a mapping, the path is added to the Secrets interceptor, to not include the secret in property names or access the secret without calling `SecretKeys.doUnlocked`.

This is initial work to support more advanced scenarios around secrets, that we should discuss.

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/SecretKeysConfigSourceInterceptor.java
+++ b/implementation/src/main/java/io/smallrye/config/SecretKeysConfigSourceInterceptor.java
@@ -1,46 +1,38 @@
 package io.smallrye.config;
 
 import java.io.Serial;
-import java.util.HashSet;
-import java.util.Iterator;
 import java.util.Set;
 
 import jakarta.annotation.Priority;
 
 import io.smallrye.config._private.ConfigMessages;
 
+/**
+ * Intercept the resolution of a configuration name and throw {@link java.lang.SecurityException} if the name is a
+ * Secret Key and the keys are locked.
+ * <p>
+ * To avoid having to recalculate the list of property names, the filter of secret keys is applied in
+ * {@code SmallRyeConfig.ConfigSources.PropertyNames}, so this interceptor does not implement
+ * {@link io.smallrye.config.ConfigSourceInterceptor#iterateNames(ConfigSourceInterceptorContext)}.
+ *
+ * @see SecretKeys
+ */
 @Priority(Priorities.LIBRARY + 100)
 public class SecretKeysConfigSourceInterceptor implements ConfigSourceInterceptor {
     @Serial
     private static final long serialVersionUID = 7291982039729980590L;
 
-    private final Set<String> secrets;
+    private final Set<PropertyName> secrets;
 
-    public SecretKeysConfigSourceInterceptor(final Set<String> secrets) {
+    public SecretKeysConfigSourceInterceptor(final Set<PropertyName> secrets) {
         this.secrets = secrets;
     }
 
     @Override
     public ConfigValue getValue(final ConfigSourceInterceptorContext context, final String name) {
-        if (SecretKeys.isLocked() && secrets.contains(name)) {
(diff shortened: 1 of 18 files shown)


dmlloyd:
The type is `Secret` but it's really a text password. Credentials however can be a variety of things - `Key` and its subclasses for example. Also we have a `Password` type and hierarchy in `wildfly-elytron`, which we use for security functions in Quarkus. This type extends `Key`, which is really the canonical supertype for secrets AFAICT.

What about supporting the existing types as secrets? Or at most maybe adding a wrapper e.g. `Secret<PrivateKey>` where we have a nested conversion process?

Also please look into how Elytron does password encoding because it could be relevant.

radcortez:
> The type is `Secret` but it's really a text password. Credentials however can be a variety of things - `Key` and its subclasses for example. Also we have a `Password` type and hierarchy in `wildfly-elytron`, which we use for security functions in Quarkus.

Yes, I was not too certain about the name, but the rationale was to keep it consistent with what we already have, `SecretKeys`, `SecretKeysHandler`.

> This type extends `Key`, which is really the canonical supertype for secrets AFAICT.

What do you mean? This only extends `Destroyable`, which, yes, I agree, is mainly for `Key`. The goal was to have a way to remove values from the mapping, because values are cached. Still, was only an idea.

> What about supporting the existing types as secrets? Or at most maybe adding a wrapper e.g. `Secret<PrivateKey>` where we have a nested conversion process?

We need the mapping generator to identify which elements are considered secrets. Either we provide a list of types for that case, or maybe we can make it work generically with a wrapper type. Let me see what I can do.

implementation/src/main/java/io/smallrye/config/SecretKeysConfigSourceInterceptor.java
```diff
@@ -12,15 +14,15 @@
...
 
     public SecretKeysConfigSourceInterceptor(final Set<String> secrets) {
-        this.secrets = secrets;
+        this.secrets = secrets.stream().map(PropertyName::new).collect(toSet());
     }
 
     @Override
     public ConfigValue getValue(final ConfigSourceInterceptorContext context, final String name) {
-        if (SecretKeys.isLocked() && secrets.contains(name)) {
+        if (SecretKeys.isLocked() && secrets.contains(new PropertyName(name))) {
```
gastaldi, line 25: Creating a new `PropertyTime` every time `getValue` is called may impact performance, perhaps consider iterating the `Set` and comparing the name instead?

radcortez:
I've moved `Secret` to a wrapper type `Secret<T>`, so this will allow using whatever type you want for the actual secret. The wrapper type only marks the property as a secret, which can be hidden from the list of names or forbid it to retrieve it directly with the API.

We need to figure out how to handle the nested converters piece. I'm wondering if we should support nested converter creation automatically.

implementation/src/main/java/io/smallrye/config/SecretKeysConfigSourceInterceptor.java
radcortez: It was to move forward with the implementation. I haven't profiled the code yet, but I am not too worried about it because when we are populating the config, the secrets are unlocked, but I'll have a look. Thanks!

radcortez:
//cc @sberyozkin @gsmet @yrodiere

radcortez:
> I wonder how that will play out with `Map<String, SomeConfigGroup>` properties? If someone only sets a secret in `SomeConfigGroup`, the entry disappears from the map?

The mapping process will keep access to all names (even if they are secrets). This will only affect the programmatic API. If required, users can use `SecretKeys.doUnlocked` to access all names and do lookups.

dmlloyd (dismissed):

PR #1238, opened 2024-10-07
Add @ConfigMapping validateUnknown to enable / disable validation of unknown configurations in the mapping path

radcortez:


dmlloyd:
What if two mappings have conflicting settings? I don't think this is the right place for this. Also we'll have to support if forever if it's on the public api like this. Even if we rewrite, which we may end up doing.

radcortez:
It is no different from the builder API.

The issue is that we have to disable the validation in Quarkus because the configuration may be split between build time and runtime. However, it is only disabled for the `quarkus` namespace. Some Quarkiverse extensions are now moving to mappings and using a different namespace. They must provide a config builder to disable the validation to their namespace, which is annoying, so I thought it would be easier to allow it in the annotation.

dmlloyd:
The config mapping interface does not correspond to a namespace (i.e. two mappings could overlap). Another difference is that this interface is supported as a Quarkus API whereas the programmatic API is not.

PR #1239, opened 2024-10-08
Add @ConfigMapping beanStyleGetter to enable / disable bean style getter names matching with configuration names

### LINKED ISSUE #1230: Are getters supported?
I think they are:

> If the member name is represented as a getter, the member name is taken from its property name equivalent, and then converted to its kebab-case format.
https://smallrye.io/smallrye-config/Main/config/mappings/

Yet, I can't get them to work, e.g.

```
java.util.NoSuchElementException: SRCFG00014: The config property graph.get-client-id is required but it could not be found in any config source
```

I'm using 3.9.1

radcortez:
- Fixes #1230

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/ConfigMapping.java
+++ b/implementation/src/main/java/io/smallrye/config/ConfigMapping.java
@@ -52,7 +52,21 @@ public @interface ConfigMapping {
      */
     NamingStrategy namingStrategy() default NamingStrategy.KEBAB_CASE;
 
-    enum NamingStrategy {
+    /**
+     * Match config mapping method names and configuration properties using bean-style getter names. This removes the
+     * prefixes <code>get</code> or <code>is</code> and decapitalizes the next character. By default, bean-style getter
+     * matching is <code>disabled</code>.
+     * <p>
+     * Config mapping declarations should prefer to use simple names that match one-to-one with their configuration
+     * names. Bean-style getter matching allows multiple method names to match the same configuration name. For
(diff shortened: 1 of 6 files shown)


implementation/src/main/java/io/smallrye/config/ConfigMapping.java
```diff
@@ -52,7 +52,21 @@
...
 
-    enum NamingStrategy {
+    /**
+     * Match config mapping method names and configuration properties using bean-style getter names. This removes the
+     * prefixes <code>get</code> or <code>is</code> and decapitalizes the next character. By default, bean-style getter
+     * matching is <code>disabled</code>.
+     * <p>
+     * Config mapping declarations should prefer to use simple names that match one-to-one with their configuration
+     * names. Bean-style getter matching allows multiple method names to match the same configuration name. For
+     * instance, <code>getFoo</code> and <code>isFoo</code> both match <code>foo</code>, which may not be intended.
```
dmlloyd, line 62: Do we know which one Kotlin uses? Maybe we should have this be an `enum` with each possiblity?

radcortez: From what I can tell is `get`, but you can override the name in Kotlin.

radcortez: @dmlloyd are you ok to merge as is? Or do you prefer something different?

dmlloyd: I had a thought to differentiate between these two cases, but I think it probably doesn't matter too much.

dmlloyd (approved):

PR #1248, opened 2024-11-11
fix #1246: honor ConfigProperties.UNCONFIGURED_PREFIX again

### LINKED ISSUE #1246: @ConfigProperties unconfigured prefix is not truncated
We use SmallRye Config in Apache TomEE and also run the Microprofile Config TCK there, after upgrading from `3.8.2` to `3.10.0` I noticed that the TCK does not pass anymore because [ConfigPropertiesTest](https://github.com/eclipse/microprofile-config/blob/3.0.3/tck/src/main/java/org/eclipse/microprofile/config/tck/ConfigPropertiesTest.java) fails.

```
11-Nov-2024 14:20:03.117 SEVERE [http-nio-33445-exec-3] org.apache.webbeans.portable.events.discovery.ErrorStack.logErrors Configuration validation failed:
        java.util.NoSuchElementException: SRCFG00014: The config property org.eclipse.microprofile.config.inject.configproperties.unconfiguredprefix.endpoint is required but it could not be found in any config source
        java.util.NoSuchElementException: SRCFG00014: The config property org.eclipse.microprofile.config.inject.configproperties.unconfiguredprefix.port is required but it could not be found in any config source
        java.util.NoSuchElementException: SRCFG00014: The config property org.eclipse.microprofile.config.inject.configproperties.unconfiguredprefix.host is required but it could not be found in any config source
```

Taking a look at the test itself one can find (note the omitted prefix in @ConfigProperties):
```
    @ConfigProperties
    @RequestScoped
    public static class BeanTwo {
        private String host;
        int port;
        String endpoint;
        /**
         * @return String return the host
         */
        public String getHost() {
            return host;
        }
    }
```

[The spec](https://download.eclipse.org/microprofile/microprofile-config-3.1/microprofile-config-spec-3.1.html#_aggregate_related_properties_into_a_cdi_bean) states:
> If the prefix is absent, the field x maps to the property name x.

> If the field name x needs to be different from the config property name y, use @ConfigProperty(name="y") to perform the transformation. If the prefix is present, the field x maps to the configuration property \<prefix\>.y, otherwise y.

So really the properties `endpoint`, `port` and `host` should be resolved instead of `org.eclipse.microprofile.config.inject.configproperties.unconfiguredprefix.(endpoint|port|host)` and I think aborting the entire deployment in the CDI extension in this case is wrong

Small side note: I noticed you don't even execute this test: https://github.com/smallrye/smallrye-config/blob/main/testsuite/tck/pom.xml#L57-L62

jungm:
- Fixes #1246

### DIFF
--- a/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
+++ b/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
@@ -99,10 +99,15 @@ public class ConfigExtension implements Extension {
             // We are going to veto, because it may be a managed bean, and we will use a configurator bean
             processAnnotatedType.veto();
 
+            // Unconfigured is represented as an empty String in SmallRye Config
+            String prefix = annotatedType.getAnnotation(ConfigProperties.class).prefix();
+            if (ConfigProperties.UNCONFIGURED_PREFIX.equals(prefix)) {
+                prefix = "";
+            }
+
             // Each config class is both in SmallRyeConfig and managed by a configurator bean.
             // CDI requires more beans for injection points due to binding prefix.
-            ConfigClass properties = ConfigClass.configClass(annotatedType.getJavaClass(),
-                    annotatedType.getAnnotation(ConfigProperties.class).prefix());
+            ConfigClass properties = ConfigClass.configClass(annotatedType.getJavaClass(), prefix);
             configProperties.add(properties);
             configPropertiesBeans.add(properties);
         }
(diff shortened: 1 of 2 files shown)


radcortez:
Thanks for the PR :)

jungm:
When do you think will the next release happen? For now I downgraded SmallRye config in TomEE but that is of course not an optimal solution 🙂

radcortez:
I should be able to do one this week.

PR #1249, opened 2024-11-11
Do not use lambdas in Converters

radcortez:


dmlloyd:
Why? Is there some kind of problem case?

dmlloyd:
I would expect a similar number of allocations in either case. The only difference would be the one time initialization of the lambda infrastructure, but that gets paid by someone at some point no matter what so there's no train to avoid it here IMO.

PR #1270, opened 2024-12-15
check duplicate keys

### LINKED ISSUE #1269: [enhancement] check duplicate keys
EDIT(2024-12-16): After advices from PR https://github.com/smallrye/smallrye-config/pull/1270, the goal has been limited to just warn for duplicate keys on properties config source.

Sometimes happens that the same config name is found multiple time in the same config source.

For instance : 

```properties
quarkus.bannner.enabled=false
quarkus.bannner.enabled=true
```

Or 

```yaml
quarkus:
  banner:
    enabled: false
  banner:
    enabled: true
```

In our quarkus projects (whose configuration is based on smallrye-config) this is just silently ignored.

I know that IDE helps avoiding it and developers should be disciplined when writing configuration, but in our large teams sometimes this happened.

But for a [Developer Joy](https://quarkus.io/developer-joy/) approach, it could be helpful to have an Out Of The Box solution.

For instance we could enable duplicate keys check with a system property like ? 

```
smallrye.config.forbid-duplicate-keys
```

Note : main goal is to use this feature to che configurations at dev or in the CI

fugerit79:
- Fixes #1269

Edit : 

Only for properties config source : duplicate keys found in te same config source will be logged as warning.

NOTE: this is just a proposal of implementation.

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/ConfigValueConfigSource.java
+++ b/implementation/src/main/java/io/smallrye/config/ConfigValueConfigSource.java
@@ -416,7 +416,8 @@ public interface ConfigValueConfigSource extends ConfigSource {
                         .withLineNumber(lr.lineNumber)
                         .build());
                 if (oldConfigValue != null) {
-                    ConfigLogging.log.duplicateValue( oldConfigValue.getName(), oldConfigValue.getConfigSourceName(), oldConfigValue.getValue() );
+                    ConfigLogging.log.duplicateValue(oldConfigValue.getName(), oldConfigValue.getConfigSourceName(),
+                            oldConfigValue.getValue());
                 }
             }
         }
--- a/implementation/src/main/java/io/smallrye/config/ConfigValueConfigSource.java
+++ b/implementation/src/main/java/io/smallrye/config/ConfigValueConfigSource.java
@@ -3,6 +3,7 @@ package io.smallrye.config;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
+import java.lang.invoke.MethodHandles;
 import java.util.AbstractCollection;
 import java.util.AbstractMap;
 import java.util.AbstractSet;
@@ -14,7 +15,9 @@ import java.util.Map;
 import java.util.Set;
 
 import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.jboss.logging.Logger;
 
+import io.smallrye.config._private.ConfigLogging;
 import io.smallrye.config._private.ConfigMessages;
 
 /**
@@ -340,6 +343,9 @@ public interface ConfigValueConfigSource extends ConfigSource {
         private final String configSourceName;
         private final int configSourceOrdinal;
 
+        private static ConfigLogging log = Logger.getMessageLogger(MethodHandles.lookup(), ConfigLogging.class,
+                "io.smallrye.config");
+
         public ConfigValueProperties(final String configSourceName, final int configSourceOrdinal) {
             this.configSourceName = configSourceName;
             this.configSourceOrdinal = configSourceOrdinal;
@@ -401,7 +407,7 @@ public interface ConfigValueConfigSource extends ConfigSource {
                 }
                 String key = loadConvert(lr.lineBuf, 0, keyLen, convtBuf);
                 String value = loadConvert(lr.lineBuf, valueStart, limit - valueStart, convtBuf);
-                put(key, ConfigValue.builder()
+                ConfigValue oldConfigValue = put(key, ConfigValue.builder()
                         .withName(key)
                         .withValue(value)
                         .withRawValue(value)
@@ -409,6 +415,10 @@ public interface ConfigValueConfigSource extends ConfigSource {
                         .withConfigSourceOrdinal(configSourceOrdinal)
                         .withLineNumber(lr.lineNumber)
                         .build());
+                if (oldConfigValue != null) {
+                    log.warnv("duplicate keys found for : {0}, source name : {1}", oldConfigValue.getName(),
+                            oldConfigValue.getConfigSourceName());
+                }
             }
         }
 
--- a/implementation/src/main/java/io/smallrye/config/ConfigValueConfigSource.java
+++ b/implementation/src/main/java/io/smallrye/config/ConfigValueConfigSource.java
@@ -416,8 +416,7 @@ public interface ConfigValueConfigSource extends ConfigSource {
                         .withLineNumber(lr.lineNumber)
                         .build());
                 if (oldConfigValue != null) {
-                    log.warnv("duplicate keys found for : {0}, source name : {1}", oldConfigValue.getName(),
-                            oldConfigValue.getConfigSourceName());
+                    ConfigLogging.log.duplicateValue( oldConfigValue.getName(), oldConfigValue.getConfigSourceName(), oldConfigValue.getValue() );
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


documentation/src/main/docs/config/getting-started.md
```diff
@@ -40,6 +40,26 @@ greeting.message=hello
...
+```
+
+Is set to *true* 
+
+A ConfigValidationException will be raised if a duplicate key is found in the same config source.
+
+i.e. 
+
+```
+-Dsmallrye.config.forbid-duplicate-keys=true
```
radcortez, line 58: I think we don't need an extra configuration to warn / fail. A simple warning should be enough.

implementation/src/main/java/io/smallrye/config/ConfigValueConfigSource.java
```diff
@@ -401,14 +405,15 @@ private void load0(LineReader lr) throws IOException {
...
-                put(key, ConfigValue.builder()
-                        .withName(key)
-                        .withValue(value)
-                        .withRawValue(value)
-                        .withConfigSourceName(configSourceName)
-                        .withConfigSourceOrdinal(configSourceOrdinal)
-                        .withLineNumber(lr.lineNumber)
-                        .build());
+                this.checkPutResult.accept(
+                        put(key, ConfigValue.builder()
```
radcortez, line 409: `.put` returns the previous object associated with the key. It can easily be used to check and warn about the old and new values.

implementation/src/main/java/io/smallrye/config/ConfigValuePropertiesUtils.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Setting system property :
+ * smallrye.config.forbid-duplicate-keys=true (default false)
+ * A ConfigValidationException will be raised if a duplicate key is found in the same config source.
+ *
+ * https://github.com/smallrye/smallrye-config/issues/1269
+ *
+ * NOTE: For backward compatibility, the default behaviour it to avoid duplicate keys check.
+ */
+public class ConfigValuePropertiesUtils {
```
radcortez, line 21: If we don't use a configuration to warm / fail and use `.put` semantics, this shouldn't be required.

radcortez (changes requested):
Thank you for the PR. I've left a few comments.

documentation/src/main/docs/config/getting-started.md
fugerit79: Thanks for the feedback, this is yeas on property configuration, I have to think at something for YAML configuration as I was relying on *LoadOptions.isAllowDuplicateKeys()*

radcortez: It ends up using a `Map` in the source, so it can use the same `.put` logic:
https://github.com/smallrye/smallrye-config/blob/1ee672abe1080f3ce212d6f9451f1a3b5a71585f/sources/yaml/src/main/java/io/smallrye/config/source/yaml/YamlConfigSource.java#L66-L98

implementation/src/main/java/io/smallrye/config/ConfigValuePropertiesUtils.java
fugerit79: Thanks for the feedback, i changed the warn/fail approach and made it just a warning.

This will only work for properties config sources.

For yaml ones, i didn't find a way to to just warn duplicate keys when reading the yaml config source with SnakeYAML.

This is going to be fine anyway for the projects I'm working on, but at start I wanted to make it a bit more general.

radcortez: Did you check?
https://github.com/smallrye/smallrye-config/blob/1ee672abe1080f3ce212d6f9451f1a3b5a71585f/sources/yaml/src/main/java/io/smallrye/config/source/yaml/YamlConfigSource.java#L66-L98

All the yaml entries are added to a `Map<String, String>`, so instead of `putAll` you could iterate the entries, call `put` and check for the returned object.

fugerit79: Hello thanks for the advice @radcortez 

Maybe I'm missing something, but as far as I understand Yaml.loadAll() already returns a single Map where duplicate keys are skipped.

image

I think SnakeYAML only allow failing in case of [duplicate](https://www.javadoc.io/doc/org.yaml/snakeyaml/latest/org/yaml/snakeyaml/LoaderOptions.html#isAllowDuplicateKeys()).

**Maybe right now we limit this PR to properties config source?**

NOTE: In ConfigValueProperties the Properties.load() logic is replicated and the check is easy.

About YAML I tried to propose something to SnakeYAML project. : [Option to log duplicate keys](https://bitbucket.org/snakeyaml/snakeyaml/issues/1101/option-to-log-duplicate-keys).

radcortez: Ah, so the YAML code before was to fail and not to remove duplicates? Sorry, I thought it did not remove the duplicates either.

fugerit79: Yes, as far I understand SnakeYAML either throws an Exception or removes the duplicate key : 

Here is the [code](https://bitbucket.org/snakeyaml/snakeyaml/src/d19d063fb5e2afb4ff2bb333ec92c6b16acc9479/src/main/java/org/yaml/snakeyaml/constructor/SafeConstructor.java#lines-112) : 

```java
        Integer prevIndex = keys.put(key, i);
        if (prevIndex != null) {
          if (!isAllowDuplicateKeys()) {
            throw new DuplicateKeyException(node.getStartMark(), key,
                tuple.getKeyNode().getStartMark());
          }
          toRemove.add(prevIndex);
        }
```

@radcortez Do you think is acceptable to just log the duplicate keys in property config source? (I think I answered to all other comments too, the current PR content is much simplier than the original one)?

implementation/src/main/java/io/smallrye/config/ConfigValueConfigSource.java
```diff
@@ -401,14 +407,18 @@ private void load0(LineReader lr) throws IOException {
...
+                ConfigValue oldConfigValue = put(key, ConfigValue.builder()
                         .withName(key)
                         .withValue(value)
                         .withRawValue(value)
                         .withConfigSourceName(configSourceName)
                         .withConfigSourceOrdinal(configSourceOrdinal)
                         .withLineNumber(lr.lineNumber)
                         .build());
+                if (oldConfigValue != null) {
+                    log.warnv("duplicate keys found for : {0}, source name : {1}", oldConfigValue.getName(),
```
radcortez, line 419: Can you please add a new method in https://github.com/smallrye/smallrye-config/blob/029c88ea3f957417b04231748b6d2431caed2959/implementation/src/main/java/io/smallrye/config/_private/ConfigLogging.java#L13 for the log message?

fugerit79: Ok, Is it ok this way? 

https://github.com/fugerit-org/smallrye-config/commit/c562fb9b71c9ca2f93c7226ab73adc3f277429e9

radcortez: Yes :)

fugerit79:
> There is an issue with the code style, can you please do a local build to fix it? I've tried it myself, but apparently I can't push changes to your fork.

Yes, sorry, my bad. After the last commit, I forgot to run:

```shell
mvn -f implementation net.revelc.code.formatter:formatter-maven-plugin:2.22.0:format
mvn formatter:validate verify
```

Now the local check completed without alerts. (tried with java 11, 17 and 21).

I had to double check it.

radcortez (approved):

radcortez:
Great! Thank you for the PR and your patience, really appreciated it :)

radcortez (approved):

PR #1272, opened 2024-12-18
fix: avoid out of range error with long profile names

jeloba:
Long profile names, such as `test-with-native-agent` may cause string length overruns on short properties.

This introduces a check to avoid that.


radcortez (changes requested):
Do you mind creating a test for this? Thank you!

jeloba:
@radcortez added a test. Before the change it would throw an out of bound exception, not it passes.

radcortez (approved):

radcortez:
Excellent. Thank you!

PR #1274, opened 2024-12-20
Replace streams with for loop in customizers handling

geoand:
This isn't going to make any difference in real life, but it does at least make stacktraces look
a little better to look at
(when an exception is thrown or in a flamegraph)

Theoretically this change does change the semantics slightly as the customizers array is now sorted, but these customizers are private, so it shouldn't matter


radcortez (approved):

radcortez:
Thanks!

geoand:
🙏

PR #1287, opened 2025-01-10
Move to Java 17

radcortez:


DirkMahler:
Is it a good idea to change the JDK requirement within a minor release, see https://github.com/jQAssistant/jqassistant/pull/769?

radcortez:
I guess it is a debatable choice, which we discussed and ultimately agreed not to burn major versions because most of our main consumers already require Java 17. We did announce the move on our ML, but I realize that is not a good media these days (but it is our official one). We probably have to consider deprecating the ML and moving to GH Discussions.

I'm sorry for the inconvenience. I'm happy to backport any fix you require from `3.11.x` to `3.10.x`.

DirkMahler:
Thanks for the clarification! There's currently no need for a fix/backport. We use dependabot to stay up-to-date for security related fixes, so we need to check if something comes up in the near future that might hit us.

But I hope that we can finally drop Java 11  as well, but this also depends a bit on our users (enterprises and public sector...)

andrewazores:
@radcortez are security updates to `3.10.x` still in the cards, now ~1 year after this change? I only just realized that my project is affected, but in my case I can't make the decision to drop JDK 11 support yet.

PR #1292, opened 2025-01-16
Use StringBuilder instead of String concat for startup code

geoand:
Unfortunately the JDK's indyfied String concatenation has a slight performance penalty at startup, so for code we know will always be run at startup, let's not pay that unnecessary price

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceLoader.java
+++ b/implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceLoader.java
@@ -264,7 +264,13 @@ public abstract class AbstractLocationConfigSourceLoader {
 
     private static URI addProfileName(final URI uri, final String profile, final String fileExtension) {
         if ("jar".equals(uri.getScheme())) {
-            return URI.create("jar:" + addProfileName(URI.create(uri.getRawSchemeSpecificPart()), profile, fileExtension));
+            URI rec = addProfileName(URI.create(uri.getRawSchemeSpecificPart()), profile, fileExtension);
+            String recString = rec.toString();
+            String finalString = new StringBuilder("jar:".length() + recString.length())
+                    .append("jar:")
+                    .append(recString)
+                    .toString();
+            return URI.create(finalString);
         }
 
         final String fileName = uri.getPath();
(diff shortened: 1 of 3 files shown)


implementation/src/main/java/io/smallrye/config/AbstractLocationConfigSourceLoader.java
```diff
@@ -264,7 +264,13 @@ private boolean validExtension(final String resourceName) {
 
     private static URI addProfileName(final URI uri, final String profile, final String fileExtension) {
         if ("jar".equals(uri.getScheme())) {
-            return URI.create("jar:" + addProfileName(URI.create(uri.getRawSchemeSpecificPart()), profile, fileExtension));
```
dmlloyd, line 267: It is `+`, not concatenation, which incurs the cost; using `String.concat()` is simpler and cleaner here. But, even better would be to pass the option to `javac` to disable indy string concatenation and leave the code as-is IMO. That way we won't miss any.

dmlloyd: The option is `-XDstringConcat=inline`.

geoand: @dmlloyd Thanks, I was looking for that `javac` flag! Were did you find that?

dmlloyd: It is used by openjdk when building the `java.base` module.

geoand: Cool, thanks!

So passing configuring Maven on `smallrye-config` to use the flag seems like a good option, but the real question is do we want to do that for Quarkus itself.... My feeling is that we should only do that for the core runtime module, WDYT?

dmlloyd: Yeah we should be careful where we apply it. And I think there will be some cases where `String.concat` will be faster too.

geoand: I think applying it to this moduke and `quarkus-core` is safe, but I'm fine using `.concat` instead

dmlloyd: Honestly I don't know why javac would ever produce anything other than `.concat` for simple `a + b` (just single concat) situations.

geoand: Yeah, sometimes the mundane approach makes a lot more sense than the fancy / complicated one...

radcortez:
Closed in favor of https://github.com/smallrye/smallrye-config/pull/1293

PR #1294, opened 2025-01-21
Branch-less alphanumeric underscore's replacement

franz1981:
This is following the same idea of https://github.com/quarkusio/quarkus/pull/45546 re using lookup tables.
If the src `String`s are rarely latin only beyond US_ASCII i.e. [128, 255] this can be further simplified by having a 128-sized lookup table.

I didn't yet looked at its performance - and, as suggested at https://github.com/mkouba/qute-benchmarks/pull/1 we should have a proper micro-benchmark which account to stress the branch-predictor too (or not - should be configurable) - unless the non-alphanumeric case is supposed to be very predictable in reality as well

### DIFF
--- a/common/src/main/java/io/smallrye/config/common/utils/StringUtil.java
+++ b/common/src/main/java/io/smallrye/config/common/utils/StringUtil.java
@@ -54,6 +54,13 @@ public class StringUtil {
         return (char) (((int) NON_ALPHANUMERIC_UNDERSCORE_REPLACEMENTS[c & 0xFF]) & 0xFF);
     }
 
+    private static byte rawReplacementOf(char c) {
+        if (c > 255) {
+            return '_';
+        }
+        return NON_ALPHANUMERIC_UNDERSCORE_REPLACEMENTS[c & 0xFF];
+    }
+
     private static final String[] NO_STRINGS = new String[0];
 
     private static final Pattern ITEM_PATTERN = Pattern.compile("(,+)|([^\\\\,]+)|\\\\(.)");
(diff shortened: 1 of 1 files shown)


franz1981:
I've created this benchmark at which gives these numbers https://github.com/franz1981/java-puzzles/commit/8fb833c9ce7ce51de1a651b09c8253448425bc60

```
Benchmark                                                     (finalDoubleQuoteProbability)  (nonAlphanumericProbability)  (samples)  (size)  Mode  Cnt    Score   Error  Units
ReplaceWithUnderscores.replaceWithUnderscoresSwitch                                      10                            10      10000      33  avgt   10  231.371 ± 9.276  ns/op
ReplaceWithUnderscores.replaceWithUnderscoresTableArray                                  10                            10      10000      33  avgt   10   32.910 ± 2.189  ns/op
ReplaceWithUnderscores.replaceWithUnderscoresTableSb                                     10                            10      10000      33  avgt   10  102.949 ± 4.079  ns/op
```
I didn't accounted yet for the code duplication, but since native image is not wow to handle `StringBuilder`s and `String`s - moving to a "ligher" abstractions should pay off (@galderz can verify it :P) - for JIT I would like to test when only C1 is running or with "cold" code as well.

franz1981:
Sadly if I got right what C1MaxInlineSize does - C1 doesn't have inlining superpowers, hence  reusing code could be relatively costly if the callee is not trivially inlinable, which seems this case

radcortez (approved):

PR #1295, opened 2025-01-23
Makes EnvName mutable to avoid allocating it for lookups

franz1981:
While working on https://github.com/smallrye/smallrye-config/pull/1294 I've discovered two interesting things:
- `EnvName::hashCode ` (which is compute intensive) it's performed twice due to a mix of  if `get == null`,  `put`: here I've enabled hashCode caching, despite it makes `EnvName` a bit fatter i.e. 24 bytes vs 16 :"(
- `EnvName` is created multiple times on the fly, to perform matching - without ever storing it (!). And most of the times these are misses: I've than allow it to be mutable to avoid keeping on allocating them

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
+++ b/implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
@@ -62,6 +62,7 @@ public class EnvConfigSource extends AbstractConfigSource {
     public static final int ORDINAL = 300;
 
     private final EnvVars envVars;
+    private EnvName reusableEnvName;
 
     protected EnvConfigSource() {
         this(ORDINAL);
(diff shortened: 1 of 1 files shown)


franz1981:
A similar effect with a smaller increment could be obtained via int[] cache of hash codes; we can still perform hash code computation directly on the String and using the int[] as a sort of bloom filter using a single hash function:
- if the check is positive we need to perform the allocation and go down the slow path of checking the map
- if the check is negative no need to further check

That depends by the collision of keys vs the given hash function, so the effect is potential reduction of allocations

implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
```diff
@@ -103,7 +105,18 @@ public String getValue(final String propertyName) {
...
+            envName = new EnvName(propertyName);
+            this.reusableEnvName = envName;
+        } else {
+            envName.setName(propertyName);
+        }
+        try {
+            return envVars.getEnv().containsKey(envName);
+        } finally {
+            envName.reset();
+        }
```
radcortez, line 119: This can cause issues if not thread-safe. While we use a simple `HashMap`, we never modify it after creation.

PR #1301, opened 2025-02-01
Use config ordinal to determine which config style to use when injecting a list, between comma-separated and indexed

radcortez:


dmlloyd:
Can you explain this heuristic? There is no doc change and it's not clear what's going on here.

radcortez:
Added a description. Sorry for not being transparent.

PR #1314, opened 2025-02-13
Compute indexed properties when building the list of PropertyNames

### LINKED ISSUE #1305: 3.10+ - Yaml lists are out of order
When upgrading Quarkus, I noticed that lists defined the YAML way in the `application.yaml` are read in a different order than defined in the config. This issue only began in `3.10.0` (and still exists in `3.11.2`). It seems to only be an issue when the list has 11 or more items. And is not an issue with comma-defined lists.

### Confg
```yaml
yaml-list:
  - thingA
  - thingB
  - thingC
  - thingD
  - thingE
  - thingF
  - thingG
  - thingH
  - thingI
  - thingJ
  - thingK
comma-list: thingA,thingB,thingC,thingD,thingE,thingF,thingG,thingH,thingI,thingJ,thingK
```

### Result
```
yaml-list  = [thingA, thingK, thingB, thingC, thingD, thingE, thingF, thingG, thingH, thingI, thingJ]
comma-list = [thingA, thingB, thingC, thingD, thingE, thingF, thingG, thingH, thingI, thingJ, thingK]
```

### Reproducer
1. Clone `https://github.com/mikethecalamity/quarkus-test`
2. Run `gradlew -p apps service3:test`
3. Note the test using the comma-separated list succeeds and the test using the yaml list fails

radcortez:
- Fixes #1305

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
+++ b/implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
@@ -43,7 +43,9 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.IntFunction;
 import java.util.function.Supplier;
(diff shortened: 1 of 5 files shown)


implementation/src/main/java/io/smallrye/config/SmallRyeConfig.java
```diff
@@ -228,40 +230,13 @@ private <T, C extends Collection<T>> C getIndexedValues(List<String> indexedProp
...
-                            && StringUtil.isNumeric(propertyName, indexStart + 1, indexEnd)) {
-                        indexedProperties.add(propertyName);
-                    }
-                }
-            }
-        }
-        Collections.sort(indexedProperties);
-        return indexedProperties;
+        Map<Integer, String> indexedProperties = configSources.getPropertyNames().indexed().get(property);
+        return indexedProperties == null ? Collections.emptyList() : indexedProperties.values().stream().toList();
```
dmlloyd, line 234: With autoboxing, I doubt that a map is going to be better than a sorted list with all of the redundant object creation. Also, I think your method of converting to a stream is suspect. Are you *certain* this will give you your items in *index* order? Your test shows values in lexicographical order so it might be working accidentally.

Maybe the original problem is the actual sort though; if you're sorting numbers lexicographically, it's likely going to be a problem. It should not be too hard to define a numerically-sorting comparator that is efficient (find first non-zero character in each, compare lengths, then compare `Character.digit(x)` for each digit from most significant to least).

radcortez: The underlying map is backed by a `TreeMap`, and the index is being extracted, converted, and used as a key in the map (basically to discard invalid indexes).

The indexed structure is `Map<String, Map<Integer, String>>`, where the first key is the unindexed name, and then the `TreeMap` contains the indexes, with the target property. This was to avoid having to reconstruct the key.

PR #1330, opened 2025-02-24
Configure all container image names in a one place

Chu3laMan:
# Centralize Container Image Configuration

## Description
This PR addresses issue #45746, which aims to make container image configuration more maintainable by centralizing all image references in one place. Currently, image references are scattered across pom.xml files and Java source files, making it difficult to change image sources (e.g., switching from Docker Hub to Quay.io).

## Implementation
- Added a `ContainerImagesConfig` interface using `@ConfigMapping` to provide type-safe access to container image names
- Explicitly specified all registries for consistency (docker.io vs quay.io)
- Centralized all container image configurations in application.properties
- Updated relevant code to use the configuration interface instead of hardcoded image names


dmlloyd:
This is very much the wrong place for this change. SmallRye Config is a general purpose configuration library, which includes configuration for things that have absolutely nothing to do with containers. Config mappings for random things like this is definitely not a part of this project's scope.

PR #1359, opened 2025-06-05
Reduce allocations of Environment variables matching

radcortez:


implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
```diff
@@ -253,16 +260,46 @@ static Optional<List<Integer>> indexOfDashes(final String property, final int of
...
+        ResizableIntArray(int initialSize) {
+            this.array = new int[initialSize];
         }
 
-        return Optional.of(dashesPosition);
+        int[] get() {
+            return array;
+        }
+
+        void ensureCapacity(int capacity) {
```
franz1981, line 289: enlarge only if needed, not always!

franz1981 (changes requested):

implementation/src/main/java/io/smallrye/config/EnvConfigSource.java
radcortez: I don't reuse the array between calls, and when `ensureCapacity` is called, it always requires increasing the size.

radcortez: Yes, the most common case scenario is to exit with `0` or `-1`. Also, I need two methods calls to check the prefix and the rest of the name. I could use two separate arrays instances and juggle these around, but I think it is not worth it.

PR #1378, opened 2025-07-11
Sort matching properties with Env to prioritize specific names first

radcortez:


comrt:
@radcortez is there any release planned soon?

radcortez:
I was investigating other potential issues, so I had been holding this off. I'll do one before the end of the week.

PR #1383, opened 2025-07-22
Support Parameterized Types Converters

radcortez:
The goal here is to support Parameterized type Converters in the programmatic API due to our new planned type for secrets `Secret` in https://github.com/smallrye/smallrye-config/pull/1232.

While we could follow the same approach we already use for collections and optionals, it might be more interesting to support any kind of Parameterized type. To be able to represent a parameterized type, we add a `TypeConverter` that must be instantiated with the type to be resolved for the converter, such as:

```java
requireConverter(new TypeConverter<Optional<String>>() {})
Converter<List<String>> listConverter = requireConverter(new TypeConverter<>() {});
```

We will retain all the current Converter APIs and add specific methods to support `TypeConverter`. Internally, we should be able to start using this new API without breaking `getValue*` methods.

I've pushed a small prototype so we can discuss the direction and what it could look like before investing too much time in this.

### DIFF
--- a/implementation/src/test/java/io/smallrye/config/TypeConvertersTest.java
+++ b/implementation/src/test/java/io/smallrye/config/TypeConvertersTest.java
@@ -1,0 +1,151 @@
+package io.smallrye.config;
+
+import static io.smallrye.config.Converters.newCollectionConverter;
+import static org.junit.jupiter.api.Assertions.assertIterableEquals;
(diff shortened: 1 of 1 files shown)


radcortez:
The goal here was to support `Secret<T>` (#1232) as a declaration in the mapping with an automatic `Converter`.

PR #1412, opened 2025-09-29
Return a pre-registered Converter if one exists when calling `getImplicitConverter`

radcortez:
- For https://github.com/quarkusio/quarkus/issues/50206


gsmet (approved):
Awesome, thanks!

PR #1443, opened 2025-12-17
Fix #1390: FileSystem source joins multiline file content into a single line

### LINKED ISSUE #1390: `FileSystem` source joins multiline file content into a single line
The FileSystemConfigSource reads file content line by line and then concatenates them,
which results in multiline content being converted into a single line.

This behavior can lead to issues when the configuration value is intended to span multiple lines.

For example, a file named `foo.key` with the following content:
```
line1
line2
```
will be read as `line 1 line 2` instead of preserving the newline characters.

Is there a specific reason for this behavior? It seems that concatenating the lines at this point may not be the intended functionality, as it can cause unexpected behavior with multiline configuration values.

https://github.com/smallrye/smallrye-config/blob/main/sources/file-system/src/main/java/io/smallrye/config/source/file/FileSystemConfigSource.java#L100

Any clarification on this would be appreciated. Thank you.

usv240:
Fixes #1390

### Description
The `FileSystemConfigSource` was reading file content using `Files.lines(file)` and joining the lines with `Collectors.joining()`, which concatenates lines without any delimiter. This caused multiline files (like keys or certificates) to be read as a single line.

### Changes
- Updated `FileSystemConfigSource.java` to use `Collectors.joining("\n")` to preserve newlines.
- Added a regression test `testMultiline` in `FileSystemConfigSourceTest.java` to verify the behavior.

### Verification
Ran `mvn test` in `sources/file-system` and verified that all tests pass.
Fixes #1390

### Description
The `FileSystemConfigSource` was reading file content using `Files.lines(file)` and joining the lines with `Collectors.joining()`, which concatenates lines without any delimiter. This caused multiline files (like keys or certificates) to be read as a single line.

### Changes
- Updated `FileSystemConfigSource.java` to use `Collectors.joining("\n")` to preserve newlines.
- Added a regression test `testMultiline` in `FileSystemConfigSourceTest.java` to verify the behavior.

### Verification
Ran `mvn test` in `sources/file-system` and verified that all tests pass.

### DIFF
--- a/sources/file-system/src/main/java/io/smallrye/config/source/file/FileSystemConfigSource.java
+++ b/sources/file-system/src/main/java/io/smallrye/config/source/file/FileSystemConfigSource.java
@@ -20,6 +20,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.Serial;
 import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


radcortez:
I'm sorry, but we are not accepting PR's done by AI. Please, create a new PR without AI.

PR #1453, opened 2026-01-09
Avoid using a lambda when releasing config

gsmet:
Screenshot From 2026-01-09 17-37-58

Screenshot From 2026-01-09 17-39-12

I can't show how it looks after the change as it disappears :).


radcortez (approved):

radcortez:
Actually, we should prefer to use https://github.com/smallrye/smallrye-config/blob/b7b4f507bc12775e38dfb578d54289e626a3c29c/implementation/src/main/java/io/smallrye/config/SmallRyeConfigProviderResolver.java#L149-L155, but that is only available in `SmallRyeConfigProviderResolver` and not `ConfigProviderResolver`.

I'll have a look.

PR #1454, opened 2026-01-09
Add get methods to retrieve SmallRyeConfig directly

### LINKED ISSUE #1388: Alternative to `ConfigProvider.getConfig`
Typically, we use `ConfigProvider.getConfig` to get an instance of the current configuration, especially since there is no other way to do so outside a CDI environment.

We have also discussed on many occasions that `ConfigProvider.getConfig` is misleading and sometimes causes hidden issues, because if there is no config instance associated with the current Classloader, it will automatically create one and return it, which may not be the expected config instance, plus the config instance may be missing components, since it uses defaults plus discovery to create the instance.

I propose to introduce four new methods into `SmallRyeConfig`:

```java
public static SmallRyeConfig get();
public static SmallRyeConfig getOrCreate();
public static SmallRyeConfig get(ClassLoader);
public static SmallRyeConfig getOrCreate(ClassLoader);
```

The `getOrCreate` methods are equivalent to the methods in `ConfigProvider`, while `get` methods will throw an exception if there is no config associated with the current Classloader. The `get` methods become the recommended way to get an instance of `SmallRyeConfig`.

There are also a couple of other benefits:
- We reduce our exposure to MP Config, which we know is going to end up in Jakarta one way or another, with a possible namespace change
- We drop the call to `.unwrap(SmallRyeConfig.class)` or cast to `SmallRyeConfig`
- Users will become more aware of `SmallRyeConfig` capabilities, since that would be the type they get instead of the MP `Config` type.

radcortez:
- Fixes #1388

### DIFF
--- a/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
+++ b/cdi/src/main/java/io/smallrye/config/inject/ConfigExtension.java
@@ -51,10 +51,10 @@ import jakarta.enterprise.inject.spi.configurator.AnnotatedTypeConfigurator;
 import jakarta.enterprise.util.Nonbinding;
 import jakarta.inject.Provider;
 
-import org.eclipse.microprofile.config.ConfigProvider;
 import org.eclipse.microprofile.config.inject.ConfigProperties;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 
+import io.smallrye.config.Config;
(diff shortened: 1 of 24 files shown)


dmlloyd:
So, it's mainly to have an interface instead of the concrete class? This seems like a pretty slim benefit but I guess I don't see any major problem with it.

PR #1455, opened 2026-01-13
Move Converters lambdas to static classes

radcortez:
See discussion: https://quarkusio.zulipchat.com/#narrow/channel/187038-dev/topic/AOT.20cache.20-.20SmallRye.20Config.20converters/with/567537198


gastaldi (dismissed):
LGTM, but maybe they should be final to match the existing converters in this class

PR #1521, opened 2026-05-12
Fix #1390: preserve line breaks and whitespace in FileSystemConfigSource

### LINKED ISSUE #1390: `FileSystem` source joins multiline file content into a single line

phateffect:
Currently, FileSystemConfigSource uses `Files.lines().collect(joining())` which strips all line endings from the source file.

This PR switches to `Files.readString()` to ensure the content is returned exactly as-is, preserving multiline values and trailing spaces.

### DIFF
--- a/sources/file-system/src/main/java/io/smallrye/config/source/file/FileSystemConfigSource.java
+++ b/sources/file-system/src/main/java/io/smallrye/config/source/file/FileSystemConfigSource.java
@@ -97,8 +97,8 @@ public class FileSystemConfigSource extends MapBackedConfigSource {
     }
 
     private static String readContent(Path file) {
-        try (Stream<String> stream = Files.lines(file)) {
-            return stream.collect(Collectors.joining());
+        try {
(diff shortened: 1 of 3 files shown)


radcortez (changes requested):
Code does not compile.

PR #1527, opened 2026-05-28
Stabilize hashCode of ConfigClass

gsmet:
Class doesn't implement its own hashCode so its hashCode is not stable at all.
During the reproducibility work made by @reaver585, we figured out it would be better to have a stable hashCode.

We don't change the equals() as we want the class loader to be taken into account in the equals().

Related to a discussion in:
https://github.com/quarkusio/quarkus/pull/54371#discussion_r3312830183

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/ConfigMappings.java
+++ b/implementation/src/main/java/io/smallrye/config/ConfigMappings.java
@@ -192,7 +192,10 @@ public final class ConfigMappings {
 
         @Override
         public int hashCode() {
-            return Objects.hash(type, prefix);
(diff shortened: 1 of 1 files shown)


gsmet:
@radcortez an easy one for when you're back and rested :).

Ladicek (dismissed):
Comment would be nice, but LGTM :-)

gsmet:
Ah I actually hesitated as if having value or not. I was on the verge of adding one, let me do it.

gsmet:
And done!

Ladicek (approved):

manovotn (approved):

radcortez (approved):

radcortez:
Thanks!

PR #1534, opened 2026-06-17
Reproduce issue with `@WithUnnamedKey` config group considered present even if unconfigured

yrodiere:
Creating for discussion following https://quarkusio.zulipchat.com/#narrow/channel/187038-dev/topic/.40WithUnnamedKey.20and.20keySet/with/603369236

Running the tests and reading the "ISSUE" comments is probably is the easiest way to grasp what is going on.

Test results:

image

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/ConfigMappingContext.java
+++ b/implementation/src/main/java/io/smallrye/config/ConfigMappingContext.java
@@ -45,6 +45,8 @@ public final class ConfigMappingContext {
 
     private NamingStrategy namingStrategy = NamingStrategy.KEBAB_CASE;
     private BeanStyleGetters beanStyleGetters = BeanStyleGetters.DISABLED;
+    private boolean resolvedNonDefault;
+
     private final StringBuilder nameBuilder = new StringBuilder();
     private final Set<String> usedProperties = new HashSet<>();
     private final List<Problem> problems = new ArrayList<>();
@@ -289,7 +291,7 @@ public final class ConfigMappingContext {
                 final Class<? extends Converter<K>> keyConvertWith,
                 final String unnamedKey,
                 final Iterable<String> keys) {
-            return map(keyRawType, keyConvertWith, unnamedKey, keys, null);
+            return map(keyRawType, keyConvertWith, unnamedKey, keys, null, null);
         }
 
         public <K, V> ObjectCreator<T> map(
@@ -297,7 +299,8 @@ public final class ConfigMappingContext {
                 final Class<? extends Converter<K>> keyConvertWith,
                 final String unnamedKey,
                 final Iterable<String> keys,
-                final Supplier<V> defaultValue) {
+                final Supplier<V> defaultValue,
+                final Supplier<V> unnamedDefaultSupplier) {
             Converter<K> keyConverter = keyConvertWith == null ? config.requireConverter(keyRawType)
                     : getConverterInstance(keyConvertWith);
             List<Consumer<Function<String, Object>>> nestedCreators = new ArrayList<>();
@@ -315,11 +318,26 @@ public final class ConfigMappingContext {
                         Map<K, V> map = defaultInstance != null ? new MapWithDefault<>(defaultInstance) : new HashMap<>();
 
                         if (unnamedKey != null) {
+                            V unnamedDefault = defaultInstance;
+                            if (unnamedDefault == null && unnamedDefaultSupplier != null) {
+                                int problemsBefore = problems.size();
+                                int length = nameBuilder.length();
+                                nameBuilder.append(".*");
+                                unnamedDefault = constructGroup(unnamedDefaultSupplier);
+                                nameBuilder.setLength(length);
+                                if (problems.size() > problemsBefore) {
+                                    problems.subList(problemsBefore, problems.size()).clear();
+                                    unnamedDefault = null;
+                                }
+                            }
+                            final V unnamedDefaultInstance = unnamedDefault;
                             nestedCreators.add(new Consumer<>() {
                                 @Override
                                 public void accept(Function<String, Object> get) {
+                                    resolvedNonDefault = false;
                                     V value = (V) get.apply(path);
-                                    if (value != null) {
+                                    if (value != null
+                                            && (resolvedNonDefault || !value.equals(unnamedDefaultInstance))) {
                                         map.put(unnamedKey.isEmpty() ? null : keyConverter.convert(unnamedKey), value);
                                     }
                                 }
@@ -590,7 +608,14 @@ public final class ConfigMappingContext {
                 final String propertyName,
                 final Converter<V> valueConverter) {
             context.usedProperties.add(propertyName);
-            return context.config.getValue(propertyName, valueConverter);
+            ConfigValue configValue = context.config.getConfigValue(propertyName);
+            if (configValue.getValue() != null && !configValue.isDefault()) {
+                context.resolvedNonDefault = true;
+            }
+            if (Converters.CONFIG_VALUE_CONVERTER.equals(valueConverter)) {
+                return (V) configValue.noProblems();
+            }
+            return context.config.convertValue(configValue, valueConverter);
         }
 
         public ObjectCreator<T> value(
@@ -644,7 +669,11 @@ public final class ConfigMappingContext {
                 final String propertyName,
                 final Converter<V> valueConverter) {
             context.usedProperties.add(propertyName);
-            return context.config.getOptionalValue(propertyName, valueConverter);
+            ConfigValue configValue = context.config.getConfigValue(propertyName);
+            if (configValue.getValue() != null && !configValue.isDefault()) {
+                context.resolvedNonDefault = true;
+            }
+            return context.config.convertValue(configValue, newOptionalConverter(valueConverter));
         }
 
         public <V> ObjectCreator<T> optionalValue(
--- a/implementation/src/main/java/io/smallrye/config/ConfigMappingGenerator.java
+++ b/implementation/src/main/java/io/smallrye/config/ConfigMappingGenerator.java
@@ -594,6 +594,11 @@ public class ConfigMappingGenerator {
                 } else {
                     ctor.visitInsn(ACONST_NULL);
                 }
+                if (mapProperty.hasKeyUnnamed()) {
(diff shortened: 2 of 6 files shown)


radcortez:
This behavior is expected with the current implementation (not saying it's right or wrong). Consider:

```java
@ConfigMapping(prefix = "mapping")
interfaceMapping {

  Group group();

  @WithDefaults
  Map<String, Group> map();

  interface Group {
    @WithDefault("value")
    String value();
  }
```

What the implementation does is to generate properties with the values and add them to our `DefaultValuesConfigSource` as:

```properties
mapping.group.value=value
mapping.*.value=value
```

The whole idea was to also be able to use the programmatic API to query for the name, so that the values you get from the mapping and the programmatic API are consistent.

When you add `@WithUnnamedKey`, this also generates a `mapping.value=value`, which gets picked by the `Map`, and returns the `unnamed` key in the `Map` keys. This pretty much works like the old pattern we used, when we had the default as a separate group and the map for the named keys. We never had a way to distinguish whether a group is only coming from defaults or had user input (doesn't matter if it is a group in a `Map`, a standalone group, or an optional group.

Should the `unnamed` key appear in the `Map` keys if it's only defaults?

Because we never required that distinction, everything we populate comes from sources, including defaults, so for the engine, it is just as if someone set `mapping.value=value`, and now you get the `unnamed` key. The default for all keys simply matches the `mapping.*` pattern and returns a single element on every `Map#get`. Of course, there is no way to include the keys in the `Map` keys, so this may seem inconsistent with how `unnamed` is handled.

We may require a concept that doesn't exist today: distinguishing defaults-only groups from user-configured ones. I'm looking into this from a standalone group perspective, because that was the motivation for us to implement `@WithDefaults` and `@WithUnnnamedKey` and to store everything in the `Map` to simplify the mapping pattern. While on a `Map`, you could express that using keys (or the absence of keys), how would we express it in a standalone group?

yrodiere:
> The whole idea was to also be able to use the programmatic API to query for the name, so that the values you get from the mapping and the programmatic API are consistent.

Makes sense, but as discussed on Zulip, you only get consistency for the unnamed key -- for other "named" keys, if they are absent from the user's configuration, you won't get any defaults through the programmatic API.

That being said, in my case I truly don't need programmatic retrieval to be consistent in any way when it comes to unnamed key -- in fact, if I currently relied on it, we'd have bugs because we don't get consistency for other "named" keys.

> Should the unnamed key appear in the Map keys if it's only defaults?

IMO, no. If it's not configured, it should not appear in the map keys.

> Of course, there is no way to include the keys in the `Map` keys, so this may seem inconsistent with how `unnamed` is handled.

Right, exactly.

> While on a `Map`, you could express that using keys (or the absence of keys), how would we express it in a standalone group?

So, first, I want to stress I truly don't need this. I just need the map keys to behave :grin: 

But, if you want a hollistic approach and not a surgical fix... `Optional<MyGroup>` seems like an obvious answer to me, which would do the job. Or, if you want to allow defaults, some custom interface, e.g. `Defaultable<MyGroup>` with `defaultable.get()` returning the value or the defaults, and `defaultable.isPresent()`/`isEmtpy()` allowing to query it. Maybe `defaultable.toOptional()`, though that can easily be a default method.

And whatever the solution, IMO it's *also* fine to not have defaults available for that optional group when retrieving values programmatically. It's a very reasonsable limitation tied to a very specific feature.

radcortez:
> > The whole idea was to also be able to use the programmatic API to query for the name, so that the values you get from the mapping and the programmatic API are consistent.
> 
> Makes sense, but as discussed on Zulip, you only get consistency for the unnamed key -- for other "named" keys, if they are absent from the user's configuration, you won't get any defaults through the programmatic API.
> 
> That being said, in my case I truly don't need programmatic retrieval to be consistent in any way when it comes to unnamed key -- in fact, if I currently relied on it, we'd have bugs because we don't get consistency for other "named" keys.

Actually, this does work for the programmatic API :)

> So, first, I want to stress I truly don't need this. I just need the map keys to behave 😁

Ok, let me see what I can do.

radcortez:
Check if this works for you.

The proposal is just to compare the unnamed group with a default group and omit the key in that case (all mappings include equals). Now, if the user explicitly sets a key, but it is a default, the unnamed key won't be included. Is this acceptable?

yrodiere (commented):
That looks perfect! Thank you.

yrodiere:
Er... maybe I talked too fast. So now, if the user configured the unnamed key by setting a property to its default value, we will omit it from the map?

radcortez:
> Er... maybe I talked too fast. So now, if the user configured the unnamed key by setting a property to its default value, we will omit it from the map?

Yes. But does that matter?

yrodiere:
> > Er... maybe I talked too fast. So now, if the user configured the unnamed key by setting a property to its default value, we will omit it from the map?
> 
> Yes. But does that matter?

Kind of, yes. The idea is to use configuration to extract a list of datasources (or persistence units, or...) that the user explicitly configured, because we know those need to be created for sure (unless they're explicitly disabled).

Before this patch, I was in trouble because the map could contain `<default>` even though the user didn't configure the default datasource.
After this patch, I'm still in trouble because the map might not contain `<default>` even though the user set, say, `quarkus.datasource.jdbc=true` (and only that).

In practice this is just one datapoint, and I'll try to also inspect injection points and other things. But still, it does mean we'll have a hole in our detection code...

radcortez:
I did it this way because it was an easy fix that didn't require messing with the names or the defaults, which I wanted to avoid.

Let me think about this again.

radcortez:
Ok, give it another try.

implementation/src/main/java/io/smallrye/config/ConfigMappingContext.java
```diff
@@ -303,11 +306,26 @@ public Object apply(final String path) {
...
+                                if (problems.size() > problemsBefore) {
+                                    problems.subList(problemsBefore, problems.size()).clear();
+                                    unnamedDefault = null;
+                                }
+                            }
+                            final V unnamedDefaultInstance = unnamedDefault;
                             nestedCreators.add(new Consumer<>() {
                                 @Override
                                 public void accept(Function<String, Object> get) {
+                                    resolvedNonDefault = false;
```
yrodiere, line 337: I don't want to be annoying, I really don't, so I'll start by saying I really appreciate you looking into this :)

But I need to mention we could have a map inside a map, too. So unless those would use a different context, there's a risk that some nested map processing would reset `resolvedNonDefault` -- e.g. if that nested map is empty, but a non-map property is configured.

We have that in Hibernate Search: 

https://github.com/quarkusio/quarkus/blob/bb15beacc7c797e1efede3e874190994d8a6bda7/extensions/hibernate-search-orm-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/orm/elasticsearch/runtime/HibernateSearchElasticsearchBuildTimeConfig.java#L31-L37

https://github.com/quarkusio/quarkus/blob/1c43f902c95818cb5620cdb5788a1e1927e60ecb/extensions/hibernate-search-orm-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/orm/elasticsearch/runtime/HibernateSearchElasticsearchBuildTimeConfigPersistenceUnit.java#L18-L25

radcortez: > I don't want to be annoying, I really don't, so I'll start by saying I really appreciate you looking into this :)

No worries, feel free to do it :)

Did you try with that scenario? That should also work. Here is a test case for it: https://github.com/smallrye/smallrye-config/pull/1534/changes#diff-10a7a73896298ce215fdc566c309d34ea2950f0dba01ccab5649f91a1edcf846R279

radcortez: Did you check?

yrodiere: Yes, sorry, I meant to come back to you. Unfortunately this does not work, but so far I was unable to find out why. I'll send a reproducer shortly.

yrodiere: I just can't get the same issue in Smallrye, maybe I'm doing something wrong.

Here is a reproducer in Quarkus, essentially just two commits on top of [my pending Quarkus PR](https://github.com/quarkusio/quarkus/pull/55032):

https://github.com/yrodiere/quarkus/commit/withunnamedkey-fix-test

The commit should have instructions on how to reproduce the problem. Careful, you need to rebuild all Quarkus first, because it's based on a rather big PR.

Let me know if something is unclear!

radcortez: Thanks. I'll have a look.

radcortez: Ok, the issue is because of `OptionalInt` (and similar). This is because these optionals call `getValue` with the converter instead of `getOptionalValue`. In this case, the default logic was not guarding against no value, and the comparison to the source was passing, because `ConfigValue` does not set the source ordinal in that case, so it was getting `0` and wrongly setting the resolve default flag. I've pushed the fix, and it should work now (I've tried it with your commit).

On the other hand, there is another issue. While this works for Build time, it won't work for Runtime. The reason is that `BUILD_AND_RUN_TIME_FIXED` configs are initialized at runtime from a separate `Config` instance with a single source set at `Integer.MAX_VALUE`. This ensures that no one can override the `BUILD_AND_RUN_TIME_FIXED` config, even programmatically. This means that at Runtime, the `BUILD_AND_RUN_TIME_FIXED` mapping will contain the unnamed key.

I'll try to think of a way to solve this.

yrodiere: Hey @radcortez , do you think we could merge your fix as-is? Even if it doesn't cover build config accessed at runtime, I think it's good enough in most cases, since we generally have other ways to iterate through datasources/persistence units at runtime.

radcortez: Sure. Was just waiting for your confirmation that it did work for the original case.

radcortez (approved):

yrodiere:
Thank you!

PR #1535, opened 2026-06-18
Fix HOCON property name rendering

### LINKED ISSUE #1225: HOCON properties rendered incorrectly
In HOCON, it's `Path.render` quotes anything it doesn't like, including things like `%` for profiles, `[]` for lists/arrays and `/` used in projects like SmallRye FaultTolerance.

Basically it produces _a lot_ of quoted sections to the properties it produces that make them incorrect. Some examples:

* Percent
  `%dev.some.property` becomes `"%dev".some.property
* Forward Slashes
  `org.acme.CoffeeResource/coffees/Retry/maxRetries=6` becomes `org.acme."CoffeeResource/coffees/Retry/maxRetries"=6`
* Brackets (which aren't usually necessary)
  `my.indexed.collection[0]=dog` becomes `my.indexed."collection[0]"=dog`

There doesn't seem to be a way, currently, to represent these properties in HOCON files.

xfocus3:
Fixes #1225

## Summary
- flatten HOCON objects by raw path segments instead of using Typesafe Config rendered paths
- render property names with SmallRye Config syntax so profile prefixes, slash-containing keys, and indexed property names remain usable
- preserve quoted segments when a segment contains dots

## Testing
- `mvn -pl sources/hocon -am -Dtest=HoconConfigSourceTest#renderedPropertyNamesUseSmallRyeConfigSyntax -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn -pl sources/hocon -am -Dtest=HoconConfigSourceTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn -pl sources/hocon -am test`

### DIFF
--- a/sources/hocon/src/main/java/io/smallrye/config/source/hocon/HoconConfigSource.java
+++ b/sources/hocon/src/main/java/io/smallrye/config/source/hocon/HoconConfigSource.java
@@ -7,8 +7,9 @@ import java.io.Reader;
 import java.io.Serial;
 import java.io.UncheckedIOException;
 import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.TreeMap;
 
 import com.typesafe.config.Config;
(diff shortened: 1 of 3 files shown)


radcortez (approved):

radcortez:
Thanks!

PR #1539, opened 2026-06-22
Remove deprecated Jasypt support

### LINKED ISSUE #1460: Remove Jasypt
Jasypt support was added due to being very popular with Spring, but the project is now unmaintained and archived:
https://github.com/jasypt/jasypt

- https://github.com/jasypt/jasypt/issues/162
- https://github.com/ulisesbocchio/jasypt-spring-boot/issues/414

We should remove it.

radcortez:
- Fixes #1460


cescoffier:
@radcortez Was this used in Quarkus? If so, we would need to add something to the migration guide.

radcortez:
With https://github.com/smallrye/smallrye-config/pull/1472, we started by throwing an exception if someone used it and with a configuration option to re-enable it.

Proper Quarkus support was provided with a Quarkiverse Extension:
https://docs.quarkiverse.io/quarkus-config-extensions/dev/jasypt.html

It was possible to use it directly in Quarkus without the Quarkiverse Extension, because it is included in the SmallRye Config BOM, but it wouldn't work in native (which required specific code for it).

Yes, I'll add the information to the migration guide. We still need to release SmallRye Config and update Quarkus with the released version.

PR #1546, opened 2026-07-15
@WithUnnamedKey#eager to control if default keys are present in the mapping Map

radcortez:
With https://github.com/smallrye/smallrye-config/pull/1534, we changed the behavior globally to not include the default key in a mapping `Map` if the configuration values were coming from the defaults. This can cause problems for existing integrations, so this behavior can now be controlled with `@WithUnnamedKey#eager` (the default is the original behavior).


yrodiere (approved):
Thank you. As discussed elsewhere, this should work for my usecase in Quarkus' Hibernate / Datasource extensions -- and we can always swap the default at a later date.

radcortez:
Thanks!

PR #1559, opened 2026-08-22
Do not collapse literal $$ in config values

### LINKED ISSUE #1056: Handling of $$ in value

arimu1:
Fixes #1056.

`Expression` treats `$$` as an escaped `$`. Mini expressions are not enabled (`NO_SMART_BRACES`), so that collapse is only needed immediately before `{` to suppress `${...}` expansion. Undocumented `$$` → `$` rewriting currently mangles values such as AWS-rotated DB passwords that contain `$$` (`ab$$cd` becomes `ab$cd`).

This keeps:
- `${...}` expansion
- `$${expr}` / `\${expr}` as a literal `${expr}` (MP Config `\$` still maps to Expression `$$`)
- `\$` as a literal `$`

and only rewrites other `$$` sequences so evaluation still emits a literal `$$`. Values with no `${` and no `\$` skip expression compilation.

## Testing

- `ExpressionConfigSourceInterceptorTest#doubleDollarLiteral` fails on current `main` (`ab$$cd` → `ab$cd`) and passes with this change
- `mvn -pl implementation -am test` (Temurin 21): **704 run, 0 fail, 15 skipped**
- `formatter:validate` and `impsort:check` on `implementation`

### DIFF
--- a/implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java
+++ b/implementation/src/main/java/io/smallrye/config/ExpressionConfigSourceInterceptor.java
@@ -62,13 +62,18 @@ public class ExpressionConfigSourceInterceptor implements ConfigSourceIntercepto
             return null;
         }
 
+        String value = configValue.getValue();
         // Avoid extra StringBuilder allocations from Expression
-        if (configValue.getValue().indexOf('$') == -1) {
+        if (value.indexOf('$') == -1) {
+            return configValue;
+        }
+        // Mini expressions are off, so a lone $ or $$ is not an expression.
+        if (value.indexOf("${") == -1 && value.indexOf("\\$") == -1) {
             return configValue;
         }
 
         ConfigValue.ConfigValueBuilder valueBuilder = configValue.from();
-        Expression expression = Expression.compile(escapeDollarIfExists(configValue.getValue()), LENIENT_SYNTAX, NO_TRIM,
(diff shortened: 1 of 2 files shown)


dmlloyd:
IMHO I think any preprocessing of expression strings is a bad idea. We end up with bizarre behavior if it's not thought through exceedingly carefully. If there's a syntax limitation with the upstream expression parser, we should try to resolve the problem there, IMO.

Also, passwords should not be given in config IMO. Any time we need a password from configuration, the configuration property should be a symbolic reference to a secure password store of some sort.

radcortez:
Our goal was to remove the `$$` escaping via https://github.com/smallrye/smallrye-common/pull/344, but we never did because it would be a major breaking change.

Due to Quarkus 4, I'm taking the opportunity to break a few things, so I think we can do it now.

radcortez:
Closing in favour of #1568

radcortez:
Thank you for the PR. We went in another direction, but it should fix the issue.

PR #1577, opened 2026-09-09
Make FileSystemConfigSource take into account Win line endings

kabir:
We found this regression here https://github.com/wildfly/wildfly/pull/20354#issuecomment-5585179334 in a test that has been working for years.

In our tests it creates the files with `Files.write(path, Collections.singletonList(value)).`, so that on Windows it gets the extra terminator. So assuming `\r\n` is allowed as a line ending in these files, this is a regression.


rhusar (approved):
LGTM

radcortez (approved):

PR #1589, opened 2026-09-15
Use ClassValue to cache ConfigMappingInterface and ConfigMappingClass

radcortez:


implementation/src/main/java/io/smallrye/config/ConfigMappingClass.java
```diff
@@ -44,39 +43,74 @@ default Object map() {
...
+     * {@code WeakHashMap} keyed by {@link Class} cannot do the same, because the metadata strongly references its own
+     * key, which keeps the entry alive forever. The {@link Holder} allows the metadata to be computed by the caller,
+     * which requires a {@link ConfigMappingHandler} that {@code ClassValue#computeValue(Class)} does not receive.
+     * <p>
+     * Only a type accepted by {@link #isConfigurationClass(Class)} may be looked up in the cache. The cache is probed
+     * with arbitrary types to find out whether they are configuration classes, and because the {@link Holder} is
+     * stored even when there is no metadata to store in it, an entry for a type we do not own, like a JDK type, keeps
+     * the SmallRye Config {@link ClassLoader} alive for as long as that type, which is forever.
+     */
+    private static final ClassValue<Holder<ConfigMappingClass>> CACHE = new ClassValue<>() {
```
dmlloyd, line 58: This will leak. The only way to use `ClassValue` in a non-leaky way is for the value to have *no* reference to any class loader other than the boot and app class loaders.

radcortez: Yes, inadvertently, I made it worse with the `Holder`, but I can fix that.

We always had a possible leak in the original design:
https://github.com/smallrye/smallrye-config/commit/8bc31bf3c313f5938924e8272e44038ad8fd12a9#diff-ac0660644bdc2ae1618e751ee16790f96edce6e2e9e7141bcfd2e8fe4437ed0fR47

But we never observed it in Quarkus, since the key is always in the same loader as SmallRye Config, or in a descendant, so when we drop the CL, we drop everything.

Should we add a check if the type we are trying to cache is in the same loader of SmallRye Config?

dmlloyd: The type of the key doesn't actually matter - it can be of any class loader, safely. It's the value that's the problem. If the value is of a class loader other than one of the three JDK built in class loaders, the value will never be unloaded. You would *think* that a `ClassValue` entry would only create a strong reference from the type in question to the value, but it doesn't actually work that way at all. Basically `ClassValue` is a trap that is no good for 95% of use cases.

The only adequate solution to this problem that I've found is to not have global by-type caches at all, and instead to have some kind of disposable long-lived context object that retains the cache information instead.

radcortez: Yes, I mentioned the key because it ends up in the value as well (because it is referenced in a field of the value), creating the strong reference.

What I've observed in Quarkus is that, since everything is in the same CL, when we drop it, everything is collected correctly, but if we end up referencing a Class in the cache from the parent CL, or any of the JDK built-in class loaders, we do get the leak, but we don't have that case.

We can certainly remove some of these pieces. The caches are mostly to avoid reintrospecting mapping classes. Maybe we could link this to the builder instead of being static, so caches are collected after we call `.build`. For Quarkus this shouldn't be a problem, because this happens in deployment, and at runtime we don't even hit the caches, since we already have the generated class. It could make other consumers that are runtime only to pay a higher init price, if multiple config instances exist in the same app, with the same mappings.

